PR #424, opened 2020-06-26
Add 'languages' to the npmignore whitelist.

jansiegel:
### Context
In `0.1.0` the `languages` directory is missing from the `npm` package, which makes languages impossible to import.

PR #432, opened 2020-06-30
Correct the misunderstanding with the language packs.

jansiegel:
### Context
#424 added the `/languages` directory to the npm package, which turned not needed, as the language packs are already present in their bundle-specific directories.

#### Changes in this PR:
- Remove the `languages` directory from the `.npmignore` whitelist (rollback #424)
- Prepend the `bundle-all` command with the `clean` command
- Correct and extend the information about importing the language packs in the documentation
- Rollback the #424 entry from the `CHANGELOG.md` file
- Remove the test for the `languages` directory from the package-checking script.

### How has this been tested?
Tested manually + with the codesandbox example.

### Types of changes

### Related issue(s):
1. #424

### DIFF
--- a/script/check-publish-package.js
+++ b/script/check-publish-package.js
@@ -33,6 +33,9 @@ const FILES_CHECKLIST = [
   'es/HyperFormula.js',
   'typings/index.d.ts',
   'typings/HyperFormula.d.ts',
+  'dist/languages/plPL.js',
+  'es/i18n/languages/plPL.js',
+  'commonjs/i18n/languages/plPL.js',
 ]
 
 process.stdin.resume()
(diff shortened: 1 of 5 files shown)


script/check-publish-package.js
```diff
@@ -33,7 +33,6 @@ const FILES_CHECKLIST = [
   'es/HyperFormula.js',
   'typings/index.d.ts',
   'typings/HyperFormula.d.ts',
-  'languages/all.js',
```
wojciechczerniak, line 36: Maybe we should check `/dist/languages/all.js` then? Might be a good idea to check that language budnles were generated

PR #450, opened 2020-07-14
Add a manual deploy option to the gh-pages workflow.

jansiegel:
### Context
Add a manual deploy option to the GH Actions `gh-pages` workflow.


wojciechczerniak:
We've decided that will get back to this. Have to discuss the documentation publishing and versioning first

PR #458, opened 2020-07-18
Add  the documentation issue template

scarletfog:
### Context
We need a separate issue template for all sorts of documentation to facilitate a smoother process of requesting new pieces of documentation and tracking errors. The issue template should add the "Documentation" label automatically so it is easily distinguishable from other tasks.

@wojciechczerniak do you have any ideas on how to make it better?


wojciechczerniak:
Maybe a request for a link to a page that should be improved or the mistake is on?

Also can be auto assigned to you 😄

PR #480, opened 2020-07-28
Pu/functions

izulin:
### Context

Added pseudo-type system for functions -- we can now specify the signature of a function, and necessary coercions are made out of it.
Example:
in BooleanPlugin:

```
    'IF': {
      method: 'conditionalIf',
      parameters: {
        list: [
          {argumentType: 'boolean'},
          {argumentType: 'scalar'},
          {argumentType: 'scalar', defaultValue: false},
        ]
      },
    },
```

and then

```
  public conditionalIf(ast: ProcedureAst, formulaAddress: SimpleCellAddress): InterpreterValue {
    return this.runFunction(ast.args, formulaAddress, this.parameters('IF'), (condition, arg2, arg3) => {
      return condition ? arg2 : arg3
    })
  }
```

A lot of functions have been rewritten into this system, which simplifies and unifies the coercion system for their arguments.
Also, fixes a few minor bugs.

### Types of changes

### Related issue(s):
1.
2.
3.

### Checklist:

- [ ] My code follows the code style of this project,


izulin:
> * Is the double nesting of `parameters` and `list` necessary?

yes, it allows us to encapsulate *just* the metadata necessary for template in a single object

izulin:
> * `repeatedArg` is for repeating a set of arguments, right?

for repeating just the single, last argument on the list (this seems limited, but is probably all that we need)

izulin:
> * Do we have an API to get the functions descriptors? ie. for formula editor and autocomplete or validation of a selected cell type

not really, once we finalize structure of those parameters (e.g. no nasty surprises in not-implemented-yet functions) we can think of it in some separate issue

izulin:
> * Should `argumentType` be an `enum`? IMO type list should be limited and known

we have
```
export type ArgumentTypes = 'string' | 'number' | 'boolean' | 'scalar' | 'noerror' | 'range' | 'integer'
```

any suggestions for better solution?

izulin:
> * Do you have a description of all params? I've seen `argumentType`, `minValue`, `maxValue`, are there others?

Added docs in src/interpreter/plugin/FunctionPlugin.ts

wojciechczerniak:
Great, thanks for the docs @izulin!

`minValue` and `greaterThan` looks kind of redundant. But I see the subtle difference between those two

I would like to propose that we flatten the `parameters` and `list`. We can keep the param attributes flat with all the function other attributes. IMO it's an attribute like any other. WDYT?
```
    'COUNTUNIQUE': {
      method: 'countunique',
      parameters: [
          {argumentType: ArgumentTypes.SCALAR},
      ],
      repeatLastArg: true,
      expandRanges: true,
    },
```

test/interpreter/function-vlookup.spec.ts
```diff
@@ -6,15 +6,15 @@ import {Sheet} from '../../src/Sheet'
 
 const sharedExamples = (builder: (sheet: Sheet, config?: Partial<ConfigParams>) => HyperFormula) => {
   describe('VLOOKUP - args validation', () => {
-    it('not enough parameters', () => {
+    it('not enough metadata', () => {
```
izulin, line 9:
```suggestion
    it('not enough metadata', () => {
```
 it('not enough parameters', () => {

test/interpreter/function-vlookup.spec.ts
```diff
@@ -6,15 +6,15 @@ import {Sheet} from '../../src/Sheet'
...
       const engine = builder([
         ['=VLOOKUP(1, A2:B3)'],
       ])
 
       expect(engine.getCellValue(adr('A1'))).toEqual(detailedError(ErrorType.NA))
     })
 
-    it('too many parameters', () => {
+    it('too many metadata', () => {
```
izulin, line 17:
```suggestion
    it('too many metadata', () => {
```
it('too many parameters', () => {

izulin:
> Great, thanks for the docs @izulin!
> 
> `minValue` and `greaterThan` looks kind of redundant. But I see the subtle difference between those two
> 
> I would like to propose that we flatten the `parameters` and `list`. We can keep the param attributes flat with all the function other attributes. IMO it's an attribute like any other. WDYT?
> 
> ```
>     'COUNTUNIQUE': {
>       method: 'countunique',
>       parameters: [
>           {argumentType: ArgumentTypes.SCALAR},
>       ],
>       repeatLastArg: true,
>       expandRanges: true,
>     },
> ```

Is it better now?

PR #484, opened 2020-08-07
Correct the docs after proofreading

scarletfog:
### Context

Apply the changes suggested by a native after the proofread. Styling, better use of articles etc.
Fix the issue with displaying multiplication sign:

image


wojciechczerniak (approved):
🍪

Issue #476, opened 2020-07-21
if returns empty string instead 0


doczoidberg:
following formula should result to 0 but results in an empty string

![88040737-a0db5000-cb49-11ea-8437-83592b2fc4ca](https://user-images.githubusercontent.com/1972209/88048528-c79e8400-cb53-11ea-9d00-6fdcd0d86e00.png)

this is important because e.g. the formula can be encapsulated by another one e.g. `VLOOKUP` or in another `IF` statement

krzysztofspilka:
@izulin do we have any update on this?

bardek8:
We started discussing this offline. Actually this case is quite complicated as this is the place of inconsistency between Excel, Google Sheets and LibreOffice Calc.
With the current implementation of HyperFormula we cannot behave as in Excel - this will require storing auxiliary data with each cell value and processing it appropriately. Currently we behave as in Google Sheets.

See the difference between Google Sheets:
![gs](https://user-images.githubusercontent.com/16952569/89042665-9427da80-d347-11ea-9de2-df264905d438.png)

and Excel:
![excel](https://user-images.githubusercontent.com/16952569/89042645-8d996300-d347-11ea-9742-e3d893c1df97.png)

doczoidberg:
maybe a global setting which format and an adaption of thes `if  function` would be enough?

only a special if check for excel would  be enough where 0==undefined results to tue?

Your plans are not to support Excel soon?

AMBudnik:
Thank you for doing the investigation @bardek8. I've spoken with @krzysztofspilka and we both agree that we should consider switching to Excel's approach but as it is not an easy process we should mark this topic as a feature request and get back to it after doing the full investigation. 

@doczoidberg we'll update you on every change to this topic.

doczoidberg:
@AMBudnik  can you give a rough time estimation?

AMBudnik:
@doczoidberg some of our crucial team members are ooo, we cannot start the process without them. As soon as they meet and agree on the next step towards the goal this topic will get updated accordingly.

doczoidberg:
any news here?

bardek8:
Sorry for the late reply. Unfortunately, currently we cannot provide any precise time estimations.
This is an involved issue that requires deep understanding of the Excel's behavior, which is very different from GoogleSheets and LibreOffice. We envisage that this will be a big change in HF, so we decided to address it in the later future, after we finish the project of adding new functions to HF.

[...]

PR #500, opened 2020-08-31
Pu/null compatibility

izulin:
### Context

Added config flag for compatibility with Excel for formulas that evaluate to null.

### Types of changes

### Related issue(s):
1. #476 
2.
3.

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added 15 date functions: WEEKDAY, DATEVALUE, HOUR, MINUTE, SECOND, TIME, TIMEVALUE, NOW, TODAY, EDATE, WEEKNUM, ISOWEEKNUM, DATEDIF, DAYS360, YEARFRAC.
 - Added 13 trigonometry functions: SEC, CSC, SINH, COSH, TANH, COTH, SECH, CSCH, ACOT, ASINH, ACOSH, ATANH, ACOTH.
 - Added 6 engineering functions: OCT2BIN, OCT2DEC, OCT2HEX, HEX2BIN, HEX2OCT, HEX2DEC.
+- Added Excel compatibility mode for formulas evaluating to null-value.
 
 ### Changed
 - Operation `moveCells` creating cyclic dependencies does not cause losing original formula. (#479)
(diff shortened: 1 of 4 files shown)


src/Config.ts
```diff
@@ -198,6 +198,16 @@ export interface ConfigParams {
...
+  /**
+   * Sets the compatibility mode for behaviour of null value.
+   * If set, formula evaluating to null evaluates to 0 instead,
+   * as in Excel.
+   *
+   * @default false
+   *
+   * @category Engine
+   */
+  nullCompatibility: boolean,
```
wojciechczerniak, line 210: This option should be generic and not mention other product by name. It's not a compatibility but a vendor behavior. We need a better name

We can write a documentation or blog post how compatiblity with certaing product can be achived, but IMO it's just a matter of time when we see other products that evaluate empty cells to zero.

test/null-compatibility.spec.ts
```diff
@@ -0,0 +1,46 @@
+import {HyperFormula} from '../src'
+import {adr} from './testUtils'
+
+describe('null compatibility', () => {
+  it('reference - default behaviour', () => {
```
wojciechczerniak, line 5: Tests should describe what are the expectations.

If someone introduces a bug in test, this may still be accepted as the "default behaviour" because noone reading the code knows what it is

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added 15 date functions: WEEKDAY, DATEVALUE, HOUR, MINUTE, SECOND, TIME, TIMEVALUE, NOW, TODAY, EDATE, WEEKNUM, ISOWEEKNUM, DATEDIF, DAYS360, YEARFRAC.
 - Added 13 trigonometry functions: SEC, CSC, SINH, COSH, TANH, COTH, SECH, CSCH, ACOT, ASINH, ACOSH, ATANH, ACOTH.
 - Added 6 engineering functions: OCT2BIN, OCT2DEC, OCT2HEX, HEX2BIN, HEX2OCT, HEX2DEC.
+- Added Excel compatibility mode for formulas evaluating to null-value.
```
wojciechczerniak, line 18: added a configuration option to evaluate reference to an empty cells as a zero?

wojciechczerniak (changes requested):

src/Config.ts
izulin: any suggestions?

bardek8: maybe evaluateNullToZero ?

izulin: does it reflect that it happens only *at the very end* of formula evaluation?

wojciechczerniak (approved):
LGTM 👍

Issue #517, opened 2020-09-18
Prepare the documentation for the release


scarletfog:
### Description
We need several things to prepare in the documentation before the upcoming release(s).

### Checklist
- [x] Add missing information in the changelog.md file (a link to: https://github.com/handsontable/hyperformula/pull/422), tidy the changelog up
- [x] Add a `@since` tag for new methods in the API reference
- [x] Mark the versions of functions (can be vuepress badges)
- [x] Check if the documentation was updated with new functions
- [x] Update changelog in the documentation, add the release date

### Type of issue
- [x] Missing documentation

scarletfog:
done in https://github.com/handsontable/hyperformula/pull/519

PR #519, opened 2020-09-19
Prepare for the release [DOCS]

scarletfog:
### Context
We need to mark which functions and methods work from 0.2.0.

### Types of changes

- [x] A change in the documentation

### Related issue(s):
1. https://github.com/handsontable/hyperformula/issues/517

@wojciechczerniak  `@since` tag does not make me too..happy when I look at how it is displayed. I put it at the end but it appeared under the example:
image

do you know if we can do something about it?

And the badges look like this 🙂 :
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Added 9 text functions LEN, TRIM, PROPER, CLEAN, REPT, RIGHT, LEFT, SEARCH, FIND. (#221)
-- Added helper methods for keeping track of cell/range dependencies: getCellPrecedents and getCellDependents. (#441)
+- Added helper methods for keeping track of cell/range dependencies: `getCellPrecedents` and `getCellDependents`. (#441)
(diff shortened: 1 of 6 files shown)


wojciechczerniak:
Badges look good. But can we have them always in a new line?

`@since` tag display is dependent on the template. Just mark the methods correctly, we can adjust the display later. First, let's make the documentation a reliable source of truth. Then we can adjust the presentation in a separate task.

scarletfog:
@wojciechczerniak please have a look. 

I think this is pretty much all for `@since` tags: `getCellPrecedents`, `getCellDependents`, `ErrorMessage` and one extra config param (`evaluateNullToZero`)?

Badges go _br...br.._ ❄️  but I have no clue how it could be done better (+ not to break other tables). I tried with `display`: `table`, `block` and `inline-block` and also a naive attempt with `::before` but there are some problems with each approach. If you have something in mind please share.

I updated changelogs too.

@aninde I remember you had some remarks, could you please check if it's better now?

wojciechczerniak:
> I think this is pretty much all for `@since` tags: `getCellPrecedents`, `getCellDependents`, `ErrorMessage` and one extra config param (`evaluateNullToZero`)?

You've got all of them 👍 

> Badges go _br...br.._ ❄️ but I have no clue how it could be done better (+ not to break other tables). I tried with `display`: `table`, `block` and `inline-block` and also a naive attempt with `::before` but there are some problems with each approach. If you have something in mind please share.

`<BR>` is the correct, semantic way for a new line.

wojciechczerniak (approved):

Issue #540, opened 2020-10-07
CI unit tests are running out of memory


wojciechczerniak:
### Description

- [ ] Investigate why Jest unit tests take 10x longer than Jasmine.

![obraz](https://user-images.githubusercontent.com/4718998/95320359-abc89580-0899-11eb-9260-10a733905124.png)

- [x] Investigate "JavaScript heap out of memory" unit test error

https://github.com/handsontable/hyperformula/pull/531/checks?check_run_id=1182759497

- [x] Update Node to v14 (may fix the above)

### Steps to reproduce

1. Add PR that will run CI tasks
2.
3.

### Demo
#538

### Your environment

* HyperFormula version: develop
* Browser Name and version: any
* Operating System: any

### Links
#538
https://github.com/handsontable/hyperformula/pull/531/checks?check_run_id=1182759497

wojciechczerniak:
Done in #547

PR #547, opened 2020-10-19
Stable unit tests in CI

swistach:
### Context
Due to memory leaks in jest test env and the fact that we cannot define an exact place in jest tool where happen memory leaks, we have to add some tricks with exposing node's GC to tests and clearing it manually in `beforeAll` hook.

### How has this been tested?
Push a commit to GA and wait for `UNIT tests` workflow results.

### Types of changes
- [x] test env improvements

### Related issue(s):
1. #540


wojciechczerniak (approved):
👍

PR #551, opened 2020-10-21
math functions

izulin:
### Context

18 mathematical functions ROMAN, ARABIC, FACT, FACTDOUBLE, COMBIN, COMBINA, GCD, LCM, MROUND, MULTINOMIAL, QUOTIENT, RANDBETWEEN, SERIESSUM, SIGN, SQRTPI, SUMX2MY2, SUMX2PY2, SUMXMY2.

### Types of changes

### Related issue(s):
1.
2.
3.

### Checklist:


test/interpreter/function-gcd.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    expect(engine.getCellValue(adr('A1'))).toEqualError(detailedError(ErrorType.VALUE, ErrorMessage.NumberCoercion))
+  })
+
+  it('checks bounds', () => {
+    const engine = HyperFormula.buildFromArray([
+      ['=GCD(-1,5)'],
+    ])
+
+    expect(engine.getCellValue(adr('A1'))).toEqualError(detailedError(ErrorType.NUM, ErrorMessage.ValueSmall))
```
wojciechczerniak, line 89: Should we also check value too big? "If a parameter to GCD is >=2^53, GCD returns the #NUM! error value."

test/interpreter/function-lcm.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    expect(engine.getCellValue(adr('A1'))).toEqualError(detailedError(ErrorType.VALUE, ErrorMessage.NumberCoercion))
+  })
+
+  it('checks bounds', () => {
+    const engine = HyperFormula.buildFromArray([
+      ['=LCM(-1,5)'],
+    ])
+
+    expect(engine.getCellValue(adr('A1'))).toEqualError(detailedError(ErrorType.NUM, ErrorMessage.ValueSmall))
```
wojciechczerniak, line 89: Can we check this edge case as well: "If LCM(a,b) >=2^53, LCM returns the #NUM! error value." ?

https://support.microsoft.com/en-us/office/lcm-function-7152b67a-8bb5-4075-ae5c-06ede5563c94

test/interpreter/function-sign.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+      ['=SIGN("foo")'],
+    ])
+
+    expect(engine.getCellValue(adr('A1'))).toEqualError(detailedError(ErrorType.VALUE, ErrorMessage.NumberCoercion))
+  })
+
+  it('should work', () => {
+    const engine = HyperFormula.buildFromArray([
+      ['=SIGN(5)'],
+      ['=SIGN(0)'],
```
wojciechczerniak, line 28: Just to be sure. `=SIGN(-0)` will be handled by us, and `-0` will be normalized before it gets to the SIGN function?

test/interpreter/function-arabic.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  it('should output correct value for roman numerals from mode 4', () => {
+    const [input, output] = inputOutput(4)
+    const engine = HyperFormula.buildFromArray([input])
+    expect(engine.getSheetValues(0)).toEqual([output])
+  })
+})
+
+function inputOutput(mode: number) {
+  const arr = [mode0, mode1, mode2, mode3, mode4][mode]
```
wojciechczerniak, line 86: We can create this array once, outside of the function. We run tests a lot, maybe it's desired to reduce time and memory footprint. WDYT?

```suggestion
const modes = [mode0, mode1, mode2, mode3, mode4];
function inputOutput(mode: number) {
  const arr = modes[mode]
```

src/interpreter/plugin/RomanPlugin.ts
```diff
@@ -0,0 +1,180 @@
...
+        [input, ans] = eatToken(input, ans,
+          {token: 'MMM', val: 3000},
+          {token: 'MM', val: 2000},
+          {token: 'M', val: 1000},);
+        [input, ans] = eatToken(input, ans,
+          {token: 'IM', val: 999},
+          {token: 'VM', val: 995},
+          {token: 'XM', val: 990},
+          {token: 'LM', val: 950},
+          {token: 'CM', val: 900},);
```
wojciechczerniak, line 75: Either ending comma or parenthesis. I would move it to the next line. Please check the whole file
```suggestion
          {token: 'CM', val: 900},
        );
```

src/interpreter/plugin/RomanPlugin.ts
```diff
@@ -0,0 +1,180 @@
...
+      method: 'roman',
+      parameters: [
+        { argumentType: ArgumentTypes.NUMBER, minValue: 1, lessThan: 4000 },
+        { argumentType: ArgumentTypes.NOERROR, optionalArg: true, defaultValue: 0 }
+      ],
+    },
+    'ARABIC': {
+      method: 'arabic',
+      parameters: [
+        { argumentType: ArgumentTypes.STRING},
```
wojciechczerniak, line 23:
```suggestion
        { argumentType: ArgumentTypes.STRING },
```

src/interpreter/plugin/MathPlugin.ts
```diff
@@ -0,0 +1,393 @@
...
+    [a, b] = [b, a]
+  }
+  while(b>0) {
+    [a, b] = [b, a%b]
+  }
+  return a
+}
+
+function binaryLCM(a: number, b: number): number {
+  if(a==0 || b===0) {
```
wojciechczerniak, line 389: One is a strict comparison, other not. Is that on purpose?

test/interpreter/function-arabic.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+    expect(engine.getCellValue(adr('A4'))).toEqualError(detailedError(ErrorType.VALUE, ErrorMessage.Roman))
+    expect(engine.getCellValue(adr('A5'))).toEqualError(detailedError(ErrorType.VALUE, ErrorMessage.Roman))
+    expect(engine.getCellValue(adr('A6'))).toEqualError(detailedError(ErrorType.VALUE, ErrorMessage.Roman))
+    expect(engine.getCellValue(adr('A7'))).toEqualError(detailedError(ErrorType.VALUE, ErrorMessage.Roman))
+  })
+
+  it('works for border cases', () => {
+    const engine = HyperFormula.buildFromArray([
+      ['=ARABIC("MMMIMIDCCCICILXXXIXIVIII")'],
+      ['=ARABIC("-I")'],
```
wojciechczerniak, line 48: I've got another border case:
- empty string: `If an empty string ("") is used as an input value, 0 is returned.`
- case insensitive: `The case of the text argument is ignored. For example, "mxmvii" evaluates to the same result (1997) as "MXMVII."`

Source: https://support.microsoft.com/en-us/office/arabic-function-9a8da418-c17b-4ef9-a657-9370a30a674f

test/interpreter/function-roman.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+  it('should properly truncate values and use defaults', () => {
+    const engine = HyperFormula.buildFromArray([
+      ['=ROMAN(499)'],
+      ['=ROMAN(499, TRUE())'],
+      ['=ROMAN(499, FALSE())'],
+      ['=ROMAN(499.9)'],
+      ['=ROMAN(499, 1.1)'],
+    ])
+    expect(engine.getCellValue(adr('A1'))).toEqual('CDXCIX')
+    expect(engine.getCellValue(adr('A2'))).toEqual('ID')
```
wojciechczerniak, line 24: I have different results in Excel. According to the documentation, both `default` aka mode `0` is "Classic", the same as `TRUE`. 

`FALSE` should be the same as mode `4`. Maybe we can add here an example with mode 4 "Simplified", next to the `FALSE`.
Screenshot 2020-10-22 at 19 40 11


Source: https://support.microsoft.com/en-us/office/roman-function-d6b0b99e-de46-4704-a518-b45a0f8b56f5

test/interpreter/function-roman.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+    expect(engine.getCellValue(adr('A3'))).toEqual('CDXCIX')
+    expect(engine.getCellValue(adr('A4'))).toEqual('CDXCIX')
+    expect(engine.getCellValue(adr('A5'))).toEqual('LDVLIV')
+  })
+
+  it('should throw correct error', () => {
```
wojciechczerniak, line 30: IMO small improvement:
```suggestion
  it('should throw correct error if arguments are out of bounds', () => {
```

wojciechczerniak (changes requested):
Awesome 💪 

Just small things and questions

test/interpreter/function-sign.spec.ts
izulin: of course

test/interpreter/function-arabic.spec.ts
izulin: there is no memory copying involved in the creation of this array, so this is already fast

izulin: (and this code is run 5 times in total during testing)

test/interpreter/function-arabic.spec.ts
izulin: but those are checked

test/interpreter/function-roman.spec.ts
izulin: ok, thanks!

test/interpreter/function-mround.spec.ts
```diff
@@ -0,0 +1,64 @@
+import {HyperFormula} from '../../src'
+import {ErrorType} from '../../src/Cell'
+import {ErrorMessage} from '../../src/error-message'
+import {adr, detailedError} from '../testUtils'
+
+describe('Function MROUND', () => {
```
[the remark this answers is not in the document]
izulin: not sure how we handle -- probably random as well. Trick is, it depends on how the (imprecise) floating point operations land against the threshold. This might be dependent on js math library implementation and might be not consistent even between browsers.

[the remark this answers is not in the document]
izulin: (we round exactly the same as Excel -- at least on my machine, with my js version. That might suggest that deep down we use the same low-level floating point operations. But still, I wouldn't count on consistency with this.)

src/error-message.ts
```diff
@@ -66,6 +66,8 @@ export class ErrorMessage {
   public static NonZero = 'Argument cannot be 0.'
   public static LessThanOne = 'Argument cannot be less than 1.'
   public static WeekendString = 'Incorrect weekend bitmask string.'
+  public static Roman = 'Invalid roman numeral.'
+  public static WrongOrder = 'Values in wrong order.'
```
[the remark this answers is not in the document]
scarletfog: I know it may sound like a pun...but `InvalidRoman` would be more descriptive.
and:
"Wrong order of values"

test/interpreter/function-arabic.spec.ts
wojciechczerniak: Ok, I see case sensitivity in sanitize section. But where is an empty string edge case covered?

test/interpreter/function-arabic.spec.ts
wojciechczerniak: Each test adds to the total 20m tests run. But ok, we can leave it as is

test/interpreter/function-mround.spec.ts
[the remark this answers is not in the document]
wojciechczerniak: Oh, that's tough. I thought that JS is based on IE 754 and we have our own precision control so with the same settings as Excel (our defaults) we should get the same result.

Maybe we can enable the test and check if it will pass on CI? If any other developer will report it, we will disable it then

test/interpreter/function-arabic.spec.ts
izulin: it is now :)

src/interpreter/plugin/RomanPlugin.ts
```diff
@@ -0,0 +1,187 @@
...
+          if(input==='') {
+            return new CellError(ErrorType.VALUE, ErrorMessage.InvalidRoman)
+          }
+        }
+        let ans = 0;
+        [input, ans] = eatToken(input, ans,
+          {token: 'MMM', val: 3000},
+          {token: 'MM', val: 2000},
+          {token: 'M', val: 1000},
+          );
```
wojciechczerniak, line 70: Can you check the file for semicolons? It's not our current coding style (I don't agree with this rule but we should be consitent if we have it)

```suggestion
        )
```

izulin: without semicolons, I getting TS compilation errors.

izulin:
```
src/interpreter/plugin/RomanPlugin.ts:66:24 - error TS2322: Type '[string, number]' is not assignable to type 'string | number'.
  Type '[string, number]' is not assignable to type 'string'.

```

test/interpreter/function-arabic.spec.ts
izulin: the speedup is not even measurable, and due to quirks of JS compilation could not even be a speedup

src/interpreter/plugin/RomanPlugin.ts
wojciechczerniak: That's why I prefer semicolons :| 

Ok. Maybe we can refactor `eatToken` and `absorb` to consume an array instead of two arguments?

```ts
res = eatToken(res, ...)
```

As a bonus we would avoid destructuring it constantly.

wojciechczerniak (approved):
Awesome! Thanks 🏆

test/interpreter/function-gcd.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    expect(engine.getCellValue(adr('A1'))).toEqualError(detailedError(ErrorType.VALUE, ErrorMessage.NumberCoercion))
+  })
+
+  it('handles overflow', () => {
+    const engine = HyperFormula.buildFromArray([
+      ['=GCD(1000000000000000000.0)'],
+    ])
+
+    expect(engine.getCellValue(adr('A1'))).toEqualError(detailedError(ErrorType.NUM, ErrorMessage.ValueLarge))
```
bardek8, line 89: interestingly, now GS allows it, when XL doesn't.

test/interpreter/function-lcm.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+    expect(engine.getCellValue(adr('A1'))).toBe(1)
+    expect(engine.getCellValue(adr('B1'))).toBe(0)
+  })
+
+  it('handles overflow', () => {
+    const engine = HyperFormula.buildFromArray([
+      ['=LCM(1000000,1000001,1000002,1000003)'],
+    ])
+
+    expect(engine.getCellValue(adr('A1'))).toEqualError(detailedError(ErrorType.NUM, ErrorMessage.ValueLarge))
```
bardek8, line 55: same here

bardek8 (changes requested):
Thanks! It's great, I have only some tiny comments about inconsistencies between XL/GS, everything else is fine.

src/interpreter/plugin/MathPlugin.ts
```diff
@@ -0,0 +1,398 @@
...
+        return ret
+      }
+    )
+  }
+
+  public combin(ast: ProcedureAst, formulaAddress: SimpleCellAddress): InternalScalarValue {
+    return this.runFunction(ast.args, formulaAddress, this.metadata('COMBIN'),
+      (n: number, m: number) => {
+        if(m>n) {
+          return new CellError(ErrorType.NUM, ErrorMessage.ValueLarge)
```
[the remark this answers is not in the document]
izulin: do you mean the error message?

src/interpreter/plugin/MathPlugin.ts
```diff
@@ -0,0 +1,398 @@
...
+  }
+}
+
+function combin(n: number, m: number): number {
+  if(2*m>n) {
+    m = n-m
+  }
+  let ret = 1
+  for(let i=1;i<=m;i++) {
+    ret *= (n-m+i)/i
```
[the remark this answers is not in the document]
izulin: yes, but then intermediate results might generate an overflow

src/interpreter/plugin/MathPlugin.ts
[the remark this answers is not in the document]
bardek8: Yes. Because it is either too large `m` or too small `n`.

bardek8 (approved):
Thanks!

PR #599, opened 2021-01-30
Pu/extratyping

izulin:
### Types of changes

### Related issue(s):
1.
2.
3.

### Checklist:

- [ ] My code follows the code style of this project,

### DIFF
--- a/docs/guide/known-limitations.md
+++ b/docs/guide/known-limitations.md
@@ -28,7 +28,6 @@ you can't compare the arguments in a formula like this:
 * There is no relative referencing in named ranges.
 * The library doesn't offer (at least not yet) the following features:
   * 3D references
-  * Complex numbers
   * Constant arrays
   * Dynamic arrays
   * Asynchronous functions
--- a/docs/guide/known-limitations.md
+++ b/docs/guide/known-limitations.md
@@ -32,7 +32,6 @@ you can't compare the arguments in a formula like this:
   * Dynamic arrays
   * Asynchronous functions
   * Structured references ("Tables")
-  * Currency data type
   * Relative named expressions
   * Functions cannot use UI metadata (e.g. hidden rows for SUBTOTAL).
 
--- a/src/HyperFormula.ts
+++ b/src/HyperFormula.ts
@@ -2845,10 +2845,10 @@ export class HyperFormula implements TypedEmitter {
    * ]);
    *
    * // should return 'NUMBER_PERCENT', cell value type of provided coordinates is a number with a format inference percent.
-   * const cellType = hfInstance.getCellValueType({ sheet: 0, col: 1, row: 0 });
+   * const cellType = hfInstance.getCellValueType({ sheet: 0, col: 0, row: 0 });
    *
    * // should return 'NUMBER_CURRENCY', cell value type of provided coordinates is a number with a format inference currency.
-   * const cellType = hfInstance.getCellValueType({ sheet: 0, col: 0, row: 0 });
+   * const cellType = hfInstance.getCellValueType({ sheet: 0, col: 1, row: 0 });
    * ```
    *
    * @category Cells
@@ -2860,7 +2860,7 @@ export class HyperFormula implements TypedEmitter {
   }
 
   /**
-   * Returns detailed type of the cell value of a given address.
+   * Returns auxilary format information of the cell value of a given address.
    * The methods accepts cell coordinates as object with column, row and sheet numbers.
    *
    * @param {SimpleCellAddress} cellAddress - cell coordinates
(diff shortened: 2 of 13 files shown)


izulin:
@wojciechczerniak what should be an API for extracting currency info?

wojciechczerniak:
> @wojciechczerniak what should be an API for extracting currency info?

We have `getCellValue`, `getCellValueType`, `getCellValueDetailedType` for all the thing related to cell value. So maybe (following the convention) we should name it `getCellValueFormat`? Will be ready for the future and other types formats.

izulin:
@wojciechczerniak ready for review

wojciechczerniak (changes requested):

src/HyperFormula.ts
```diff
@@ -2858,6 +2859,35 @@ export class HyperFormula implements TypedEmitter {
...
+   * ```js
+   * const hfInstance = HyperFormula.buildFromArray([
+   *  ['1$', '1'],
+   * ]);
+   *
+   * // should return '$', cell value type of provided coordinates is a number with a format inference currency, parsed as using '$' as currency.
+   * const cellType = hfInstance.getCellValueType({ sheet: 0, col: 0, row: 0 });
+   *
+   * // should return undefined, cell value type of provided coordinates is a number with no format information.
+   * const cellType = hfInstance.getCellValueType({ sheet: 0, col: 1, row: 0 });
```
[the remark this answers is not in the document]
izulin: riiiiight

docs/guide/known-limitations.md
```diff
@@ -28,7 +28,6 @@ you can't compare the arguments in a formula like this:
 * There is no relative referencing in named ranges.
 * The library doesn't offer (at least not yet) the following features:
   * 3D references
-  * Complex numbers
```
wojciechczerniak, line 31: We can remove "  * Currency data type" from the list as well? https://github.com/handsontable/hyperformula/pull/590#issuecomment-770292976

izulin: sure

src/interpreter/ArithmeticHelper.ts
```diff
@@ -695,38 +720,40 @@ function escapeNoCharacters(pattern: string, caseSensitive: boolean): string {
   return str
 }
 
-function inferExtendedNumberTypeAdditive(left: NumberType, right: NumberType): NumberType {
+function inferExtendedNumberTypeAdditive(left: NumberType, right: NumberType, leftFormat: FormatInfo, rightFormat: FormatInfo): [NumberType, FormatInfo] {
```
wojciechczerniak, line 723: I'm just wondering if it will not be better to pass the whole `rightType` and the `ExtendedNumberFactory` could choose from it what it needs. We would be ready for other properties we chould choose from in the future.

```suggestion
function inferExtendedNumberTypeAdditive(left: RichNumber, right: RichNumber): RichNumber {
  if(left === NumberType.NUMBER_RAW) {
    return right
  }
```

wojciechczerniak: Then it's up to the `ExtendedNumberFactory` to decide if there are any additional properties to choose from:
```ts
public ExtendedNumberFactory(value: number, type: RichNumber): ExtendedNumber {
  const { numberType, format, futureProp } = type;
```

wojciechczerniak: WDYT?

wojciechczerniak (changes requested):
I took another look and have few more remarks

src/interpreter/ArithmeticHelper.ts
```diff
@@ -508,6 +515,24 @@ export class ArithmeticHelper {
       return [0, val]
     }
   }
+
+  public ExtendedNumberFactory(value: number, type: NumberType, format?: FormatInfo): ExtendedNumber {
+    switch (type) {
+      case NumberType.NUMBER_RAW:
+        return value
+      case NumberType.NUMBER_CURRENCY: {
+        return new CurrencyNumber(value, format ?? this.config.currencySymbol[0])
```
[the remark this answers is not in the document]
izulin: if the `currencySymbols` are empty, then the correct semantics is `undefined`, right?

src/interpreter/ArithmeticHelper.ts
izulin: we not always have a RichNumber lying around to pass it as an example, and creating one artificially just to pass the type to the function seems.. artificial. I'll refactor it in some way into that direction.

src/interpreter/ArithmeticHelper.ts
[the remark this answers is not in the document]
wojciechczerniak: I see two options here:
- validate array length to be at least > 1, we need currency for PMV function and others
- it is possible to switch off the currency symbol and PMV will return just a number without format, but this should be done with `currencySymbols: undefined` or `null`, because doing it with an empty array `currencySymbols: []` IMO looks odd

src/interpreter/ArithmeticHelper.ts
wojciechczerniak: Right! It might be any type that have `format` property or others. But if we have to artificially pass something weird it may not be worth to do it now. We may wait for a real use case.

izulin: should be better now

src/interpreter/ArithmeticHelper.ts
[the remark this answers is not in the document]
izulin: `currencySymbols: undefined` is explicitly asking for default value, which is '$'.
IMO setting it to [] is perfectly understandable way of doing exactly what you are asking for: I want type hints with currency, but no predefined symbol for currency. `format` accepts `undefined` anyway, so I think this is WAD.

wojciechczerniak (approved):

PR #656, opened 2021-04-27
fixes to argument validation and config

### LINKED ISSUE #654: getSheetId fails with an error when called without an argument.
### Description
`hf.getSheetId` throws the `Cannot read property 'toLowerCase' of undefined`, when called without an argument.
It also fails with `sheetDisplayName.toLowerCase is not a function` if called with a number as an argument.

### Steps to reproduce

1. Call `hf.getSheetId()`...
2. ... or `hf.getSheetId(333)`

### Your environment

* HyperFormula version: latest `develop`

izulin:
### Context

Fixes #654.
Adds minor refactor of config getter.

### Types of changes

### Related issue(s):
1.
2.
3.

### Checklist:


src/ArgumentSanitization.ts
```diff
@@ -0,0 +1,85 @@
...
+    if (typeof inputValue === expectedType) {
+      return inputValue
+    } else {
+      throw new ExpectedValueOfTypeError(expectedType, paramName)
+    }
+  } else {
+    if (expectedType.includes(inputValue)) {
+      return inputValue
+    } else {
+      throw new ExpectedOneOfValuesError(expectedType.map((val: string) => '\'' + val + '\'').join(' '), paramName)
```
wojciechczerniak, line 29:
```suggestion
      throw new ExpectedOneOfValuesError(expectedType.map((val: string) => `'${val}'`).join(' '), paramName)
```

src/Config.ts
```diff
@@ -621,19 +625,33 @@ export class Config implements ConfigParams, ParserConfig {
...
+      } else {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+        // @ts-ignore
+        ret[key] = val
+      }
+    }
+    return ret as ConfigParams
   }
 
+
```
wojciechczerniak, line 654: unnecessary new line

```suggestion
```

src/Config.ts
```diff
@@ -621,19 +625,33 @@ export class Config implements ConfigParams, ParserConfig {
...
 
-  public getConfig(): ConfigParams { //TODO: avoid pollution
-    return this
+  public getConfig(): ConfigParams {
+    const ret = {}
+    for (const key in Config.defaultConfig) {
+      const val = this[key as ConfigParamsList]
+      if (Array.isArray(val)) {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+        // @ts-ignore
```
wojciechczerniak, line 643: Do we need those because of missing type for `ret`? If so, an utilty type `Partial<T>` may help. We may need to cast it to full ConfigParams later but I see that we do it anyway

```ts
const ret: Partial<ConfigParams> = {}

// ...

return ret as ConfigParams
```

test/config.spec.ts
```diff
@@ -204,3 +204,13 @@ describe('Config', () => {
     expect(() => new Config({ nullYear: 101 })).toThrowError('Config parameter nullYear should be at most 100')
   })
 })
+
+describe('getConfig', () => {
+  it('should copy returned values', () => {
```
wojciechczerniak, line 209: Can we also check that it is not an instance of `Config` just to be sure? The main requirement was to have a literal object returned

src/HyperFormula.ts
```diff
@@ -1423,8 +1463,9 @@ export class HyperFormula implements TypedEmitter {
    * Checks if it possible to reorder columns of a sheet according to a permutation.
    *
    * @param {number} sheetId - ID of a sheet to operate on
-   * @param {number[]} newColumnOrder - permutation of columns
+   * @param {number[]} newRowOrder - permutation of columns
```
wojciechczerniak, line 1466: Good catch but method is called `isItPossibleToSetColumnOrder`, TSDoc is correct, it's the argument name that is wrong

wojciechczerniak (changes requested):
Nice 👍

src/Config.ts
izulin: I don't think there is an easy way to type this.

`this[key]` is a heterogenously typed object, and `ret[key]` is also heterogenously typed, and IMO typescript has a hard time understanding that those will be of the same type.

src/Config.ts
```diff
@@ -621,17 +627,28 @@ export class Config implements ConfigParams, ParserConfig {
...
+      const val = this[key as ConfigParamsList]
+      if (Array.isArray(val)) {
+        //eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+        (ret[key as ConfigParamsList] as ConfigParamsTypes) = [ ...val ]
+      } else {
+        //eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+        (ret[key as ConfigParamsList] as ConfigParamsTypes) = val
+      }
+    }
+    return ret as ConfigParams
```
wojciechczerniak, line 651: 😬 Maybe something more generic? Those types are assigning anything to anything anyway.

```suggestion
  public getConfig(): ConfigParams {
    const ret: { [key: string]: any } = {};

    for (let [key, val] of Object.entries(Config.defaultConfig)) {
      if (Array.isArray(val)) {
        ret[key] = [ ...val ]
      } else {
        ret[key] = val
      }
    }

    return ret as ConfigParams
```

wojciechczerniak (approved):
Thanks for addressing my comments 👍

PR #665, opened 2021-05-07
range interpolation

izulin:
### Context

Adds a new feature: range interpolation. When range is provided in place of a scalar, and position of the formula vs range is special (range is width 1 and formula is in the row that intersects the range, or range is height 1 and formula is in the column that intersects the range) the scalar value can be taken from range.

### Types of changes

### Related issue(s):
1.
2.
3.

### Checklist:


test/arrays.spec.ts
```diff
@@ -182,3 +181,35 @@ describe('coercion of array to scalar', () => {
...
 
+describe('range interpolation', () => {
+  it('with function', () => {
+    const engine = HyperFormula.buildFromArray([[0, 1, 2], ['=EXP(A1:C1)']])
+    expect(engine.getCellValue(adr('A2'))).toEqual(1)
+  })
+
+  it('with binary op', () => {
+    const engine = HyperFormula.buildFromArray([[0, 1, 2], [undefined, '=A1:C1+A1:C1']])
+    expect(engine.getCellValue(adr('B2'))).toEqual(2)
```
wojciechczerniak, line 192: I don't get the `undefined`s in those tests. Why do we add complexity and shift the formula instead keeping it in A2?

wojciechczerniak (approved):

test/interpreter/operator-division.spec.ts
```diff
@@ -57,16 +57,17 @@ describe('Operator DIVISION', () => {
...
 
-  // Inconsistency with Product 1
   it('range value results in VALUE error', () => {
     const engine = HyperFormula.buildFromArray([
-      ['1', '=10 / A1:A3'],
-      ['9', '=A1:A3 / 10'],
+      ['1'],
+      ['9'],
       ['3'],
+      ['=10 / A1:A3'],
```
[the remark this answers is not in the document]
izulin: No, since those tests are in {arrays: false} mode. In {arrays: true} mode, we will compute this exactly as XL, and XL is by default in {arrays: true} mode. We should add the info about the flag to the docs somewhere.

test/arrays.spec.ts
izulin: to test that we are able to extract something from the middle of the range. Otherwise how do we know that we are not getting things by default from top-left corner of the range, always.

wojciechczerniak: ok 👍

test/interpreter/operator-division.spec.ts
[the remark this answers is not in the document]
wojciechczerniak: Should we add the `array: false` to those tests then to make it obvious and have those tests sealed in case we change `array` to default `true`?

PR #680, opened 2021-05-25
some fixes to namedexpressions

### LINKED ISSUE #319: Use Named Expressions with `buildFromArray` and `buildFromSheets` factories
### Description

Currently, the engine has to be initialized with `buildEmpty`, then Named Expressions added with API methods, and finally we can `addSheet` / `setSheetContent`.

A new config option `namedExpressions: { expressionName: string, expression: RawCellContent, scope: number, ... otherOptions }` could help. Where `scope` is the sheetId and global if missing). Depends how we're going to implement it in #240 

```ts
HyperFormula.buildFromArray([ [ ... ] ], {
  namedExpressions: [
     { expressionName: 'TRUE', expression: '=TRUE()', visible: false }
  ]
})
```

### LINKED ISSUE #335: Get all Named Expressions method needed
### Description

I've noticed that we don't have a method to get all Named Expressions. It should be roughly the same format that #319 will accept. For example `getAllNamedExpressionsSerialized`

### Use case

We can get all sheets serialized with `getAllSheetsSerialized` send them to server and load the next day. But if sheets contain Named Expressions, they will not calculate correctly anymore. There should be a method that can get all Named Expressions serialized so we can store both in the database and recreate the engine in another browser session.

```ts
const sheetsData = engine.getAllSheetsSerialized()
const namedExpressionsData = engine.getAllNamedExpressionsSerialized()

const newEngine = HyperFormula.buildFromSheets(sheetsData, {
  namedExpressions: namedExpressionsData
})
```

izulin:
### Context

Fixes #335.
Fixes #319.

### Types of changes

### Related issue(s):
1.
2.
3.

### Checklist:

### DIFF
--- a/src/HyperFormula.ts
+++ b/src/HyperFormula.ts
@@ -3650,14 +3650,14 @@ export class HyperFormula implements TypedEmitter {
    * ]);
    *
    * // add two named expressions and one scoped
-   * hfInstance.addNamedExpression('prettyName', '=Sheet1!A1+100');
-   * hfInstance.addNamedExpression('prettyName2', '=Sheet1!A2+100');
-   * hfInstance.addNamedExpression('prettyName3', '=Sheet1!A3+100', 0);
+   * hfInstance.addNamedExpression('prettyName', '=Sheet1!$A$1+100');
+   * hfInstance.addNamedExpression('anotherPrettyName', '=Sheet1!$A$2+100');
+   * hfInstance.addNamedExpression('alsoPrettyName', '=Sheet1!$A$3+100', 0);
    *
-   * // list the expressions, should return: ['prettyName', 'prettyName2'] for this example
+   * // list the expressions, should return: ['prettyName', 'anotherPrettyName'] for this example
    * const listOfExpressions = hfInstance.listNamedExpressions();
    *
-   *  // list the expressions, should return: ['prettyName3'] for this example
+   *  // list the expressions, should return: ['alsoPrettyName'] for this example
    * const listOfExpressions = hfInstance.listNamedExpressions(0);
    * ```
    *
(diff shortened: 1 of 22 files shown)


src/HyperFormula.ts
```diff
@@ -293,8 +293,8 @@ export class HyperFormula implements TypedEmitter {
    *
    * @category Factories
    */
-  public static buildEmpty(configInput?: Partial<ConfigParams>): HyperFormula {
-    return this.buildFromEngineState(BuildEngineFactory.buildEmpty(configInput))
+  public static buildEmpty(configInput: Partial<ConfigParams> = {}, namedExpressions: SerializedNamedExpression[] = []): HyperFormula {
```
wojciechczerniak, line 296: Is it a good choice for `namedExpressions`? I think that by similarity to `functionPlugins` the should be placed within Config object.

On the other hand, Config is not a user content. It may be stored separately and be a constant. Sometimes hardcoded into the app. When we look at it like that, it makes sense to separate those contexts.

@izulin You probably thought that through already, WDYT?

wojciechczerniak (changes requested):

src/HyperFormula.ts
izulin: I think it should go other way around: If there was a way to hardcode it into the actual data (like in `buildFromArray` or `buildFromSheets`), it should be. But I cannot think of a way to make it so consistently without making unnecessary layers of complexity, so I think keeping it as a separate argument should do fine.

wojciechczerniak: Yup. And having it in Config will complicate updateConfig logic. Good choice

wojciechczerniak (approved):

PR #681, opened 2021-05-25
Change the hash declaration of a private member to a local weakMap.

jansiegel:
### Context
Because of https://github.com/handsontable/handsontable/issues/8099 I tried to move the `#licenseKeyValidityState` declaration of https://github.com/handsontable/hyperformula/blob/master/src/Config.ts#L514 to a `WeakMap`, so it wouldn't produce the `#private` entry in the declaration file.

@wojciechczerniak please review

### Related issue(s):
1. https://github.com/handsontable/handsontable/issues/8099


wojciechczerniak (approved):
👍

Issue #695, opened 2021-06-09
Add README.md in /docs how to run docs


kirszenbaum:
- [x] First draft added: https://github.com/handsontable/hyperformula/blob/feature/issue-695/docs/README.md

PR #702, opened 2021-06-11
Docs: Adding the docs README.md

kirszenbaum:
- Added a README.md file into the /docs folder: https://github.com/handsontable/hyperformula/blob/feature/issue-695/docs/
- The content structure follows the handsontable docs README.md pattern, can be expanded later on
- Any feedback welcome!
#695

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,12 +1,12 @@
-# HyperFormula docs
+# HyperFormula documentation
 
 HyperFormula comes with a dedicated, regularly-updated documentation portal.
 
-View the docs' latest production version at https://handsontable.github.io/hyperformula/.
+View the documentation's latest production version at https://handsontable.github.io/hyperformula/.
 
-## About HyperFormula docs
+## About HyperFormula documentation
 
-The HyperFormula docs are built with [VuePress](https://vuepress.vuejs.org/), a Vue-powered Static Site Generator.
+The HyperFormula documentation is built with [VuePress](https://vuepress.vuejs.org/), a Vue-powered Static Site Generator.
 
 Our VuePress setup:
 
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -0,0 +1,64 @@
...
+   * Earlier than 16.0
+2. From the `hyperformula` directory, install the docs dependencies:
+    ```bash
+    npm install
+    ```
+3. From the `hyperformula` directory, start your local docs server:
+   ```bash
+   npm run docs:dev
+   ```
+4. In your browser, go to: http://localhost:8080/hyperformula/.
```
wojciechczerniak, line 34: Step missing: `npm run docs:api` before starting the server. Generate API reference from the HF source code.

And HF should be already build in the `/dist` directory: `npm bundle-all` I think. Otherwise you will get an error that `hyperformula.min.js` file was not found.

docs/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+The HyperFormula documentation is built with [VuePress](https://vuepress.vuejs.org/), a Vue-powered Static Site Generator.
+
+Our VuePress setup:
+
+* Publishes the docs as a sleek, responsive single-page application (SPA).
+* Boasts a great search engine.
+* Lets us write the docs in Markdown (and use Vue inside Markdown!).
+* Lets us quickly build the docs output onto a local server.
+* [COMING SOON] Versions and deploys the docs together with the software.
```
wojciechczerniak, line 17: IMO we should not describe VuePress, for me it's an internal implementation detail, and we may change the engine behind the docs in the future. But, we can have a short mention because some editing features are based on VuePress capabilities so the editor may use features described here https://vuepress.vuejs.org/guide/markdown.html

wojciechczerniak (changes requested):

wojciechczerniak (approved):
Thanks!

PR #708, opened 2021-06-12
Array CRUDs improvements (part #1)

voodoo11:
### Context

This is a first part of improvements of CRUD operations with array formulas.

1. Transpose no longer needs to be limited to a numeric type.
2. Allow to add row across array. Array size won't be affected.
3. Allow to remove row across array. Array size won't be affected.
4. Allow to change content inside array. SPILL will be returned if changing value other that left corner

Splitted, as it is already huge PR. Should be merged with second part (#732). 

### Types of changes

### Related issue(s):
1.
2.
3.

### Checklist:


src/ContentChanges.ts
```diff
@@ -54,19 +52,27 @@ export class ContentChanges {
...
+  }
+
+  private add(address: SimpleCellAddress, change: CellValueChange) {
+    const value = change.value
+    if (value instanceof SimpleRangeValue) {
+      for (const cellAddress of value.effectiveAddressesFromData(address)) {
+        this.changes.delete(`${cellAddress.sheet},${cellAddress.col},${cellAddress.row}`)
+      }
+    }
+    this.changes.set(addressKey((address)), change)
```
izulin, line 69: I have not seen how it is used, but: what if first `SimpleRangeValue` is inserted and later a scalar value?
Or maybe this scenario does not happen?

src/DependencyGraph/AddressMapping/AddressMapping.ts
```diff
@@ -98,6 +98,25 @@ export class AddressMapping {
...
   }
 
+  public moveCell(source: SimpleCellAddress, destination: SimpleCellAddress) {
+    const sheetMapping = this.mapping.get(source.sheet)
+    if (!sheetMapping) {
+      throw Error('Sheet not initialized')
+    }
+    if (source.sheet !== destination.sheet) {
+      throw Error('Cannot move cells between sheets.')
+    }
```
izulin, line 108: what is the rationale behind function that moves cell, but only inside the same sheet? how do we handle other moves?

src/DependencyGraph/DependencyGraph.ts
```diff
@@ -81,12 +84,9 @@ export class DependencyGraph {
     this.graph = new Graph<Vertex>(this.dependencyQueryVertices)
   }
 
-  public setFormulaToCell(address: SimpleCellAddress, ast: Ast, dependencies: CellDependency[], size: MatrixSize, hasVolatileFunction: boolean, hasStructuralChangeFunction: boolean) {
+  public setFormulaToCell(address: SimpleCellAddress, ast: Ast, dependencies: CellDependency[], size: MatrixSize, hasVolatileFunction: boolean, hasStructuralChangeFunction: boolean): ContentChanges {
     const newVertex = FormulaVertex.fromAst(ast, address, size, this.lazilyTransformingAstService.version())
     this.exchangeOrAddFormulaVertex(newVertex)
-
-    this.addressMapping.setCell(address, newVertex)
-
```
izulin, line 89: how is this handled now?

src/DependencyGraph/DependencyGraph.ts
```diff
@@ -125,15 +128,15 @@ export class DependencyGraph {
...
 
     this.correctInfiniteRangesDependency(address)
+
+    return this.getAndClearContentChanges()
   }
 
-  public setCellEmpty(address: SimpleCellAddress) {
-    const vertex = this.addressMapping.getCell(address)
+  public setCellEmpty(address: SimpleCellAddress): ContentChanges {
+    const vertex = this.shrinkPossibleMatrixAndGetCell(address)
```
izulin, line 136: what if vertex is already `null`, why do we then shrink matrix that contains it?

src/DependencyGraph/DependencyGraph.ts
```diff
@@ -125,15 +128,15 @@ export class DependencyGraph {
...
     if (vertex === null) {
-      return
+      return ContentChanges.empty()
```
izulin, line 138: especially that we assume that there are no changes, but could matrix shrinking generate content changes?

src/interpreter/SimpleRangeValue.ts
```diff
@@ -134,6 +143,7 @@ export class SimpleRangeValue {
   }
 
   public rawData(): InternalScalarValue[][] {
+    this.ensureThatComputed()
```
izulin, line 146: return this._data ?? [] can be changed to return this._data!

izulin (approved):

src/DependencyGraph/DependencyGraph.ts
voodoo11: inside `exchangeOrAddFormulaVertex`

src/ContentChanges.ts
voodoo11: This scenario should not happen. It can be handled nicer but at this moment there are to much conflicts in places not related to this PR.

src/DependencyGraph/AddressMapping/AddressMapping.ts
voodoo11: It moves array cells in address mapping after adding rows or columns and right now it has nothing to do with `moveCells` public api. Maybe I should change name of this function.

test/cruds/adding-row.spec.ts
```diff
@@ -337,3 +306,119 @@ describe('Adding row - column index', () => {
...
+
+  it('adding row across array should not change array', () => {
+    const engine = HyperFormula.buildFromArray([
+      [], [], [],
+      ['=-A1:B3'],
+      [], [],
+      ['foo']
+    ], {useArrayArithmetic: true})
+
+    engine.addRows(0, [4, 1])
```
wojciechczerniak, line 340: Is it "across array"? Looks like after an array, and after the formula.

In excel when we add row within the range it will expand

wojciechczerniak: Maybe we miss one test, something like:
```suggestion
  it('adding row across array should not change array', () => {
    const engine = HyperFormula.buildFromArray([
      [], [], [],
      ['=-A1:B3'],
      [], [],
      ['foo']
    ], {useArrayArithmetic: true})

    engine.addRows(0, [2, 1])
```

wojciechczerniak (changes requested):

test/cruds/adding-row.spec.ts
voodoo11: >Is it "across array"? 

It is. Above A1:B3 there is 3 rows.

> In excel when we add row within the range it will expand

But I do not add row within a range but across spilled array. And spilled array is never expanded that way.

test/cruds/adding-columns.spec.ts
```diff
@@ -203,7 +203,8 @@ describe('Adding column - FormulaCellVertex#address update', () => {
   })
 })
 
-describe('Adding column - MatrixVertex', () => {
+/* TODO size of matrix can be dependent on structural changes */
+xdescribe('Adding column - MatrixVertex', () => {
```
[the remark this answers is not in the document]
voodoo11: Yes, it is handled in second part.

test/cruds/adding-row.spec.ts
wojciechczerniak: Ooohh, across spilled array, right! Now makes sense to me

test/cruds/adding-columns.spec.ts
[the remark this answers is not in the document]
wojciechczerniak: 👍

src/DependencyGraph/FormulaCellVertex.ts
```diff
@@ -151,6 +160,11 @@ export class MatrixVertex extends FormulaVertex {
     }
   }
 
+  setNoSpace(): InterpreterValue {
+    this.matrix = new ErroredMatrix(new CellError(ErrorType.REF, ErrorMessage.NoSpaceForArrayResult), MatrixSize.error())
```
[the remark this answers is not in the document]
voodoo11: > Shouldn't this be a SPILL error type?

GS would return REF and that's what I followed. Do we want to introduce new error type at this moment? 

> Can we provide some additional info to the error, as width and height that would allow to display it as shown on the screenshot?

It would be a nice feature indeed but right now we do not store any additional data inside errors except address of a root of a problem.  I would save it for the future.

src/HyperFormula.ts
```diff
@@ -1596,7 +1594,6 @@ export class HyperFormula implements TypedEmitter {
    * Returns information whether it is possible to remove rows from a specified position in a given sheet.
    * Checks against particular rules to ascertain that removeRows can be called.
    * If returns `true`, doing [[removeRows]] operation won't throw any errors.
-   * Returns `false` if the operation might be disrupted and causes side-effects by the fact that there is a matrix inside the selected rows.
```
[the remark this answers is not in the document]
voodoo11: Yes, you are right. First case is for exceeding sheet size (for adding rows), but also the arguments of the function may not make sense. I've fixed tsdocs.

src/DependencyGraph/FormulaCellVertex.ts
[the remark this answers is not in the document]
wojciechczerniak: > > Shouldn't this be a SPILL error type?
> 
> GS would return REF and that's what I followed. Do we want to introduce new error type at this moment?

Yes, please. It will be an additional information that helps to differentiate them. ODFF allows vendors to introduce custom errors.
 
> > Can we provide some additional info to the error, as width and height that would allow to display it as shown on the screenshot?
> 
> It would be a nice feature indeed but right now we do not store any additional data inside errors except address of a root of a problem. I would save it for the future.

Ok 👍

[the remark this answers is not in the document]
voodoo11: Done.

wojciechczerniak (approved):
Awesome, thank you 🏆

izulin (approved):

Issue #557, opened 2020-11-02
Improve Custom Functions page


wojciechczerniak:
### Description
Since we've created custom functions tutorial there were some changes. Custom functions have now arguments validation, extended meta-data and `runFunction` callback. Maybe we can keep the basic (if it still works correctly) and need an Advanced Custom Functions page. How to create function alias.

BTW. There's an error
![Screenshot 2020-11-02 at 11 35 07](https://user-images.githubusercontent.com/4718998/97858563-9e1ef800-1cff-11eb-861b-62209fa06830.png)
![Screenshot 2020-11-02 at 11 43 28](https://user-images.githubusercontent.com/4718998/97859295-b5121a00-1d00-11eb-9f40-3855e7ff42fb.png)

### Links
https://handsontable.github.io/hyperformula/guide/custom-functions.html

### Type of issue

- [x] Missing documentation
- [x] Error in the documentation
- [ ] Request for a demo
- [ ] Request for a tutorial, guide etc.
- [ ] Other issues, suggestions or ideas

krzysztofspilka:
@wojciechczerniak  Absolutely, we should do that. However, a custom function is already an advanced feature of the engine, thus no need to limit the new additions to so-called "advanced users".

I think we just need to add another section to the existing page, and that's it.

kirszenbaum:
Self-note - take this into account: https://github.com/handsontable/hyperformula/discussions/734#discussioncomment-925414

kirszenbaum:
- [x] Fixed errors depicted in the screenshots

PR #758, opened 2021-07-07
Docs: expanding the Custom Functions topic

kirszenbaum:
#557 

In [Custom Functions](https://handsontable.github.io/hyperformula/guide/custom-functions.html):
- [x] Added an "Optional parameters" section, based on [this](https://github.com/handsontable/hyperformula/blob/develop/src/interpreter/plugin/FunctionPlugin.ts#L41-L90)
- [x] Added an "Argument validation options" section, based on [this](https://github.com/handsontable/hyperformula/blob/develop/src/interpreter/plugin/FunctionPlugin.ts#L147-L186)
- [x] Added a "runFunction()" section
- [x] Added a "Throwing errors" section, based on [this](https://github.com/handsontable/hyperformula/discussions/734)


docs/guide/custom-functions.md
```diff
@@ -140,16 +220,72 @@ the calculations.
...
+
+  public hyper(ast, state) {
+    this.runFunction(HYPER())
+    
+    return 'Hyperformula'.length
+    }
+  };
+```
+
+### Throwing errors
```
wojciechczerniak, line 260: We're not throwing errors here. We're returning them. And they are correctly handled and displayed with translation.

docs/guide/custom-functions.md
```diff
@@ -140,16 +220,72 @@ the calculations.
...
+```javascript
+// import `CellError` and `ErrorType`
+import { FunctionPlugin, CellError, ErrorType } from "hyperformula";
+
+public hyper({ args }) {
+    if (!args.length) {
+      // create a `CellError` instance with an `ErrorType` of `ERROR`, and your custom error message
+      const error = new CellError(ErrorType.ERROR, 'Not enough arguments!');
+      
+      return error;
```
wojciechczerniak, line 275: We can simplify this line
```suggestion
      return new CellError(ErrorType.ERROR, 'Not enough arguments!');
```

docs/guide/custom-functions.md
```diff
@@ -140,16 +220,72 @@ the calculations.
...
+For example, if you want to throw a custom error message when the user doesn't pass enough arguments to your function, use the [`CellError` class](../api/classes/cellerror.md):
+
+```javascript
+// import `CellError` and `ErrorType`
+import { FunctionPlugin, CellError, ErrorType } from "hyperformula";
+
+public hyper({ args }) {
+    if (!args.length) {
+      // create a `CellError` instance with an `ErrorType` of `ERROR`, and your custom error message
+      const error = new CellError(ErrorType.ERROR, 'Not enough arguments!');
```
wojciechczerniak, line 273: Also IMO we can choose more iconic error type so devs will recognize we're referring to spreadsheets cell errors here

```suggestion
      const error = new CellError(ErrorType.DIV_BY_ZERO, 'Not enough arguments!');
```

And add a description that it will be displayed as `#DIV/0!` and will be correctly translated

docs/guide/custom-functions.md
```diff
@@ -140,16 +220,72 @@ the calculations.
...
+      }
+    }
+  };
+
+  public hyper(ast, state) {
+    this.runFunction(HYPER())
+    
+    return 'Hyperformula'.length
+    }
+  };
```
wojciechczerniak, line 257: Formatting
```suggestion
  }
};
```

docs/guide/custom-functions.md
kirszenbaum: @wojciechczerniak - done: 304dfa0

docs/guide/custom-functions.md
kirszenbaum: @wojciechczerniak - done: 304dfa0

docs/guide/custom-functions.md
```diff
@@ -140,16 +220,72 @@ the calculations.
...
+    }
+    
+    else {
+      return 'Hyperformula'.length;
+    }
+  };
+```
+
+The error displays as `#DIV/0`, and gets properly translated.
+
```
wojciechczerniak, line 283: I've just noticed that my suggestion removed the optional error message. Maybe we should mention that all error types support optional string message that we encourage to give users more details on the reasons for error and how they can avoid it?

Also, here is a good place to mention that errors will be returned by ie getCellValue methods but wrapped in DetailedCellError type that will translate those errors according to i18n settings. I think that would be a full 101 on errors then

kirszenbaum: @wojciechczerniak - done: 11d9bed

docs/guide/custom-functions.md
```diff
@@ -46,25 +71,79 @@ import { FunctionPlugin } from 'hyperformula';
...
+| Option | Type | Description |
+| --- | --- | --- |
+| `arrayFunction` | Boolean | If set to `true`, the function enables the [array arithmetic mode](arrays.md) in its arguments and nested expressions. |
+| `doesNotNeedArgumentsToBeComputed` | Boolean | If set to `true`, the function treats reference or range arguments as arguments that don't create dependency.<br><br>Other arguments are properly evaluated. |
+| `expandRanges` | Boolean | If set to `true`, ranges in the function's arguments are inlined to (possibly multiple) scalar arguments. |
+| `isDependentOnSheetStructureChange` | Boolean | If set to `true`, the function gets recalculated with each sheet shape change. |
+| `isVolatile` | Boolean | If set to `true`, the function is [volatile](volatile-functions.md). |
+| `repeatLastArgs` | Number | For functions with a variable number of arguments: sets how many last arguments can be repeated indefinitely. |
+| `vectorizationForbidden` | Boolean | If set to `true`, prevents the function from ever being vectorized (but it's up to your function implementation to properly handle vectorization). |
```
wojciechczerniak, line 102: We do not explain what does the vectorization means and what will happen if it's not handled by user 🤔 or how to do it.

docs/guide/custom-functions.md
```diff
@@ -46,25 +71,79 @@ import { FunctionPlugin } from 'hyperformula';
...
+```
+
+You can set the following optional parameters:
+
+| Option | Type | Description |
+| --- | --- | --- |
+| `arrayFunction` | Boolean | If set to `true`, the function enables the [array arithmetic mode](arrays.md) in its arguments and nested expressions. |
+| `doesNotNeedArgumentsToBeComputed` | Boolean | If set to `true`, the function treats reference or range arguments as arguments that don't create dependency.<br><br>Other arguments are properly evaluated. |
+| `expandRanges` | Boolean | If set to `true`, ranges in the function's arguments are inlined to (possibly multiple) scalar arguments. |
+| `isDependentOnSheetStructureChange` | Boolean | If set to `true`, the function gets recalculated with each sheet shape change. |
```
wojciechczerniak, line 99: Maybe we should mention defaults? Does all our functions are recalculated on sheet shape change? What does the "shape change" mean? I guess, adding/removing rows or cols

docs/guide/custom-functions.md
```diff
@@ -46,25 +71,79 @@ import { FunctionPlugin } from 'hyperformula';
...
-each change of the shape of the engine sheets. 
-`doesNotNeedArgumentsToBeComputed` marks functions that treat references or ranges in their arguments
-as arguments that do not create dependency. Other arguments are properly evaluated.
-`arrayFunction` denotes functions that enable array arithmetic in its arguments and nested expressions.
-`vectorizationForbidden` when set prevents function from ever being vectorized (however, it is up to implementation of a function to properly handle vectorization).
+You can set the following argument validation options:
+
+| Parameter | Type | Description |
+| --- | --- | --- |
+| `passSubtype` | Boolean | If set to `true`, arguments need to be passed with full type information.<br>(e.g. for numbers: `Date` or `DateTime` or `Time` or `Currency` or `Percentage`) |
```
wojciechczerniak, line 139: I believe they don't "need to", more like they are. It's a setting for the evaluator to provide them / it means we're requesting types @izulin ?

docs/guide/custom-functions.md
```diff
@@ -140,16 +220,83 @@ the calculations.
...
+      parameters: {
+        passSubtype: false
+      }
+    }
+  };
+
+  public hyper(ast, state) {
+    this.runFunction(HYPER())
+    
+    return 'Hyperformula'.length
```
wojciechczerniak, line 255: @izulin should we return `runFunction` results? I think it parses args, and vectorizes or sth the results

```suggestion
    return this.runFunction(() => 'Hyperformula'.length )
```

docs/guide/custom-functions.md
```diff
@@ -140,16 +220,83 @@ the calculations.
 // arguments here are displayed just to show the structure
 public hyper(ast, state) {
     return 'Hyperformula'.length;
+  };
+```
+
+### `runFunction()`
+
+To test your function's configuration, run the built-in `runFunction()` function.
```
wojciechczerniak, line 228: `runFunction` will validate arguments according to configuration but from this sentence and example below it looks like it's the only purpose. As the name indicates, it need to run the custom function. So it wraps your calculations, prepares the arguments and then parses results. 

But I'm not sure if I have all the details right, CC @izulin

docs/guide/custom-functions.md
```diff
@@ -46,25 +71,79 @@ import { FunctionPlugin } from 'hyperformula';
...
-as arguments that do not create dependency. Other arguments are properly evaluated.
-`arrayFunction` denotes functions that enable array arithmetic in its arguments and nested expressions.
-`vectorizationForbidden` when set prevents function from ever being vectorized (however, it is up to implementation of a function to properly handle vectorization).
+You can set the following argument validation options:
+
+| Parameter | Type | Description |
+| --- | --- | --- |
+| `passSubtype` | Boolean | If set to `true`, arguments need to be passed with full type information.<br>(e.g. for numbers: `Date` or `DateTime` or `Time` or `Currency` or `Percentage`) |
+| `defaultValue` | `InternalScalarValue` \| `RawScalarValue` | If an argument is missing, its value defaults to `defaultValue`. |
+| `optionalArg` | Boolean | If set to `true`: if an argument is missing, and no `defaultValue` is set, the argument defaults to `undefined` (instead of throwing an error).<br><br>This is logically equivalent to setting `defaultValue` to `undefined`. |
```
wojciechczerniak, line 141: Those are interesting subject for a subsection. Why we have both options? Can we have optionalArg between required ones? Should arguments in the middle have default value then? Can we run function with empty args like `=LPV(1, , , 12)`?

wojciechczerniak (changes requested):

docs/guide/custom-functions.md
kirszenbaum: @wojciechczerniak - that's right (this bit is taken from an old description of this option).

I [added a link](https://github.com/handsontable/hyperformula/pull/758/commits/195c9a15db173963ecb6222dcd2193a07860349b#diff-aeaf0b9b496edad5d31a7315f0be860167daf08de1368a16e6d4b7f2bf81badeR102) to the only place where vectorization is mentioned so far, a [section](https://github.com/handsontable/hyperformula/blob/feature/issue-740/docs/guide/arrays.md#passing-arrays-to-scalar-functions) in the new "Arrays" topic.

I'm guessing that to properly describe vectorization, we'd need a separate section in the Arrays topic, or even a completely separate topic. Do we want to do that now? I guess it depends on what possible implications are - if it's vital to "properly handle vectorization" when implemenitng any sort of custom function, we should add it now, if not, we could do it later on

docs/guide/custom-functions.md
kirszenbaum: @wojciechczerniak 

Shape change: done edf99e9

Defaults: @izulin what are the above options' default values? (I'm guessing `false` for all booleans)

docs/guide/custom-functions.md
kirszenbaum: @wojciechczerniak - updated: 1d1bada

docs/guide/custom-functions.md
kirszenbaum: @wojciechczerniak - updated: 1d1bada

docs/guide/custom-functions.md
kirszenbaum: @izulin can you help with the above questions? Thanks!

docs/guide/custom-functions.md
wojciechczerniak: No, we don't have to. It's a lot better than outdated custom functions page and will be much more helpful even without all the possible details.

izulin: It is tricky -- `vectorizationForbidden` is an internal flag, that is used by our internal wrapper function. Any external function is free to ignore it.

izulin: (in fact, right now external functions cannot return arrays correctly)

docs/guide/custom-functions.md
izulin: Yes, this is handled by the evaluator. So external function with this flag set to true will receive wrapped numbers instead of just numbers.

docs/guide/custom-functions.md
izulin: yes, but `runFunction` takes more args

docs/guide/custom-functions.md
izulin: Yes, `runFunction` does all the heavy lifting of argument sanitization, output sanitization, vectorization, flag handling etc.

docs/guide/custom-functions.md
izulin: see example usage:

```
  public isodd(ast: ProcedureAst, state: InterpreterState): InterpreterValue {
    return this.runFunction(ast.args, state, this.metadata('ISODD'),
      (val) => (val % 2 === 1)
    )
  }
```

docs/guide/custom-functions.md
izulin: both options are to handle varying number of args:

setting `defaultValue` to any value makes the argument `optionalArg` automatically.
`optionalArg` is there for situations where we really do not want to set `defaultValue` (the latter automatically supplies defaultValue to the function, so the function is not aware of actual number of true arguments provided), e.g. function has two different behaviours depending on number of arguments

of course those flags are from `runFunction` wrapper

izulin: > Can we have optionalArg between required ones? 

yes, but that would give no effect

> Should arguments in the middle have default value then? 
> Can we run function with empty args like `=LPV(1, , , 12)`?

No, this would result in `=LPV(1, null, null, 12)` with `null` being the internal `emptyValue`, which will be coerced to `0`, `""` or any other equivalent of `null`.

docs/guide/custom-functions.md
kirszenbaum: @wojciechczerniak - updated: f023154

docs/guide/custom-functions.md
kirszenbaum: @wojciechczerniak - done: e916b01

docs/guide/custom-functions.md
```diff
@@ -46,25 +71,89 @@ import { FunctionPlugin } from 'hyperformula';
...
+| `minValue` | Number | If set, numerical arguments need to be greater than or equal to `minValue`. |
+| `maxValue` | Number | If set, numerical arguments need to be less than or equal to `maxValue`. |
+| `lessThan` | Number | If set, numerical argument need to be less than `lessThan`. |
+| `greaterThan` | Number | If set, numerical argument need to be greater than `greaterThan`. |
+
+#### Handling missing arguments
+
+The `defaultValue` and `optionalArg` options let you decide what happens when a user doesn't pass enough valid arguments to your custom function.
+
+When you set `defaultValue` to any value, the `optionalArg` setting doesn't matter.
```
wojciechczerniak, line 151: Should we say it explicitly that `defaultValue` always makes it an optional arg?

docs/guide/custom-functions.md
```diff
@@ -140,16 +230,81 @@ the calculations.
...
+public hyper({ args }) {
+    if (!args.length) {
+      // create a `CellError` instance with an `ErrorType` of `DIV_BY_ZERO`
+      // with your custom error message (optional)
+      return new CellError(ErrorType.DIV_BY_ZERO, 'Sorry, cannot divide by zero!');
+    }
+    
+    else {
+      return 'Hyperformula'.length;
+    }
```
wojciechczerniak, line 291: `else` is not necessary here as we return error above. Aloe we've lost `runFunction` from the previous step
```suggestion
      return this.runFunction(() => 'Hyperformula'.length)
```

docs/guide/custom-functions.md
```diff
@@ -164,7 +319,6 @@ export class CountHF extends FunctionPlugin {
     return 'Hyperformula'.length
     }
   };
```
wojciechczerniak, line 321: Excessive bracket is removed but indentation should be fixed as well
```suggestion
    return 'Hyperformula'.length
  }
};
```

docs/guide/custom-functions.md
```diff
@@ -84,16 +173,17 @@ class CountHF extends FunctionPlugin {
 // define functions inside this plugin
   public static implementedFunctions = {
     'HYPER': {
-    // this method's functionality will be defined in the next step
+      // this method's functionality will be defined below
       method: 'hyper',
     }
   };
   public static aliases = {
```
wojciechczerniak, line 180:
```suggestion
  };

  public static aliases = {
```

docs/guide/custom-functions.md
```diff
@@ -84,16 +173,17 @@ class CountHF extends FunctionPlugin {
 // define functions inside this plugin
   public static implementedFunctions = {
```
wojciechczerniak, line 174:
```suggestion
  // define functions inside this plugin
  public static implementedFunctions = {
```

wojciechczerniak (changes requested):

docs/guide/custom-functions.md
```diff
@@ -46,25 +71,89 @@ import { FunctionPlugin } from 'hyperformula';
...
+| --- | --- | --- |
+| `arrayFunction` | Boolean | If set to `true`, the function enables the [array arithmetic mode](arrays.md) in its arguments and nested expressions. |
+| `doesNotNeedArgumentsToBeComputed` | Boolean | If set to `true`, the function treats reference or range arguments as arguments that don't create dependency.<br><br>Other arguments are properly evaluated. |
+| `expandRanges` | Boolean | If set to `true`, ranges in the function's arguments are inlined to (possibly multiple) scalar arguments. |
+| `isDependentOnSheetStructureChange` | Boolean | If set to `true`, the function gets recalculated with each sheet shape change (e.g. when adding/removing rows or columns). |
+| `isVolatile` | Boolean | If set to `true`, the function is [volatile](volatile-functions.md). |
+| `repeatLastArgs` | Number | For functions with a variable number of arguments: sets how many last arguments can be repeated indefinitely. |
+| `vectorizationForbidden` | Boolean | If set to `true`, prevents the function from ever being [vectorized](arrays.md#passing-arrays-to-scalar-functions) (but it's up to your function implementation to properly handle vectorization). |
```
wojciechczerniak, line 102: We should remove this entry as stated by @izulin https://github.com/handsontable/hyperformula/pull/758#discussion_r666980119 that it's internal flag

docs/guide/custom-functions.md
```diff
@@ -140,16 +230,81 @@ the calculations.
...
+    if (!args.length) {
+      // create a `CellError` instance with an `ErrorType` of `DIV_BY_ZERO`
+      // with your custom error message (optional)
+      return new CellError(ErrorType.DIV_BY_ZERO, 'Sorry, cannot divide by zero!');
+    }
+    
+    else {
+      return 'Hyperformula'.length;
+    }
+  };
```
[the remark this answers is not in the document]
kirszenbaum: @wojciechczerniak - sure. I think I'd rather display the whole example (except optional parameters) in each step, WDYT? I also checked the indendation in all the code samples: 5b0f149

docs/guide/custom-functions.md
```diff
@@ -140,16 +230,81 @@ the calculations.
...
+
+#### Error localization
+
+Errors returned by methods such as `getCellValue` are wrapped in the [`DetailedCellError` type](../api/classes/detailedcellerror.md).
+
+`DetailedCellError` localizes the error based on your [internationalization settings](localizing-functions.md).
+
 ## A complete example of the class definition
 
 To sum up, here is a complete example of a custom `CountHF` class:
```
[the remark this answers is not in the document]
kirszenbaum: @wojciechczerniak - done: 5b0f149

@jansiegel - can we update the [demo](https://codesandbox.io/s/github/handsontable/hyperformula-demos/tree/1.0.x/custom-functions?from-embed=&file=/src/countHFFunctionPlugin.js) to include:
* [Optional parameters](https://github.com/handsontable/hyperformula/blob/feature/issue-557/docs/guide/custom-functions.md#optional-parameters)
* [Argument validation options](https://github.com/handsontable/hyperformula/blob/feature/issue-557/docs/guide/custom-functions.md#argument-validation-options)
* [Aliases](https://github.com/handsontable/hyperformula/blob/feature/issue-557/docs/guide/custom-functions.md#aliases)
* [Translations](https://github.com/handsontable/hyperformula/blob/feature/issue-557/docs/guide/custom-functions.md#translations)
* [The `runFunction()` wrapper](https://github.com/handsontable/hyperformula/blob/feature/issue-557/docs/guide/custom-functions.md#runfunction)
* [Returning errors](https://github.com/handsontable/hyperformula/blob/feature/issue-557/docs/guide/custom-functions.md#returning-errors)

Thanks for your help!

docs/guide/custom-functions.md
kirszenbaum: @wojciechczerniak - done: f4313f1

docs/guide/custom-functions.md
kirszenbaum: @wojciechczerniak - done: d984703

wojciechczerniak:
With translations added as static key on the plugin class the "using plugin" section have to change:

```diff
import Hyperformula, { plPL } from 'hyperformula';
import { CountHF } from './file_with_your_custom_function';
- import { myTranslations } from '/myTranslationFile';

// register the language
HyperFormula.registerLanguage('plPL', plPL);

// register your custom plugin and the translation
- HyperFormula.registerFunctionPlugin(CountHF, myTranslations);
+ HyperFormula.registerFunctionPlugin(CountHF, CountHF.translations);

// build HF instance where you can use the function directly
const hfInstance = HyperFormula.buildFromArray([['=HAJPER()']]);

// read the value of cell A1
const A1Value = hfInstance.getCellValue({ sheet: 0, col: 0, row: 0 });

// open the browser's console to see the results
console.log(A1Value);
```

docs/guide/custom-functions.md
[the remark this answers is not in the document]
wojciechczerniak: 🤔 With translations and aliases it's a little bit harder to find what has changed between steps. Also the "complete" examples is a copy of last step without any changes and this sections makes little sense. But I don't have a better idea right now

kirszenbaum:
@wojciechczerniak - translation updated: a1f532b, d786b8a

docs/guide/custom-functions.md
kirszenbaum: @wojciechczerniak - done: 3776f72

wojciechczerniak (approved):
🏆

PR #786, opened 2021-07-28
Pu/destructors

izulin:
### Context

Introduces destructors for basic components of the project.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature or improvement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):
1. #775 
2.
3.

### Checklist:

- [x] My code follows the code style of this project,
- [ ] My change requires a change to the documentation,
- [ ] I described the modification in the CHANGELOG.md file.

PR #788, opened 2021-07-31
simple destroy

izulin:
### Context

Simpler version of #786 

### Types of changes

### Related issue(s):
1.
2.
3.

### Checklist:

### DIFF
--- a/src/HyperFormula.ts
+++ b/src/HyperFormula.ts
@@ -4285,13 +4285,10 @@ export class HyperFormula implements TypedEmitter {
    * @category Instance
    */
   public destroy(): void {
-    this.dependencyGraph.destroy()
-    this.columnSearch.destroy()
-    this.evaluator.destroy()
-    this._parser.destroy()
-    this._lazilyTransformingAstService.destroy()
-    this._stats.destroy()
-    this._crudOperations.clearClipboard()
+    for(const key of Object.keys(this)) {
+      delete (this as Record<string, any>)[key]
+    }
+    Object.setPrototypeOf(this, null)
   }
 
   /**
--- a/src/HyperFormula.ts
+++ b/src/HyperFormula.ts
@@ -31,6 +31,7 @@ import {
   SheetMapping,
   Vertex,
 } from './DependencyGraph'
+import {objectDestroy} from './Destroy'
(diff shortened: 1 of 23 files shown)


src/interpreter/Interpreter.ts
```diff
@@ -281,12 +281,6 @@ export class Interpreter {
     return this.gpu
   }
 
-  public destroy() {
-    if (this.gpu) {
-      this.gpu.destroy()
```
wojciechczerniak, line 286: I'm not sure about removing this one. GPU.js docs say that we should run a proper cleanup: https://github.com/gpujs/gpu.js/#cleanup

src/HyperFormula.ts
```diff
@@ -4288,13 +4288,10 @@ export class HyperFormula implements TypedEmitter {
...
-    this.columnSearch.destroy()
-    this.evaluator.destroy()
-    this._parser.destroy()
-    this._lazilyTransformingAstService.destroy()
-    this._stats.destroy()
-    this._crudOperations.clearClipboard()
+    for(const key of Object.keys(this)) {
+      delete (this as Record<string, any>)[key]
+    }
+    Object.setPrototypeOf(this, null)
```
wojciechczerniak, line 4294: This will end up with an error being thrown `xxx is not a function` if someone will try to call methods on a destroyed instance. It is a good solution but maybe HOT has a better example. As @budnix pointed we replace all the functions with `postMortem` method: https://github.com/handsontable/handsontable/blob/b0a09bde9a9d4164bee4f28cd047fa843f42a7a9/src/core.js#L4006-L4016

And the `postMortem` method displays a warning:
https://github.com/handsontable/handsontable/blob/b0a09bde9a9d4164bee4f28cd047fa843f42a7a9/src/core.js#L4043-L4047

wojciechczerniak (changes requested):

src/interpreter/Interpreter.ts
izulin: ok, this should be then delegated to the main-class `destroy`.

src/HyperFormula.ts
izulin: good idea

wojciechczerniak (approved):
Thank you!

PR #790, opened 2021-07-31
Pu/isformula

### LINKED ISSUE #781: Array formulas should either be classed as formula & array type or not.
### Description
There's a mismatch in functions for the cells with an ARRAY type that are also formulas.

See here:

image

`doesCellHaveFormula()` returns false.
But `getSheetFormulas()` returns it as part of it's results.

This is mis-leading to the user. It should either be classes as a formula or not but not in between.

### Steps to reproduce
```js
const sheetName = hf.addSheet(state.currentSheet);
const sheetId = hf.getSheetId(sheetName);

// Fill the HyperFormula sheet with data.
hf.setSheetContent(sheetId, [[0, 1]]);
hf.setCellContents(
  { sheet: sheetId, row: 0, col: 2 },
  "=ARRAYFORMULA(ISEVEN(A1:A2*10))"
);

console.log(hf.doesCellHaveFormula({ sheet: sheetId, row: 0, col: 2 }));
console.log(hf.getSheetFormulas(sheetId));
```

### Demo
https://codesandbox.io/s/fervent-cannon-voj87?file=/src/hyperformulaConfig.js

### Your environment

* HyperFormula version: 1.0.0
* Browser Name and version: Chrome latest
* Operating System: Windows 10

izulin:
### Context

Fixes #781 

### Types of changes

### Related issue(s):
1.
2.
3.

### Checklist:

### DIFF
--- a/src/Cell.ts
+++ b/src/Cell.ts
@@ -14,6 +14,7 @@ import {
   NumberType,
 } from './interpreter/InterpreterValue'
 import {SimpleRangeValue} from './interpreter/SimpleRangeValue'
+import {Maybe} from './Maybe'
 import {CellAddress} from './parser'
 import {AddressWithSheet} from './parser/Address'
 import {FormulaVertex} from './DependencyGraph/FormulaCellVertex'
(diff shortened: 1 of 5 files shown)


src/Cell.ts
```diff
@@ -54,18 +55,23 @@ export enum CellType {
   VALUE = 'VALUE',
   ARRAY = 'ARRAY',
   EMPTY = 'EMPTY',
+  ARRAYFORMULA = 'ARRAYFORMULA',
```
wojciechczerniak, line 58: Pls add it on docs page as well: https://github.com/handsontable/hyperformula/blob/master/docs/guide/types-of-values.md

test/engine.spec.ts
```diff
@@ -556,7 +556,7 @@ describe('#getCellType', () => {
   it('formula matrix', () => {
     const engine = HyperFormula.buildFromArray([['=TRANSPOSE(C1:C2)']])
 
-    expect(engine.getCellType(adr('A1'))).toBe(CellType.ARRAY)
+    expect(engine.getCellType(adr('A1'))).toBe(CellType.ARRAYFORMULA)
     expect(engine.getCellType(adr('B1'))).toBe(CellType.ARRAY)
   })
```
wojciechczerniak, line 561: Can we add test with an array that is not a result of a formula? With `{1,2,3;4,5,6}` we will have `CellType.ARRAY` returned for all cells, A1 included, right?

wojciechczerniak (changes requested):

test/engine.spec.ts
izulin: I don't think it is possible to create array without formula.

wojciechczerniak: Ok, I forgot that `=` is needed

https://github.com/handsontable/hyperformula/blob/413e199a632452bffb45f33c747b53a6f8af2cab/test/arrays.spec.ts#L234-L237

wojciechczerniak (approved):
👍

Issue #812, opened 2021-09-17
Remove gpu.js from the project


krzysztofspilka:
It turned out that the gpu.js slows down the `npm i` of Handsontable. The problem lies in the dependency of HyperFormula that creates the WebGL context in Node.js (see [gl](https://www.npmjs.com/package/gl)). As there is no easy workaround for this problem, I recommend removing gpu.js from the source code, documentation, marketing materials, README; and all other mentions of GPU usage inside of HyperFormula.

GPU.js was a successful experiment to speed up the calculation time of some functions (MMULT, MAXPOOL, MEDIANPOOL, and TRANSPOSE) used within big data sets. It will be left in [version 1.0.0](https://github.com/handsontable/hyperformula/releases/tag/1.1.0) for all users who still need the GPU acceleration in their projects.

Handsontable 9.0.2 / HyperFormula 1.1.0

krzysztofspilka:
Update: the removal will be done in the next `major` version

AMBudnik:
Changed added to [HyperFormula v2.0.0](https://github.com/handsontable/hyperformula/issues/964) released today. I'm closing this issue as solved.

PR #813, opened 2021-09-17
Remove gpu.js & set node version to 16 & set coverageProvider to babel

voodoo11:
### Context

1. This PR removes completely `gpu.js` usage and dependency.
2. `node@16` is the version that works the most reliably with HF. I set it in Building Guide and CI scripts.
3. I set code coverage provider to `babel` (which is recommended by Jest) because it calculates the coverage more accurately. The side effect of that is that our coverage increased from ~92% to ~96% with this PR. It is the effect of changing the `coverageProvider`.

Documentation was adjusted by #814 

### How has this been tested?

Build & bundle works, unit tests pass.

### Types of changes

### Related issue(s):
1. #812 

### Checklist:


src/Config.ts
```diff
@@ -150,36 +147,6 @@ export interface ConfigParams {
...
-   *
-   * For more information, see the [GPU.js documentation](https://github.com/gpujs/gpu.js/#readme).
-   *
-   * @deprecated since version 1.2
-   *
-   * @default 'gpu'
-   *
-   * @category Engine
-   */
-  gpuMode: GPUMode,
```
sequba, line 182: These config options are marked as deprecated since version `1.2`. Should we keep them here for backward compatibility (they'd have no effect) or can they be removed?

warpech:
Thank you @sequba, it will be reviewed shortly.

sequba:
> ```diff
> @@             Coverage Diff             @@
> ##           develop     #813      +/-   ##
> ===========================================
> + Coverage    92.14%   95.70%   +3.55%     
> ===========================================
>   Files          167      161       -6     
>   Lines        40527    14018   -26509     
>   Branches      3734     2901     -833     
> ===========================================
> - Hits         37344    13416   -23928     
> + Misses        3177      602    -2575     
> + Partials         6        0       -6     
> ```

Codecove reports a huge difference because I changed the `coverageProvider` and it is calculated a bit differently (more accurately according to my analysis).

src/Config.ts
```diff
@@ -23,10 +23,7 @@ import {FunctionPluginDefinition} from './interpreter'
 import {Maybe} from './Maybe'
 import {ParserConfig} from './parser/ParserConfig'
 
-type GPUMode = 'gpu' | 'cpu' | 'dev'
-
-const PossibleGPUModeString: GPUMode[] = ['gpu', 'cpu', 'dev']
-const privatePool: WeakMap<Config, { licenseKeyValidityState: LicenseKeyValidityState }> = new WeakMap()
+const privatePool: WeakMap<Config, {licenseKeyValidityState: LicenseKeyValidityState}> = new WeakMap()
```
wszymanski, line 26:
```suggestion
const privatePool: WeakMap<Config, {licenseKeyValidityState: LicenseKeyValidityState}> = new WeakMap()
```

wszymanski (approved):
LGFM 🎉  I'm leaving just minor code style suggestions.

docs/guide/building.md
```diff
@@ -38,6 +38,8 @@ format , builds ES6 version
 * `npm run bundle:typings` - generates TS typing, only emits
 ‘.d.ts’ declaration files
 
+Note: We recommend using `node@16.14.0` for building HF. Other versions are not supported.
```
warpech, line 41: Nowhere in the docs, except for some code comments, we refer to the product at "HF". Let's use the full name.

Also, even after this change, our current CI setup uses the latest version of Node 16, so specifying such an exact version seems odd.

I propose to change this text to:

We use the latest version of Node 16 LTS in the build-chain and recommend this version for building. Note that for using (not building) HyperFormula, a wider range of Node versions is supported.

src/Config.ts
sequba: @wszymanski 
image

wszymanski: Changed.

docs/guide/building.md
sequba: done

Issue #9, opened 2019-11-14
Async functions support


wojciechczerniak:
### Description

Common request for our previous formula engine. Let's keep it as a feature request.

### Sample
```ts
parser.setFunction('SUM_ASYNC', (params) => {
  return new Promise((resolve) => {
    return setTimeout(() => resolve(params[0] + params[1]), 1000));
  }
});

parser.parse('SUM_ASYNC(2, 3)').then((result, error) => { });
```

MartinDawson:
This would be an extremely useful feature, especially for custom functions that need to load data from the backend like our `FINANCIAL` API.

Currently we preload data on page load but later on will need async functions I think as we can't preload all financial data for everything.

Is this on the roadmap soon? If not, is their an idea of how this will be done? Because I could try and do it in a few months when we will need it.

MartinDawson:
@wojciechczerniak I'm going to try and implement this feature because we need it for our site.

Could you give some rough guidance on the steps that need to be taken to do this please?

Unless this feature is a massive re-write or something.

Thanks

wojciechczerniak:
[Key Concepts](https://handsontable.github.io/hyperformula/guide/key-concepts.html) will be a good start to understand how HyperFormula works with AST. I would start with Evaluator or Interpreter, who calls function and when. Then go from there through each caller to make it async as well. This would stop at public methods, I think. Maybe @bardek8 or @izulin have more detailed ideas. Sorry, I couldn't be more helpful here, not my expertise.

Some of my thoughts and design considerations:

### Async and non async

Both function types should work. Preferably without changing everything to automatically resolved promises. We need to discover when function is async or not.

### Parallel evaluation

The naive implementation may traverse AST as synchronous code does. One function at a time. But while for synchronous code we cannot do anything about this, the asynchronous code is a different story. If asynchronous functions are used to make HTTP requests, we may parallelize them with [Promise.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) and continue resolving formulas after all input is ready.

### Repetitive ASTs

As mentioned in [Key Concepts](https://handsontable.github.io/hyperformula/guide/key-concepts.html) guide, same functions and ranges are grouped to avoid calculating the same value twice. But does it apply to asynchronous functions? I bet they will not always return same value. ie `=REMOTE_GUID()` to generate some unique ID, or `=RANDOM_PRODUCT()` that may return random output for each use case.

### Error handling

[...]

PR #863, opened 2021-11-27
[Draft] Added async functionality

MartinDawson:
Checklist for completed PR:

- [x] Tested in a real application
- [ ] Async and non async. We need to convert `runFunctionWithReferenceArgument` to async as well and add tests. Also the argumentTypes needs to allow async as users could want default async parameters. I.e like this:

`{argumentType: ArgumentTypes.ASYNC, defaultValue: '=PRICE("AMZN")'},`

- [x] Parallel evaluation
- [x] Repetitive ASTs - `volatile: true` should handle this automatically. Not async specific.
- [x] Error handling - Added TIMEOUT error & handling normal errors.
- [x] Circular dependencies - Shouldn't matter. It's already handled by existing code unless it's part of parallel evaluation which is mentioned blelow.
- [x] Named Ranges
- [x] Volatile - Users should implement their own cache on backend for requests. If a function is volatile then use `volatile: true`
- [x] Undo-redo - Should re-do the function again like normal. Google Sheets does this. (needs tests)
- [x] API has to be async. Used a Tuple for this.
- [x] Helper functions
- [x] Batch

### Context
Unfinished PR. Draft here for continuous review from handsontable devs

Key points:

- The public functions should not be async by default as we want to return the changes still for the first run through. So I've returned a Tuple with [changes, promise] and internally. This allows users to await the promise if they need to however it's not mandatory as they can also listen for the `AsyncValuesUpdated` (async equivalent to `ValuesUpdated` event) event to be fired and update their cells.

### How has this been tested?
async-functions.spec.ts

### Types of changes
- [x] Additional language file or change to the existing one (translations)

### Related issue(s):
1. https://github.com/handsontable/hyperformula/issues/9

### Checklist:


MartinDawson:
Possible enhancements:

- [ ] Pre-fetch. We could allow users of Hyperformula to fire the async formulas & it's dependencies but not do the calculations on resolve of the promise. This is similar to pre-fetching a links that webpages do when the user hovers over them. For example, the user types `=PRICE("GM")` into a cell but doesn't click enter we could immediately start fetching the async data.

MartinDawson:
Converted `Loading...` to be a cell error instead of normal text. Same as Google Sheets this allows all dependent cells to automatically have the `Loading...` error to which is correct.

MartinDawson:
Parallel evaluation is done.

Just thinking it might be better to split out asyncMethod into a seperate one like the arraySize method. This way we don't need to convert runFunction stuff I think.

Can do that at a later date though.

MartinDawson:
Not correct implementation in the end after testing, so closing.

zukilover:
Hi @MartinDawson , I've noticed that your 2 PRs on async function have been closed. I'm wondering if there is still any ongoing development on this specific feature?

warpech:
@zukilover I replied in https://github.com/handsontable/hyperformula/issues/9#issuecomment-1029829216

PR #868, opened 2021-12-07
Docs: Expanding `precisionRounding` info

kirszenbaum:
This PR:
- Expands the [`precisionRounding` API ref](https://handsontable.github.io/hyperformula/api/classes/config.html#precisionrounding)
- Removes the `precisionRounding` option from examples where it's not necessary

### DIFF
--- a/docs/guide/configuration-options.md
+++ b/docs/guide/configuration-options.md
@@ -17,7 +17,6 @@ static method called to initiate a new instance of HyperFormula.
 // define options 
 const options = {
     licenseKey: 'gpl-v3',
-    precisionRounding: 10,
     nullDate: { year: 1900, month: 1, day: 1 },
     functionArgSeparator: '.'
 };
--- a/docs/guide/configuration-options.md
+++ b/docs/guide/configuration-options.md
@@ -17,6 +17,7 @@ static method called to initiate a new instance of HyperFormula.
 // define options 
 const options = {
     licenseKey: 'gpl-v3',
+    precisionRounding: 10,
     nullDate: { year: 1900, month: 1, day: 1 },
     functionArgSeparator: '.'
 };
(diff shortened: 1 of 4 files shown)


src/Config.ts
```diff
@@ -321,9 +321,13 @@ export interface ConfigParams {
...
   /**
-   * Sets calculations' precision level.
+   * Sets the precision level of calculations' output.
    *
-   * Numerical outputs are rounded to the `precisionRounding` number of digits after the decimal.
+   * Internally, all arithmetic operations are performed using JavaScript's built-in numbers.
+   * But when HyperFormula exports a cell's value, it rounds the output
+   * to the `precisionRounding` number of significant digits.
+   *
+   * We recommend setting `precisionRounding` to a value between `1` and `14`.
```
bardek8, line 330:
```suggestion
   * For a typical use we recommend setting `precisionRounding` to a value between `10` and `14`.
   * Too small value might result in losing precision of large numbers, e.g. 100005 becomes 100010 when `precisionRounding=4`.
```

docs/guide/configuration-options.md
```diff
@@ -17,7 +17,6 @@ static method called to initiate a new instance of HyperFormula.
 // define options 
 const options = {
     licenseKey: 'gpl-v3',
-    precisionRounding: 10,
```
bardek8, line 20: I wouldn't remove it here - it's a reasonable example of config option, but I have no strong opinion on that.

src/Config.ts
bardek8: Please decide if we want to include the second line with the example.

bardek8 (approved):

src/Config.ts
kirszenbaum: @bardek8 - sure, added: 8ee4f1c

docs/guide/configuration-options.md
kirszenbaum: I agree, putting it back: 8ee4f1c

Issue #765, opened 2021-07-13
Call `runFunction` internally so it's easier to create custom functions


wojciechczerniak:
### Description

Currently each function calls `runFunction` with a callback. It does a lot of useful things and reduces the boilerplate (ie. argument validation) so it's natural that most implementations will want to use it. Also having `meta` properties for each functions kind of feels that validations is always done by the engine.

https://github.com/handsontable/hyperformula/blob/43071aa2a822ddcea04a7b1016fa3e8dda78e751/src/interpreter/plugin/ModuloPlugin.ts#L22-L30

Would become a simple 
```ts
public mod(ast: ProcedureAst, formulaAddress: SimpleCellAddress, args: ModArgs): InternalScalarValue { 
  const { dividend, divisor } = args
 
  if (divisor === 0) { 
     return new CellError(ErrorType.DIV_BY_ZERO) 
  }

  return dividend % divisor 
} 
```

Interpreter already have access to AST, function meta and other properties, no need to create all this boilerplate code.

### Other thoughts
- We may have a flag to make the internal validation optional if someone want so implement everything on their own.
- We should keep `ast` and `formulaAddress`, may be useful

thilgen:
Note related conversation (https://github.com/handsontable/hyperformula/pull/1227) about the introduction and usage of `runFunctionWithReferenceArgument`

PR #875, opened 2021-12-14
Added support for infering return types

MartinDawson:
### Context
Consider any formula that you do not know the exact return type of or the return type is too large and can change based on parameters :

`=GET_FINANCIALS("AMZN", "revenue")`, this would return a currency value.

`=GET_FINANCIALS("AMZN", "operatingMargin")`, this would return a percentage value.

`=GET_FINANCIALS("AMZN", "balanceSheet")`, this would return an object of currencies & percentages.

Hyperformula currently has this for custom plugins: `returnNumberType`. This isn't sufficient for functions who's return types can change dynamically.

The percentages should be automatically formatted as percentages. The currencies as currencies.

Currently this is not possible in HF and this causes bugs where other formulas don't work if you put the above as parameters because the cell types are wrong.

However we should not apply it to all functions automatically because some functions such as `=TEXT()` want to format stuff as strings specifically.

My PR currently only infers if `runFunction*` is called. When https://github.com/handsontable/hyperformula/issues/765 is done it should be done automatically regardless if the user has `inferReturnTypes` enabled.

### How has this been tested?
See the .spec file. It isn't fully tested yet for all scenario's.

I need approval that this feature is needed first before I spend time on tests.

### Types of changes

### Related issue(s):
1. N/A

### Checklist:

### DIFF
--- a/src/CellContentParser.ts
+++ b/src/CellContentParser.ts
@@ -3,187 +3,216 @@
  * Copyright (c) 2021 Handsoncode. All rights reserved.
  */
 
-import {CellError, ErrorType} from './Cell'
-import {Config} from './Config'
-import {DateTimeHelper, timeToNumber} from './DateTimeHelper'
-import {ErrorMessage} from './error-message'
-import {UnableToParseError} from './errors'
-import {fixNegativeZero, isNumberOverflow} from './interpreter/ArithmeticHelper'
-import {
-  cloneNumber,
-  CurrencyNumber,
-  DateNumber,
-  DateTimeNumber,
-  ExtendedNumber,
-  getRawValue,
-  PercentNumber,
-  TimeNumber
-} from './interpreter/InterpreterValue'
-import {Maybe} from './Maybe'
-import {NumberLiteralHelper} from './NumberLiteralHelper'
-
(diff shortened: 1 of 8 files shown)


src/CellContentParser.ts
```diff
@@ -3,187 +3,216 @@
...
+         return new CellContent.Number(new DateTimeNumber(val, 'Date()'))
+       }
+     } else if (typeof content === 'string') {
+       if (isBoolean(content)) {
+         return new CellContent.Boolean(content.toLowerCase() === 'true')
+       } else if (isFormula(content)) {
+         return new CellContent.Formula(content)
+       } else if (isError(content, this.config.errorMapping)) {
+         return new CellContent.Error(this.config.errorMapping[content.toUpperCase()])
+       } else if (isPercent(content)) {
```
MartinDawson, line 145: I don't know why git is saying this entire file changed. I simply separated out the percent/currency checks into if statements as well instead of a variable condition.

sequba:
HyperFormula team decided that this feature is not aligned with our current roadmap. I am merging it to a feature branch in HyperFormula repository in case we wanted to use this code in the future.

PR #882, opened 2022-01-03
Fix duplicated messages about deprecated options usage

budnix:
### Context

The PR fixes an issue related to the doubled or tripled printed deprecated warning messages. 

After the changes once the Hyperformula is initialized the warn messages are logged only once. The messages are logged again only if the `engine.updateConfig()` method is called and only when the deprecated options are passed.

Do prints deprecated warn messages:
 - `HyperFormula.buildEmpty({ binarySearchThreshold: 10 });`
 - `engine.updateConfig({ binarySearchThreshold: 10 })`

Do not print deprecated warn messages: 
 - `engine.updateConfig({ precisionRounding: 10 })`
 - `engine.rebuildAndRecalculate()`

I had a problem with tests in the browser. It turned out that there is a problem with unregistering Jasmine spies that interfere with other tests. To fix it up I upgraded the Jasmine to the latest version.

The PR solves the problem with Handsontable integration where its _Formulas_ plugin calls the HF `rebuildAndRecalculate` method. The method internally merges the Config object with default values (e.g. `binarySearchThreshold: 20`) that once passed in the constructor logs again warning messages about the usage of deprecated options. See https://github.com/handsontable/handsontable/issues/8771.

### How has this been tested?

I tested the changes locally and I added some tests that cover the fix. 

### Types of changes

### Related issue(s):
1. #830

### Checklist:


src/Config.ts
```diff
@@ -659,9 +663,7 @@ export class Config implements ConfigParams, ParserConfig {
     validateNumberToBeAtLeast(this.precisionEpsilon, 'precisionEpsilon', 0)
     this.useColumnIndex = configValueFromParam(useColumnIndex, 'boolean', 'useColumnIndex')
     this.useStats = configValueFromParam(useStats, 'boolean', 'useStats')
-    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-    // @ts-ignore
-    this.binarySearchThreshold = undefined
+    this.binarySearchThreshold = binarySearchThreshold ?? Config.defaultConfig.binarySearchThreshold
```
warpech, line 666: Unless I am terribly mistaken, this PR missed the most important point, that `binarySearchThreshold` is not used anywhere. 

Therefore, it should be removed as a member of this class and from the default config object.

We can keep the warning if someone tries using it.

warpech (changes requested):

src/Config.ts
budnix: As a continuation of the original PR (https://github.com/handsontable/hyperformula/pull/830) I tried to focus on fixing the double warnings and covering the change with tests only. To keep the PR as simple as possible with the possibility to release the patch version.

Removing the `binarySearchThreshold` option from defaults would generate a new TS definition. I don't know if that wouldn't be a breaking change?

warpech: Yes, you are right that it would be an unnecessary breaking change. With your change, it is possible to release this as a PATCH version.

warpech (approved):

AMBudnik:
[Released in HyperFormula 1.3.1](https://github.com/handsontable/hyperformula/releases/tag/1.3.1)

Issue #905, opened 2022-02-15
RATE function misbehaves


mrfullset-yolk:
### Description

RATE function misbehaves for some valid inputs.

Examples:

- input: (300, -465.96, 100000), expected: ~0.25%, got: #NUM!
- input: (200, -500, 200000), expected: ~-0.62%, got: #NUM!

### Steps to reproduce
```
const hf = HyperFormula.buildFromArray([['=RATE(200, -500, 200000)']], { licenseKey: 'gpl-v3' });
console.log(hf.getCellValue({ col: 0, row: 0, sheet: 0 }));
```

### Your environment
MB Pro 2016
macOS 12.2
node v14.17.4
npm v6.14.14

* HyperFormula version: 13.3.1
* Browser Name and version: Chrome 98.0.4758.80 x86_64
* Operating System: macOS 12.2

### Links
https://gist.github.com/kucukharf/677d8f21660efaa33f72 -- this algorithm works ok. maybe it's ok as a reference for you.

sequba:
Hi @mrfullset-yolk!

I investigated your issue. In fact our algorithm is able to compute the correct values for these inputs but it takes too many iterations. We'll see what we can do to make it converge faster.

> https://gist.github.com/kucukharf/677d8f21660efaa33f72 -- this algorithm works ok. maybe it's ok as a reference for you.

How did you verified that kucukharf's implementation works correctly? Could you share your code? It might be helpful for us.

mrfullset-yolk:
Hi @sequba,

I've verified that algorithm with provided examples in the issue. kucukharf's implementation returns correct values, which are the same with MS Excel.

The only code I had is also provided in the description. It was a usual React functional component with only those 2 rows inside.

sequba:
Thanks. In the meantime I managed to fix out implementation and it will be included in the upcoming release of HF: #933

AMBudnik:
Hi @mrfullset-yolk I'm just here to share good the good news. HyperFormula v2 is already online with @sequba 's fix for this issue.

Here https://github.com/handsontable/hyperformula/issues/964 you can read more about the changes within that version.

PR #933, opened 2022-03-10
Improve convergence of the RATE function

sequba:
### Context
RATE function was slow to converge for some input data (#905)

### How has this been tested?
- More unit tests added for the RATE function with some less trivial inputs
- Build works, unit test pass.

### Types of changes

### Related issue(s):
1. #905 

### Checklist:


docs/guide/known-limitations.md
```diff
@@ -72,11 +72,7 @@ To remove the differences, you can create custom implementations of those functi
 | DAYS          | =DAYS(-1, 0)                                               |          NUM |      -1,0000 |        NUM |
 | DAYS          | =DAYS(0, -1)                                               |          NUM |       1,0000 |        NUM |
 | DATEDIF       | =DATEDIF(-1, 0, "Y")                                       |          NUM |       0,0000 |        NUM |
-| RATE          | =RATE(12, -100, 400, 100, 1)                               |      -0,4997 |       0,3100 |    -0,4997 |
-| RATE          | =RATE(12, -100, 400, 1, 1)                                 |      -0,9901 |       0,3200 |    -0,9901 |
-| RATE          | =RATE(12, -100, 400, 0, 1)                                 |          NUM |       0,3200 |        NUM |
-| RATE          | =RATE(12, -100, 400, -100, 1)                              |          NUM |       0,3200 |        NUM |
-| RATE          | =RATE(0.9, -100, 400)                                      |      -0,7962 |          NUM |    -0,7962 |
+| RATE          | =RATE(12, -100, 400, 0, 1)                                 |      -1,0000 |          NUM |        NUM |
```
warpech, line 75: It is weird that on this docs page we use the comma as the decimal separator. The ODFF standard uses the dot. Maybe make a new issue about that?

docs/guide/known-limitations.md
```diff
@@ -72,11 +72,7 @@ To remove the differences, you can create custom implementations of those functi
 | DAYS          | =DAYS(-1, 0)                                               |          NUM |      -1,0000 |        NUM |
```
warpech, line 72: The pages https://handsontable.github.io/hyperformula/guide/specifications-and-limits.html and https://handsontable.github.io/hyperformula/guide/known-limitations.html appear disturbingly similar in topic. I know that one of them is about *limits* and the other about *limitations*, but I guess that the reader is interested in both. Maybe the pages should be cross linked at least?

Anyway, should something about `RATE` be added to the other page (https://handsontable.github.io/hyperformula/guide/specifications-and-limits.html)?

test/interpreter/function-rate.spec.ts
```diff
@@ -13,29 +15,211 @@ describe('Function RATE', () => {
...
+  })
+
+  it('should set the value type to percent', () => {
     const engine = HyperFormula.buildFromArray([
       ['=RATE(12, -100, 400)', ],
-      ['=RATE(12, -100, 400, 100, 1)', ],
-      ['=RATE(12, -100, 400, 1, 1)', ],
-      ['=RATE(12, -100, 400, 0, 1)', ],
-      ['=RATE(12, -100, 400, -100, 1)', ],
-      ['=RATE(12, -100, 400, 100, 1, -1)', ],
```
warpech, line 23: The assertion for the formula presented above is missing from the new tests. I know that you added another assertion (`=RATE(12, -100, 400, 0, 0, -1)`), so disregard this comment if you think it does not have a point.

warpech (commented):
No real issues, just a few comments. Thanks for much more detailed tests!

src/interpreter/plugin/FinancialPlugin.ts
```diff
@@ -464,15 +464,16 @@ export class FinancialPlugin extends FunctionPlugin implements FunctionPluginTyp
...
       (periods, payment, present, future, type, guess) => {
         if (guess <= -1) {
           return new CellError(ErrorType.VALUE)
         }
 
-        const epsMax = 1e-10
+        const epsMax = 1e-7
 
-        const iterMax = 20
+        const iterMax = 50
```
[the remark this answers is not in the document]
sequba: > the spec mentions that the maximum number iterative calculations of cyclic references should be configurable by the host app (HOST-ITERATION-STEPS), but that is a different number than used here, right?

I believe it's a parameter for the algorithm that finds cycles in the dependency graph. It has nothing to do with the `RATE` function.

ODFF specifies [here](https://docs.oasis-open.org/office/OpenDocument/v1.3/os/part4-formula/OpenDocument-v1.3-os-part4-formula.html#RATE) what this function should calculate but not what algorithm should be used let alone the parameters of this algorithm.

In my opinion the value of this parameter is implementation-specific. [This software](https://support.microsoft.com/en-us/office/rate-function-9f665657-4a7e-4bb7-a030-83fc59e748ce) claims that it limits the number of iteration to `20`, but we cannot really compare it, because their implementation is certainly different. One could write this function is such a way that it would perform 2 steps per 1 iteration of a loop and than it would converge in less iterations but it would still take the same time to complete.

> If we hardcode a value that is unreasonably big, it might become a problem. However, this function does not look particularly expensive, so it shouldn't be a problem, would you agree?

I checked and even with `iterMax = 5000` **our implementation** of this algorithm completes in ~10ms on my PC.

docs/guide/known-limitations.md
sequba: Most (all?) english speaking cultures use `.` so I believe we also should. Issue: https://github.com/handsontable/hyperformula/issues/942

docs/guide/known-limitations.md
sequba: > Maybe the pages should be cross linked at least?

Issue: Maybe the pages should be cross linked at least?

> Anyway, should something about RATE be added to the other page

Well, IMO the values of `epsMax` and `iterMax` parameters should appear in our documentation. But I would rather put them in the section about `RATE` function (which doesn't exist). @kirszenbaum what do you think?

test/interpreter/function-rate.spec.ts
sequba: Line 28 of this file

sequba: Yeah, they are equivalent. They check if an error will be thrown for the `guess` value of `-1`.

docs/guide/known-limitations.md
kirszenbaum: ### `RATE`

For now, we don't have any sections devoted to individual functions. The only places where we describe any details of individual functions are:
- [Bulit-in functions](https://handsontable.github.io/hyperformula/guide/built-in-functions.html)
- [Known limitations](https://handsontable.github.io/hyperformula/guide/known-limitations.html#nuances-of-the-implemented-functions)

IMO, if `RATE` works differently than GS or Excel (`iterMax`), we should add an entry about it to this table:
- [Google Sheets and Microsoft Excel](https://handsontable.github.io/hyperformula/guide/known-limitations.html#google-sheets-and-microsoft-excel)

Any other information about `RATE` – I'd add it here:
- [Nuances of the implemented functions](https://handsontable.github.io/hyperformula/guide/known-limitations.html#nuances-of-the-implemented-functions)

### Known limitations

I'd definitely change the title of the [Known limitations](https://handsontable.github.io/hyperformula/guide/known-limitations.html) page:
- It looks bad (especially in the **Getting started** section)
- It doesn't match the contents of the page (implementation differences and "function nuances" are not limitations)
- It indeed sounds similar to "Specifications and limits"

My proposition is to:
- Rename "Known limitations" to "Implementation details" or similar
- Move this page from the **Getting started** section to the **Advanced topics** section
- Change the introduction sentence to actually cover what's on the page (also, there's a typo in the 1st sentence at the moment: "This pages lists")
- Turn the top bullet list ("Node.js versions older than...") into a separate sub-section of the page (and **that** section could be called "Known limitations", or "Areas for improvement", or "Areas for further development")

sequba: I described the implementation details in 'Nuances of the implemented functions':

>* RATE function might have no solutions (depending on the inputs). Sometimes it has multiple solutions. Our implementation uses an iterative algorithm (Newton's Method) for finding an approximation for one of them to within `1e-7`. If it is not found after `50` iterations function returns the `#NUM!` error.

But - really - from the perspective of our user, it would make much more sense to keep all the function-related information in the description of a function instead of spreading them across [List of available functions](https://handsontable.github.io/hyperformula/guide/built-in-functions.html#list-of-available-functions). [Nuances of the implemented functions](https://handsontable.github.io/hyperformula/guide/known-limitations.html#nuances-of-the-implemented-functions) and [Google Sheets and Microsoft Excel](https://handsontable.github.io/hyperformula/guide/known-limitations.html#google-sheets-and-microsoft-excel) sections.

warpech: Thank you for bright ideas about these pages. It is definitely off-topic though so let's track it in the new issue.

src/interpreter/plugin/FinancialPlugin.ts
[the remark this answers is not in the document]
warpech: Thanks!

warpech (approved):
Thank you @sequba for a very scrupulous work on this.

Issue #58, opened 2019-12-04
Global settings


wojciechczerniak:
### Description

There are some differences between spreadsheet applications, and some of them can be compensated with customizable options.

|Done| Option Name  | Default | Description |
|---|-----------|--------|----|
| ✅  | undoLimit    | `20`   |   #295 |
| ✅  | maxRows    | `40,000`   |  #294 for #198 |
| ✅  | maxColumns    | `18,278`   |  #294 for #198 |
| ✅  | licenseKey    | undefined   |  #222 |
| ✅  | nullDate    | `1900`   |  Date is stored as a number of days since this date. There are a few "date systems" in the world of spreadsheets: 1900 Date System, 1904 Date System [1] #3 |
| ✅ | nullYear    |   `30` | If date contains two-digit year it will be parsed as `19xx` for numbers above this setting and `20xx` for numbers below  i.e `2029` but `1930`. ODFF section 6.10.23 |
| ✅ | leapYear1900 | `false` | Callback that will allow handling leap year compatibility problem (introduced in Lotus 1-2-3) in various spreadsheets [2] #3 |
| ✅  | decimalSeparator  |  `.` | Number format #42 |
| ✅  | thousandSeparator  |  `` | Number format #42 |
| ✅ | language | `enGB` | Formula translations |
| ✅ |  accentSensitive | `false` |  String comparison #130 |
| ✅ |  caseFirst | `"lower"` | String comparison #130  |
| ✅ |  ignorePunctuation | `false` | String comparison #130  |
| ✅ |   localeLang | `"en"` | String comparison #130 |
| ✅ | caseSensitive | `false` | Comparison operators `=`, `<>`, `<`, `<=`, `>`, and `>=` can be either case sensitive or insensitive. Exposed as an option in many spreadsheet apps | 
| | intersectionOperator | ` ` (space) | ODFF section 5.5 and 6.4.12, default is `!` but some implementations (XL, GS) use a space ` ` character. Example `=SUM([.A1:.C4]![.B1:.B5])` vs `=SUM(A1:C4 B1:B5)` [3] #15 |
| | unionOperator | `,` | ODFF section 6.4.13, default is `~` but some implementations use comma `,` or `+` character. Example `=SUM([.A1:.C4];[.B1:.B5])` vs `=SUM(A1:C4, B1:B5)` [3] #15 |
| | rangeOperator | `:` | Optional. There are no other, known implementations. |
[...]

PR #935, opened 2022-03-10
Changed whitespace regex to `/s+` (intermediate branch to fix the conflicts)

sequba:
This is a technical PR that was necessary to resolve the conflicts from the forked repository

### DIFF
--- a/src/parser/LexerConfig.ts
+++ b/src/parser/LexerConfig.ts
@@ -95,7 +95,7 @@ export const ErrorLiteral = createToken({name: 'ErrorLiteral', pattern: /#[A-Za-
 /* skipping whitespaces */
 export const WhiteSpace = createToken({
   name: 'WhiteSpace',
-  pattern: /[ \t\n\r]+/,
+  pattern: /\s+/,
 })
 
 export interface ILexerConfig {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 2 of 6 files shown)


src/parser/LexerConfig.ts
```diff
@@ -95,7 +95,7 @@ export const ErrorLiteral = createToken({name: 'ErrorLiteral', pattern: /#[A-Za-
 /* skipping whitespaces */
 export const WhiteSpace = createToken({
   name: 'WhiteSpace',
-  pattern: /[ \t\n\r]+/,
+  pattern: /\s+/,
```
warpech, line 98: @sequba wrote in the OP:

> The standard does not mention whitespace characters other than SPACE, TAB, LINE FEED, and CARRIAGE RETURN so I assume it is up to us how we want to handle them.

Our implementation will become non-standard if we do whatever we want. I think we should make the behavior configurable, as we did in the past with other ideas that go beyond the ODFF standard: https://github.com/handsontable/hyperformula/issues/58

warpech (changes requested):

src/parser/LexerConfig.ts
sequba: IMO it is highly unlikely that someone wants to display `#ERROR!` for formulas including non-breakable spaces but it won't hurt to add a config option for controlling that behaviour. I'll work it.

sequba: I added `allowAllWhitespace` parameter

warpech (commented):
Technically it looks very good but I have a doubt about the option itself.

warpech (approved):
Thanks, love it!

Issue #884, opened 2022-01-11
The github actions workflow for publishing the docs crashed for 1.3.1.


jansiegel:
### Description
The GHA workflow responsible for building and publishing the documentation crashed on `npm ci` after publishing `1.3.1`: https://github.com/handsontable/hyperformula/runs/4773934020?check_suite_focus=true

It looks to me like it might be caused by the `gl` package being built on the fly and having some internal issues. 
The `ci` calls were successful for other workflows, which leads me to a conclusion, that the problem might be caused by using a different Node version for the publishing workflow (others use Node@14 and Node@15, the publishing one doesn't have a fixed Node version).

I think it'd be a good idea to check if setting the same (tested) version of Node for all workflows would fix the issue.

### Steps to reproduce
1. See the log at https://github.com/handsontable/hyperformula/runs/4773934020?check_suite_focus=true

warpech:
Adding this to the scope of the next release. If we fail to fix it easily, then we can postpone it.

sequba:
I could not reproduce this issue on develop. It might have been fixed earlier by something else or it happens only on master. We must make sure that the workflow completes after the release.

sequba:
Fixed by #946

PR #946, opened 2022-03-23
Improve publishing GH pages workflow

sequba:
### Context
 Issue #884 could not be reproduced on `develop`. It might have been fixed earlier by something else or it happens only on `master`. So we must make sure that the workflow completes after the release.

This PR improves gh-pages workflow by adding the `matrix` to control the node version better.

### How has this been tested?
I enabled running this workflow on a feature branch and did so. It completed with no errors: https://github.com/handsontable/hyperformula/actions/runs/2029078092

### Types of changes

### Related issue(s):
1. #884 
2.
3.

### Checklist:

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -7,16 +7,28 @@ on:
 
 jobs:
   docs:
+    strategy:
+      matrix:
+        node-version: [ '16' ]
+        os: [ 'ubuntu-latest' ]
     name: DOCS
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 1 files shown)


jansiegel (approved):
As the issue was most likely caused by the `gl` package (described in #884), the problem seems to have been resolved by removing the `gpu.js` dependency (bec4ec0d159c399ee57839bca9aa123eea60f028), which makes it work fine on `develop`.

I tested the workflow on a setup with `gpu.js` still listed as a dependency and it fails the same way as it did before.

That being said, the issue doesn't seem to be a problem anymore and unifying the node version between the workflows is a good idea anyway 👍

Issue #952, opened 2022-03-31
Explain 3D range references in the docs


warpech:
### Description

HyperFormula supports 3D range references (ranges spanning across multiple sheets). This is useful in many ways, as presented in multiple advanced Excel tutorials found online, e.g.:

- https://www.ablebits.com/office-addins-blog/2015/12/09/excel-3d-reference-formula/
- https://support.microsoft.com/en-us/office/create-a-3-d-reference-to-the-same-cell-range-on-multiple-worksheets-40ca91ff-9dcb-4ad1-99d2-787d0bc888b6

We should present a viable use case in the Guide. 

I am not sure if the name "3D range reference" is a correct one.

### Type of issue

- [x] Missing documentation
- [ ] Error in the documentation
- [ ] Request for a demo
- [ ] Request for a tutorial, guide etc.
- [ ] Other issues, suggestions or ideas

krzysztofspilka:
It's described here https://hyperformula.handsontable.com/guide/cell-references.html#referencing-ranges and since it's a niche feature from the user's perspective, I'm closing this issue.

PR #955, opened 2022-04-04
Docs: Adding information about range references

kirszenbaum:
This PR:
- Adds a new **Range references** sectio to the **Cell references** page
-

### DIFF
--- a/docs/guide/cell-references.md
+++ b/docs/guide/cell-references.md
@@ -49,8 +49,8 @@ different cells in the workbook.
     </tr>
     <tr>
       <td style="text-align:left">Range</td>
-      <td style="text-align:left">=A1:C10</td>
-      <td style="text-align:left">=Sheet2!A1:C10</td>
+      <td style="text-align:left">=A1:D10</td>
+      <td style="text-align:left">=Sheet2!A1:D10</td>
     </tr>
   </tbody>
 </table>
@@ -133,7 +133,7 @@ const changes = hfInstance.moveCells(source, 1, 1, destination);
 console.log(changes);
 ```
 
-## Absolute reference
+## Absolute references
 
 A reference to a column (a letter) or a row (a number) may be
 preceded with a dollar sign `$` to remain intact when the cell is
@@ -160,7 +160,37 @@ copied between different places.
   </tbody>
 </table>
 
-## Circular reference
+## Range references
+
+In HyperFormula, a range is a group of two or more adjacent cells.
+
+You can reference ranges in the following ways:
+
+| Range reference type | Example | Reversed example |
+| -------------------- | ------- | ---------------- |
+| Two cell addresses   | A1:D10  | D10:A1           |
+| Two columns          | A:D     | D:A              |
+| Two rows             | 3:5     | 5:3              |
+
+You can reference ranges:
+- With relative references (A1:D10)
+- With absolute references (A$1:$D$10)
+- In different sheets (=Sheet2!A1:D10)
+
+You can't mix two different types of range references together (A1:D).
+
+Range expressions can't contain [named expressions](/guide/named-expressions.md).
+
+::: tip
+In contrast to Google Sheets or Microsoft Excel, HyperFormula doesn't treat single cells as ranges. Instead, it immediately instantiates references to single cells as their values. Applying a scalar value to a function that takes ranges throws the [`CellRangeExpected`](/api/classes/errormessage.md#cellrangeexpected) error.
+:::
+
+#### More about ranges
+- [Key concepts: Handling ranges](/guide/key-concepts.md#handling-ranges)
+- [Types of operators: Reference operators](/guide/types-of-operators.md#reference-operators)
+- [API reference: Ranges](/api/classes/hyperformula.md#ranges)
+
+## Circular references
 
 Since HyperFormula does not embed any UI, it allows for the input of a circular reference into a cell. Compared to popular spreadsheets,
 HyperFormula does not force any specific interaction with the user
@@ -182,9 +212,9 @@ used in a cell.
 
 Consider the following example:
 
-| Formula in C1 | Action | Result in B1 |
-| :--- | :--- | :--- |
-| =A1+B1+20 | Delete column A | #REF! |
+| Formula in C1 | Action          | Result in B1 |
+| :------------ | :-------------- | :----------- |
+| =A1+B1+20     | Delete column A | #REF!        |
 
 The #REF! error may also occur in other specific situations:
 


docs/guide/cell-references.md
```diff
@@ -160,7 +160,38 @@ copied between different places.
...
+
+In HyperFormula, a range is a group of two or more adjacent cells.
+
+You can reference ranges in the following ways:
+
+| Range reference type | Example | Reversed examples       |
+| -------------------- | ------- | ----------------------- |
+| Cell addresses       | A1:B2   | A2:B1<br>B1:A2<br>B2:A1 |
+| Columns              | A:B     | B:A                     |
+| Rows                 | 1:2     | 2:1                     |
```
sequba, line 173: I'd rather describe these as separate types of range than ways of refencing.

- Cell ranges are always in shape of a finite rectangle
- Column ranges contain the whole columns (they are infinite in the row dimension)
- Row ranges contain the whole rows

docs/guide/cell-references.md
```diff
@@ -160,7 +160,38 @@ copied between different places.
...
+
+You can reference ranges:
+- Through relative references (A1:B2)
+- Through absolute references (A$1:$B$2)
+- Across different sheets (=Sheet2!A1:B2)
```
sequba, line 178: I'd mention that we do support ranges that spans across multiple sheets (`Sheet1!A1:Sheet2!B2`). Ref: #952

sequba: Also:
> If you don't specify a sheet name for the second address, it's assumed that it is the same as in first address, so `Sheet2!A1:B2` is equivalent to `Sheet2!A1:Sheet2!B2`

docs/guide/cell-references.md
```diff
@@ -160,7 +160,38 @@ copied between different places.
   </tbody>
 </table>
 
-## Circular reference
+## Range references
+
+In HyperFormula, a range is a group of two or more adjacent cells.
```
sequba, line 165: >  Range `<Addr1>:<Addr2>` is a reference to the smallest 3-dimensional cube of cells that include both `Addr1` and `Addr2` including the cells on sheets positioned between them.

IMO this sentence describes it pretty well. Although it might be too technical for some users. @kirszenbaum I leave it up to you if you want to use this sentence or not.
 
 From: https://docs.oasis-open.org/office/OpenDocument/v1.3/os/part4-formula/OpenDocument-v1.3-os-part4-formula.html#__RefHeading__1018032_715980110

docs/guide/cell-references.md
```diff
@@ -160,7 +160,38 @@ copied between different places.
...
+- Through absolute references (A$1:$B$2)
+- Across different sheets (=Sheet2!A1:B2)
+
+The following restraints apply:
+- You can't mix two different types of range references together (A1:B).
+- Range expressions can't contain [named expressions](/guide/named-expressions.md).
+- At the moment, HyperFormula doesn't support multi-cell range references (A1:B2:C3).
+
+::: tip
+In contrast to Google Sheets or Microsoft Excel, HyperFormula doesn't treat single cells as ranges. Instead, it immediately instantiates references to single cells as their values. Applying a scalar value to a function that takes ranges throws the [`CellRangeExpected`](/api/classes/errormessage.md#cellrangeexpected) error.
```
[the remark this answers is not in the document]
kirszenbaum: @sequba - I think it's mentioned [here](https://github.com/handsontable/hyperformula/blob/master/docs/guide/known-limitations.md?plain=1#L32) and [here](https://github.com/handsontable/hyperformula/blob/master/docs/guide/known-limitations.md?plain=1#L50)

docs/guide/cell-references.md
kirszenbaum: @sequba I went for a simpler, "end-user" option because I wasn't so sure about the ODF standard's definition, particularly about the **3-dimensional cube** metaphor:
- It's a tautology (a cube that has three dimensions by default)
- A cube has 12 edges of identical length (that's not the case with ranges, right?). I think they meant **cuboid** here
- Looking at [this](https://docs.oasis-open.org/office/OpenDocument/v1.3/os/part4-formula/OpenDocument-v1.3-os-part4-formula.html#Reference), all we want to say by mentioning "3 dimensions" is that ranges support cross-sheet referencing, am I right?

To make that definition consistent with the [4.8 ref definitions](https://docs.oasis-open.org/office/OpenDocument/v1.3/os/part4-formula/OpenDocument-v1.3-os-part4-formula.html#Reference), we'd need to reword it like this:

> Range `<Cell position 1>:<Cell position 2>` is a reference to the smallest possible cell cuboid that includes `<Cell position 1>`, `<Cell position 2>`, and all cells on all sheets positioned between `<Cell position 1>` and `<Cell position 2>`.

But IMO the above only makes sense when you're actually familiar with the [4.8 Reference](https://docs.oasis-open.org/office/OpenDocument/v1.3/os/part4-formula/OpenDocument-v1.3-os-part4-formula.html#Reference) terminology: **cell position**, **cell strip**, **cell rectangle**, **cell cuboid**, and **reference**. And we're not using that terminology anywhere else in the HF docs. I believe the definition should be more-or-less self-explanatory for people not familiar with the ODF standard.

I'd also differentiate between a **range** of cells and a **range reference** (a reference to a range of cells).

TL;DR
How about something like this:

> Range `<Cell address 1>:<Cell address 2>` is the smallest possible group of adjacent cells that includes:
> - The cell at `<Cell address 1>`
> - The cell at `<Cell address 2>`
> - All cells on all sheets between `<Cell address 1>` and `<Cell address 2>`
>
> Tip: In HyperFormula, a single cell is not treated as a range.

WDYT?

docs/guide/cell-references.md
kirszenbaum: > I'd mention that we do support ranges that spans across multiple sheets 

I'm adding the `Sheet1!A1:Sheet2!B2` example, but I'd say it's **across two different sheets**, not **multiple sheets**, right? Because you can't do `Sheet1!A1:Sheet2!B2:Sheet3!C3`

kirszenbaum:
@sequba thank you for the input! I've committed a second draft.

docs/guide/cell-references.md
sequba: Range `Sheet1!A1:Sheet3!B2` contains cells from `Sheet1` and `Sheet3` but also from all the sheets between them (ID-wise, not name-wise!!!) so e.g. `Sheet2` assuming that 
```
getSheetId('`Sheet1') = 1
getSheetId('`Sheet2') = 2
getSheetId('`Sheet3') = 3
```

See the example with seperate sheets for each month (the syntax is a bit different but the idea is the same): https://www.ablebits.com/office-addins-blog/2015/12/09/excel-3d-reference-formula/

docs/guide/cell-references.md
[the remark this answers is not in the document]
sequba: You're right. Thanks

docs/guide/cell-references.md
sequba: Ok, the word 'cube' is inaccurate indeed. It was only a suggestion. What you propose is better.

Another idea that I have is to define a 2D range and than say, that it's also possible to span it across multiple sheets, so something like this:

> Range `<Cell address 1>:<Cell address 2>` is the smallest rectangular collection of cells including both `<Cell address 1>` and `<Cell address 2>`.
>
> Note: Ranges can also span multiple sheets. `<SheetA>!<Cell address 1>:<SheetB>!<Cell address 2>` includes the cells from `SheetA`, `SheetB` and all the sheets between them (ID-wise).

 As I said, I'm leaving the decision to you because you are in a better position to know what wording is more clear for ourusers.

sequba (changes requested):
Fix this please:
> Across **two** different sheets (=Sheet1!A1:Sheet5!B2)

kirszenbaum:
@sequba - thank you! I made the changes here: [7804c61](https://github.com/handsontable/hyperformula/pull/955/commits/7804c61026aa19363be6d42de953835b542d7805)

As discussed on Slack, I'm merging this into `feature/issue-834` (please ping me there if there are some further changes required, thanks!)

PR #995, opened 2022-05-24
Test building HF on various envs automatically

### LINKED ISSUE #974: Test building HyperFormula on different OS and different node versions automatically
### Description

Create a way of testing if building works in different environments:
- MacOS
- Windows
- Linux
- different node versions

Should it be run automatically on each PR to develop/master? Or manually before each release?

### Links
- https://handsoncode.slack.com/archives/C031GBX62SE/p1649681545935479 (internal discussion)

sequba:
### Description

TBP

### How has this been tested?

NOT TESTED YET!

### Types of changes

- [x] Internal tooling

### Related issue(s):
1. Fixes #974 
2.
3.

### Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,38 @@
+name: Build on various envs
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize # the head branch is updated from the base branch, new commits are pushed to the head branch, or the base branch is changed
+  push:
+    branches:
+      - 'master'
+      - 'develop'
+      - 'release/**'
+
+jobs:
+  build:
+    name: Build on ${{ matrix.os }} & node ${{ matrix.node-version }}
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: [ '16', '18' ]
+        os: [ ubuntu-latest, windows-latest, macos-latest ]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@722adc63f1aa60a57ec37892e133b1d319cae598 # https://github.com/actions/checkout/releases/tag/v2.0.0
+
+      - name: Setup Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@56899e050abffc08c2b3b61f3ec6a79a9dc3223d # https://github.com/actions/setup-node/releases/tag/v1.4.4
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Install dependencies
+        run: |
+          npm ci
+
+      - name: Build
+        run: |
+          npm run bundle-all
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
...
+      - 'develop'
+      - 'release/**'
+
+jobs:
+  build:
+    name: build
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: [ '14', '16' ]
```
warpech, line 21: Can you please add an issue that it does not build with Node 17 and 18? We should look into that in the next release.

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
+name: Build on various envs
```
warpech, line 1: It the PR comment, you wrote:

> Adjust repository settings to changes in the workflows.

What exactly should to be done?

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
...
+      - uses: actions/checkout@722adc63f1aa60a57ec37892e133b1d319cae598 # https://github.com/actions/checkout/releases/tag/v2.0.0
+
+      - name: Setup Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@56899e050abffc08c2b3b61f3ec6a79a9dc3223d # https://github.com/actions/setup-node/releases/tag/v1.4.4
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Install dependencies
+        run: |
+          npm ci
```
warpech, line 34: What do you think about trying both `npm ci` and `npm install`? They are very likely to result in different versions.

warpech (changes requested):

.github/workflows/build.yml
sequba: In `Branch protection rules` there is a list of workflows required to pass before merging a PR. I renamed some of the workflows so that list must be updated. Also the new workflow should be added to it.

.github/workflows/build.yml
sequba: Do you mean that we should test both ways of installing dependencies on every PR? E.g. add it to matrix in this workflow?

warpech: Yes, I think so. Using `npm install` will give us an early warning if a newer version of a dependency that matches our semver ranges in `package.json` has a problem.

sequba: Just one question - how much does GH charge us for running these workflows? Should I be concerned about it?

.github/workflows/build.yml
sequba: #996

.github/workflows/build.yml
warpech: > Just one question - how much does GH charge us for running these workflows? Should I be concerned about it?

According to GitHub docs:

> GitHub Actions usage is free for both public repositories and self-hosted runners.

So it is free, as this is a public repo. I am sure there is some unspoken limit,  but probably we are far from it.

warpech (approved):

PR #1009, opened 2022-06-09
Docs: Upddate browser list (Edge Chromium -> Edge)

kirszenbaum:
This PR:
- Updates the list of supported browsers, by changing **Edge Chromium** to **Edge**

Affected pages:
- https://handsontable.github.io/hyperformula/guide/supported-browsers.html

### DIFF
--- a/docs/guide/supported-browsers.md
+++ b/docs/guide/supported-browsers.md
@@ -7,15 +7,15 @@ users: performance and accuracy of calculations.
 
 ## List of supported browsers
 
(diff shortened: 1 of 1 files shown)


sequba (approved):

sequba:
In this guide, there is also a sentence saying:
> Each release of HyperFormula is tested on the **two latest versions**of every modern browser, on both mobile and desktop.
We don't do it, so IMO it should be removed.

PR #1000, opened 2022-06-01
Update custom-functions guide

sequba:
### Context

Changes to the `custom-functions` guide:
- start the guide with a step-by step tutorial about implementing the simplest possible function
- make it clear that defining translations is a required step
- encourage using `runFunction` helper method
- convert the code examples from Typescript to Javascript
- fix minor bugs in code examples

Other changes:
- fixa bug in production code: `implementedFunctions.parameters` should be optional
- add a few unit tests not related to custom functions. I wrote these tests while investigating other issues.

TODO:
- [ ] update hyperformula-demos

### How has this been tested?

- [ ] verify code examples in CodeSandbox
- [ ] run documentation locally and make sure it renders correctly 

### Types of changes

- [x] Docs

### Related issue(s):
1. Fixes #968 
2. https://github.com/handsontable/hyperformula/issues/965
3.

### Checklist:

- [x] My code follows the code style of this project,
- [ ] My change requires a change to the documentation,
- [x] I described the modification in the CHANGELOG.md file.

PR #1010, opened 2022-06-09
Docs: Custom functions guide rewrite (review suggestions)

kirszenbaum:
This PR:
- Adds review suggestions to #1000

My remarks:
- I added a TOC (compared to the Handsontable docs, I really miss the right-hand side navigation), but I'm not sure if it looks good, I'm leaving it up to you
- I moved the demo to the top of the **Create a custom function** section, to show the goal that we're trying to achieve upfront, and to make the page less text-heavy at first glance
- The **Create a custom function** section covers the HYPER example (which is cool), but the remaining sections (e.g., **Custom function options**) should work independently (they should be understandable without additional context, so you can start in the middle of the page, without the need to read the whole guide – this way, we can link to individual sections). That's why I changed `CountHF` to `MyFunction`, `hyper` to `myFunctionMethod` etc. (this way, we can tell straight away if sth's a part of an example, or an integral part of HyperFormula)

### DIFF
--- a/docs/guide/custom-functions.md
+++ b/docs/guide/custom-functions.md
@@ -1,100 +1,119 @@
 # Custom functions
 
-Expand the library of functions available in your app, by implementing a custom function.
+Expand the library of functions available in your app, by creating custom functions.
 
-## Custom function example
+## Create a custom function
 
-Let's create a custom function that returns the number of letters in the word "HyperFormula".
+As an example, let's create a custom function that returns the number of letters in the word "HyperFormula".
 
-### 1. Create a plugin class
+<iframe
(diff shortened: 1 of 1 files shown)


docs/guide/custom-functions.md
```diff
@@ -104,149 +126,169 @@ const result = hfInstance.getCellValue({ sheet: 0, col: 0, row: 0 });
...
+  };
+
+  translations: {
+    enGB: {
+      HYPER: 'HYPER',
+    };
+  };
 
-Passing optional parameters to the `implementedFunctions` object, you can configure your function to:
-* Use the [array arithmetic mode](arrays.md)
```
sequba, line 120: This is missing

sequba (approved):
Great job! It's much better than my version. Only a few minor comments and we can merge it.

kirszenbaum:
@sequba - thanks! Feedback applied

sequba (approved):

PR #1021, opened 2022-06-28
Fix SUMIF to handle date/time values, percent values & currency values

### LINKED ISSUE #951: SUMIF not working on dates
### Description

SUMIF does not work when dates are included in the conditional values range and the criterion is also a date. It seems to be failing regardless of whether the date has been parsed using dateFormats.

### Steps to reproduce
```
      ['1', '4', 43384, '25000'],
      ['2', '2', '11/10/2018', '15000'],
      ['=SUMIF(C1:C3, "11/10/2018" ,D1:D3)']
    ], {dateFormats: ['DD/MM/YYYY']})
    expect(engine.getCellValue(adr('A3'))).toEqual(40000)
```

```
      ['1', '4', 43384, '25000'],
      ['2', '2', '11/10/2018', '15000'],
      ['=SUMIF(C1:C3, 43384 ,D1:D3)']
    ], {dateFormats: ['DD/MM/YYYY']})
    expect(engine.getCellValue(adr('A3'))).toEqual(40000)
```

Both tests fail with a result of 25000

MBP 2021
macOS 12.3
node v16.13.0
npm v8.1.0

HF version: 1.3.1

Seems like the issue is that DateNumbers are not RawScalarValues from here
https://github.com/handsontable/hyperformula/blob/fd0a40581d1dea2a19c2307228825ad0c3b6dedb/src/interpreter/CriterionFunctionCompute.ts#L160

sequba:
#### The cause of the bug
Implementation for SUMIF (and other similar functions) assumes that the values inside the criterion range (1st argument to SUMIF) are simple scalars (integers, strings, etc.). It fails for complex numeric values e.g. dates, 

#### The fix
When evaluating the criteria, map the complex value (date) to its simple representation (number).

#### Affected functions
- SUMIF
- SUMIFS
- AVERAGEIF
- COUNTIF
- COUNTIFS

#### Affected data types
- date
- time
- datetime
- currency
- percent

### How has this been tested?

- Existing unit tests passed
- Added new unit tests to verify issue #951

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)

### Related issues:
Fixes #951 

### Checklist:

- [ ] My change requires a change to the documentation.
- [ ] My change requires a migration guide.


test/interpreter/function-sumif.spec.ts
```diff
@@ -452,6 +452,28 @@ describe('Function SUMIF(S) - calculations and optimizations', () => {
...
+    const engine = HyperFormula.buildFromArray([
+      ['31/05/2019', '1'],
+      ['=DATEVALUE("31/05/2019")', '1'],
+      ['=DATE(2019, 5, 31)', '1'],
+      ['=A1', '1'],
+      ['43616', '1'],
+      ['=SUMIF(A1:A5, "31/05/2019", B1:B5)']
+    ], {dateFormats: ['DD/MM/YYYY']})
+
+    expect(engine.getCellValue(adr('A6'))).toEqual(5)
```
sequba, line 475: Before the fix, this assertion was failing with value `1`, because it was matching only the integer (line 471).

BrianHung:
@sequba Can you check that it works for booleans too? Right now, a simple `=SUMIFS(B:B, true)` doesn't work and returns `#NAME?` while `=SUMIFS(B:B, "true")` computes and returns `0`.

BrianHung:
@sequba #1029 Oops, meant `SUMIF`. Provided a sandbox in #1029.

test/interpreter/function-sumif.spec.ts
```diff
@@ -452,6 +452,71 @@ describe('Function SUMIF(S) - calculations and optimizations', () => {
...
+    expect(engine.getCellValueDetailedType(adr('A1'))).toEqual(CellValueDetailedType.NUMBER_DATETIME)
+    expect(engine.getCellValue(adr('A2'))).toEqual(1)
+  })
+
+  it('works when criterion arg is a currency value', () => {
+    const engine = HyperFormula.buildFromArray([
+      ['$1', '1'],
+      ['1$', '1'],
+      ['=SUMIF(A1:A2, "$1", B1:B2)'],
+      ['=SUMIF(A1:A2, "1$", B1:B2)']
```
warpech, line 503: Because of the text trimming, the following criterions also succeed:

```
      ['=SUMIF(A1:A2, "  $1", B1:B2)'], // start with whitespace
      ['=SUMIF(A1:A2, "1$  ", B1:B2)'] // end with whitespace
```

However:
1. is that according to the spec?
1. if yes, then a test for it is missing
1. if yes, then it seems that Excel has slightly different implementation. In my Excel, start with whitespace succeeds, but end with whitespace fails.

test/interpreter/function-sumif.spec.ts
```diff
@@ -452,6 +452,71 @@ describe('Function SUMIF(S) - calculations and optimizations', () => {
...
+
+    expect(engine.getCellValueDetailedType(adr('A1'))).toEqual(CellValueDetailedType.NUMBER_CURRENCY)
+    expect(engine.getCellValue(adr('A3'))).toEqual(2)
+    expect(engine.getCellValue(adr('A4'))).toEqual(2)
+  })
+
+  it('works when criterion arg is a percent value', () => {
+    const engine = HyperFormula.buildFromArray([
+      ['20%', '1'],
+      ['=SUMIF(A1:A1, "20%", B1:B1)']
```
warpech, line 514: Because of the text trimming, the following criterion also succeeds:

```
      ['=SUMIF(A1:A1, "20  %  ", B1:B1)']
```

However:
1. is that according to the spec?
   - it seems it is compatible with Excel, because whitespace on both sides of % also succeeds there
1. if yes, then a test for it is missing

warpech (changes requested):
Some observations regarding the test suite

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Fixed functions SUMIF, SUMIFS, AVERAGEIF, COUNTIF, COUNTIFS to handle complex numeric values correctly. [#951](https://github.com/handsontable/hyperformula/issues/951)
```
warpech, line 10: As a sidenote, can you please check the performance of this PR against develop and share the results?

sequba:
## Whitespaces inside the _Criterion_ string

### ODFF
ODFF says nothing about whitespace characters in the [description of the _Criterion_](https://docs.oasis-open.org/office/OpenDocument/v1.3/os/part4-formula/OpenDocument-v1.3-os-part4-formula.html#__RefHeading__434502_49873571), but in the [general section about whitespaces](https://docs.oasis-open.org/office/OpenDocument/v1.3/os/part4-formula/OpenDocument-v1.3-os-part4-formula.html#__RefHeading__1017970_715980110) it states:
> For calculation purposes, whitespace is ignored unless it is inside the contents of string constants or text surrounded by single quotes. **Evaluators shall ignore any whitespace characters before and/or after any operators, constant numbers**, constant strings, constant errors, inline arrays, parentheses used for controlling precedence, and the closing parenthesis of a function call. Whitespace shall be ignored following the initial equal sign(s). Whitespace shall be ignored just before a function name, but whitespace shall not separate a function name from its initial opening parenthesis. Whitespace shall not be used in the interior of a terminating grammar rule (a rule that references no other rule other than character sets, internally or externally-defined), unless specifically permitted by the terminating grammar rule, since these rules define the lexical properties of a component.

My interpretation of this description (confirmed by the LibreOffice Calc implementation) is that:
- `" 1$"`, `"1$ "`, `"1 $"` and `"1$"` shall be all semantically equivalent
- `" 20%"`, `"20% "`, `"20 %"` and `"20%"` shall be all semantically equivalent

### Excel
I verified that Excel also ignores the whitespaces in all the examples mentioned above.

> it seems that Excel has slightly different implementation. In my Excel, start with whitespace succeeds, but end with whitespace fails.

My tests on Windows & Excel Version 2206 were successful in both scenarios.

test/interpreter/function-sumif.spec.ts
sequba: I [investigated it](https://github.com/handsontable/hyperformula/pull/1021#issuecomment-1195773214) and added a test to make sure whitespace characters are ignored when evaluating the criterion string.

test/interpreter/function-sumif.spec.ts
sequba: I [investigated it](https://github.com/handsontable/hyperformula/pull/1021#issuecomment-1195773214) and added a test to make sure whitespace characters are ignored when evaluating the criterion string.

sequba:
## Performance test

I ran the performance benchmark for **develope** and **feature/issue-951** (this PR) on my laptop. There is no significant performance difference. In my opinion these benchmarks have little practical value without the controlled environment for running them and a statistical methodology of interpreting the results

#### feature/issue-951
```
┌─────────┬─────────────────┬───────────┐
│ (index) │      name       │ totalTime │
├─────────┼─────────────────┼───────────┤
│    0    │    'Sheet A'    │   277.4   │
│    1    │    'Sheet B'    │   116.3   │
│    2    │    'Sheet T'    │   160.8   │
│    3    │ 'Column ranges' │   328.7   │
└─────────┴─────────────────┴───────────┘


┌─────────┬─────────────────────────────────────────────────┬───────────┐
│ (index) │                      name                       │ totalTime │
├─────────┼─────────────────────────────────────────────────┼───────────┤
│    0    │ 'Sheet A:  change value, add/remove row/column' │    25     │
│    1    │ 'Sheet B: change value, add/remove row/column'  │    173    │
│    2    │          'Column ranges - add column'           │    133    │
│    3    │         'Column ranges - without batch'         │    370    │
│    4    │             'Column ranges - batch'             │    87     │
└─────────┴─────────────────────────────────────────────────┴───────────┘

```
#### develop
```
┌─────────┬─────────────────┬───────────┐
│ (index) │      name       │ totalTime │
├─────────┼─────────────────┼───────────┤
│    0    │    'Sheet A'    │   278.6   │
│    1    │    'Sheet B'    │    118    │
│    2    │    'Sheet T'    │   159.7   │
│    3    │ 'Column ranges' │   334.8   │
└─────────┴─────────────────┴───────────┘

┌─────────┬─────────────────────────────────────────────────┬───────────┐
│ (index) │                      name                       │ totalTime │
├─────────┼─────────────────────────────────────────────────┼───────────┤
│    0    │ 'Sheet A:  change value, add/remove row/column' │    25     │
│    1    │ 'Sheet B: change value, add/remove row/column'  │    174    │
│    2    │          'Column ranges - add column'           │    148    │
│    3    │         'Column ranges - without batch'         │    358    │
│    4    │             'Column ranges - batch'             │    94     │
└─────────┴─────────────────────────────────────────────────┴───────────┘

CHANGELOG.md
sequba: https://github.com/handsontable/hyperformula/pull/1021#issuecomment-1196475510

warpech (approved):
Great, thanks!

PR #1042, opened 2022-08-03
GH workflow that measures the performance

sequba:
### Changes

- node script that runs all benchmarks and writes the results to file
- node script that compares the results from a few files and prints the report to a file
- GH workflow that uses these scripts and posts the result as a PR comment
- **new dev-dependency**: [as-table](https://www.npmjs.com/package/as-table) - a tiny library to format data as a plain-text table

### How has this been tested?

On this PR with `feature/issue-973-fake-develop` as a base branch instead of develop. It doesn't work with `develop` until this PR is merged.

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] New feature or improvement (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Additional language file or change to the existing one (translations)
- [ ] Change to the documentation
- [x] Internal tooling

### Related issues:
Fixes #973

### Checklist:

- [x] My code follows the code style of this project.
- [ ] I described the modification in the CHANGELOG.md file.
- [ ] My change requires a change to the documentation.
- [ ] My change requires a migration guide.

PR #1051, opened 2022-08-23
GH workflow that measures the performance

### LINKED ISSUE #973: Run performance benchmark automatically for each PR
### Description

Run performance benchmark automatically for each PR. Something similar is done in Spreadsheet Viewer repo. It might be helpful

### Links
- https://github.com/handsontable/hyperformula/pull/938#issuecomment-1090052862
- https://handsoncode.slack.com/archives/D02T462B3GS/p1648823588399299 (internal discussion - accessible by @sequba )

sequba:
### Changes

- node script that runs all benchmarks and writes the results to file
- node script that compares the results from a few files and prints the report to a file
- GH workflow that uses these scripts and posts the result as a PR comment
- **new dev-dependency**: [as-table](https://www.npmjs.com/package/as-table) - a tiny library to format data as a plain-text table

### How has this been tested?

On this PR with `feature/issue-973-fake-develop` as a base branch instead of develop. It doesn't work with `develop` until this PR is merged.

### Types of changes

- [x] Internal tooling

### Related issues:
Fixes #973

### Checklist:

- [ ] My change requires a change to the documentation.
- [ ] My change requires a migration guide.


sequba:
I accidentally "broke" pull request https://github.com/handsontable/hyperformula/pull/1042. Sorry about that. This is a copy.

In the previous PR @warpech wrote:
> Great job! Just a small request
> 
> - sample output presented in this PR looks fine, apart from a minor comment
> - `as-table` new dev-dependency looks fine, when it comes to the license (and the license of its dependencies) and security
> - `performance.yml` looks fine
> - ran `benchmark:basic`, `benchmark:cruds`, `benchmark:write-to-file`, `benchmark:compare-benchmarks` on macOS

And:

> Since this job will not run on every commit, could the results report contain the Git SHA of the versions used for the comparison?
> 
> For reference, the Codecov comment contains information about the version (see the below screenshot).
> 
> I don't think we need the PR number or the branch name there. Just a simple Git SHA will suffice for me.
> 
> Screen Shot 2022-08-23 at 3 05 32 PM

I added commit SHA's to the performance report:
image

.github/workflows/performance.yml
```diff
@@ -0,0 +1,66 @@
...
+        uses: marocchino/sticky-pull-request-comment@6804b5ad49d19c10c9ae7cf5057352f7ff333f31 # https://github.com/marocchino/sticky-pull-request-comment/tree/v1.6.0
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          message: |
+            # Performance comparison of head (${{ github.event.pull_request.head.sha }}) vs base (${{ github.event.pull_request.base.sha }})
+
+      - name: Publish a comment - performance comparison report
+        uses: marocchino/sticky-pull-request-comment@6804b5ad49d19c10c9ae7cf5057352f7ff333f31 # https://github.com/marocchino/sticky-pull-request-comment/tree/v1.6.0
+        with:
+          append: true
```
warpech, line 64: `append: true` will append new results at the end of the comment after each workflow run. Is this on purpose?

.github/workflows/performance.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Compare the results
+        run: |
+          npm run benchmark:compare-benchmarks base.json head.json performance-report.md
+
+      - name: Publish a comment - header
+        uses: marocchino/sticky-pull-request-comment@6804b5ad49d19c10c9ae7cf5057352f7ff333f31 # https://github.com/marocchino/sticky-pull-request-comment/tree/v1.6.0
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          message: |
+            # Performance comparison of head (${{ github.event.pull_request.head.sha }}) vs base (${{ github.event.pull_request.base.sha }})
```
warpech, line 59: Can you please change `#` to `##`, so that the header text will fit in one line?

warpech (changes requested):

.github/workflows/performance.yml
sequba: done

.github/workflows/performance.yml
sequba: This workflow runs `marocchino/sticky-pull-request-comment` twice. First time with `append: false` to create a comment (or overwrite if it already exists) and second time with `append: true` to add new information to the previously created comment.

warpech: True, thanks for explanation

warpech (approved):
Looking good now!

PR #1117, opened 2022-12-01
Test the new LGTM integration

warpech:
add an unused import to see if it will trigger a code quality alert


warpech:
Test shows that CodeQL was able to detect a problem. Cool!

PR #1183, opened 2023-03-09
Add checkbox about CLA to the pull request template

sequba:
### Context

HOT has such a checkbox. HF will also benefit from it.

### Types of changes

- [x] Internal tooling


sequba:
Opened by mistake

Issue #885, opened 2022-01-16
information plugin has incorrect `optional: true` argument


MartinDawson:
```
export class InformationPlugin extends FunctionPlugin implements FunctionPluginTypecheck<InformationPlugin> {
  public static implementedFunctions = {
    'COLUMN': {
      method: 'column',
      parameters: [
        {argumentType: ArgumentTypes.NOERROR, optional: true}
      ],
      isDependentOnSheetStructureChange: true,
      doesNotNeedArgumentsToBeComputed: true,
      vectorizationForbidden: true,
    },
```

`optional:true` -> `optionalArg: true`

Same bug is also on the 'ROW' Function.

`optional` isn't an actually doing anything here 

Test descriptions are also wrong for it:

```
it('should take one or zero arguments', () => {
    const [engine] = HyperFormula.buildFromArray({ cells: [
      [{ cellValue: '=COLUMN(B1, B2)' }],
      [{ cellValue: '=COLUMN(B1, B2, B3)' }],
    ]})

```

sequba:
I am sorry you experienced an issue with HyperFormula. Our team will investigate it.

thilgen:
PR submitted with proposed fix - https://github.com/handsontable/hyperformula/pull/1227

PR #1227, opened 2023-04-03
Fix for #885

thilgen:
### Context
Fix for https://github.com/handsontable/hyperformula/issues/885

The following `FUNCTIONS`
```
FORMULATEXT
ISFORMULA
COLUMN
ROW
SHEET
SHEETS
``` 
* All make use of `FunctionPlugin.runFunctionWithReferenceArgument()` - which does its own pre-check of arguments before calling `FunctionPlugin.runFunction()` (which most all the other `FUNCTIONS` invoke).
* Because of this, any of these `FUNCTIONS` invoked with zero parameters will short-circuit the arg-checking that happens in `FunctionPlugin.protected runFunction()`
* I believe these functions are working properly with their own checking, but this is definitely sort of unexpected behavior when looking at the pervasive pattern of `FunctionArgument` blocks that are used to map `FUNCTIONS` to their handlers
* Ideally `FunctionPlugin.runFunctionWithReferenceArgument()` would not exist - but I couldn't find a non-scary way of adding the overloaded arg behavior that these functions utilize with the existing built-in arg checking in `FunctionPlugin.runFunctionWithReferenceArgument()` - so I just removed the extraneous/ignored `optional` config option.
* I also couldn't find a way to add a unit test that would validate that all the properties declared on a function parameter (sort of like how the i18n code is validated) matched the properties in `FunctionArgument` since `FunctionArgument` is an interface (with a bunch of optionals that disappear after compilation)
* The comment for the text block is actually correct - the test is checking for the expected failure when more than one argument is passed - but admittedly is kinda confusing - so tweaked that.

`npm:test`

### Types of changes

- [X] Bug fix (a non-breaking change that fixes an issue)

### Related issues:
1. https://github.com/handsontable/hyperformula/issues/885

### Checklist:

- [X] My change is compliant with the [OpenDocument](https://docs.oasis-open.org/office/OpenDocument/v1.3/os/part4-formula/OpenDocument-v1.3-os-part4-formula.html) standard.
- [ ] My change is compatible with Microsoft Excel.
- [ ] My change is compatible with Google Sheets.
- [X] My code follows the code style of this project.
- [ ] I described my changes in the [CHANGELOG.md](https://github.com/handsontable/hyperformula/blob/master/CHANGELOG.md) file.


sequba (changes requested):
Thanks for helping us with this issue. The arguments should be marked as optional. The problem is that the name of the flag is invalid.

sequba:
@thilgen 
> I also couldn't find a way to add a unit test that would validate that all the properties declared on a function parameter (sort of like how the i18n code is validated) matched the properties in FunctionArgument since FunctionArgument is an interface (with a bunch of optionals that disappear after compilation)

We can use the type checker for that. Try adding a type annotation in line 19 of `InformationPlugin.ts`:
```ts
  public static implementedFunctions: ImplementedFunctions = {
```

thilgen:
@sequba 

> Thanks for helping us with this issue. The arguments should be marked as optional. The problem is that the name of the flag is invalid.

Apologies - I probably did not do a great job explaining why I did not just change the property as you suggest.

If you take a look at the `FUNCTIONS` that make use of `FunctionPlugin.runFunctionWithReferenceArgument()` rather than the far-more-common `FunctionPlugin.runFunction()` - you see that it is doing its own argument checking+validation that bypasses the checking done in `FunctionPlugin.runFunction()` (which looks at the `FunctionArgument` properties)

```
FunctionPlugin.runFunctionWithReferenceArgument(
  ...
  noArgCallback
  referenceCallback
  nonReferenceCallback
) {

  // if args.length === 0

     -> Always calls FunctionPlugin.returnNumberWrapper(noArgCallback)
        FunctionPlugin.returnNumberWrapper() does no checking+validation of FunctionArgument properties

  // if args.length > 1

     -> Always returns CellError

  // if args.length === 1

     -> Depending on type of args[0]...

       -> Calls FunctionPlugin.returnNumberWrapper(referenceCallback)
        FunctionPlugin.returnNumberWrapper() does no checking+validation of FunctionArgument properties

       -> Calls FunctionPlugin.runFunction(nonReferenceCallback) - passes-through args, state, and metadata
        FunctionPlugin.runFunction() does check+validate the FunctionArgument properties

       -> Returns CellError
}
```
`ROW` and `COLUMN` are two `FUNCTIONS` that make use of `FunctionPlugin.runFunctionWithReferenceArgument)` - and in most cases will never fall into the `FunctionPlugin.runFunction()` branch of the code - so the properties of `FunctionArgument` are never examined nor validated.

I do thing this is pretty not-great - the overwhelming pattern that exists for `FunctionPlugins` makes use of the `FunctionArgument` properties to validate+check arguments - but for the cases mentioned above - the `FUNCTION` handler is electing to process things in a different way - which is pretty confusing as expectations don't meet the reality.

Since I couldn't figure out a clean way to untangle these - **eliminating the need for `FunctionPlugin.runFunctionWithReferenceArgument()` still feels like the best resolution** - I opted instead to remove the largely-ignored  `optional` property - rather than change it to `optionalArg` which would also be largely-ignored - and so that there would be less confusion for maintainers that the property in `FunctionArgument` actually had any affect.

thilgen:
> We can use the type checker for that. Try adding a type annotation in line 19 of `InformationPlugin.ts`:

Ah. Brilliant. Showing my relative inexperience with `TypeScript`. Will make that change in another PR as it is a bunch of boilerplate changes to the `FunctionPlugins`

thilgen:
> > We can use the type checker for that. Try adding a type annotation in line 19 of `InformationPlugin.ts`:
> 
> Ah. Brilliant. Showing my relative inexperience with `TypeScript`. Will make that change in another PR as it is a bunch of boilerplate changes to the `FunctionPlugins`

PR Submitted: https://github.com/handsontable/hyperformula/pull/1228

sequba:
@thilgen Ok, I get it.

I agree that the implementation of `runFunctionWithReferenceArgument` that ignores the function metadata is a wrong solution. But if we decide to refactor it in the future, it's better to have these arguments already marked as optional. `implementedFunctions` property is also helpful as documentation of these functions.

sequba:
Since #1228 contains a sufficient fix for #885, I'm closing this PR.

Issue #1253, opened 2023-04-25
Custom func that returns an array doesn't seem to resize properly


qcurtemanjc:
### Description
Custom function that returns a `SimpleRangeValue` value doesn't recalculate using the `arraySizeFunc` first and it breaks. Not sure if this is a bug or if I have my set up incorrect.

### Steps to reproduce
Using this codepen: https://codepen.io/qcurtemanjc/pen/abRWzbK?editors=1111
In cell `A1`, try to add another value (i.e. enter `1,2,3,4` into cell `A1`)

Expected: 
In row 2, it should have 1 2 3 4 in incremental column cells respectively

Actual:
It breaks.

### Demo
https://codepen.io/qcurtemanjc/pen/abRWzbK?editors=1111

### Your environment

* HyperFormula version: 2.4.0
* Browser Name and version: Chrome (`104.0.5112.79`)
* Operating System: Mac OS

### Note:
My chrome console throws an error when following the steps to reproduce but code pen doesn't for some reason. Here's the error in chrome:
```
hyperformula.full.min.js:815 Uncaught Error: Resizing to smaller array
    at i.resize (hyperformula.full.min.js:815:1693)
    at h.setCellValue (hyperformula.full.min.js:810:1074)
    at t.Evaluator.recomputeFormulaVertexValue (hyperformula.full.min.js:1004:2362)
    at hyperformula.full.min.js:1004:628
    at hyperformula.full.min.js:825:2965
    at Array.forEach (<anonymous>)
    at t.Graph.getTopSortedWithSccSubgraphFrom (hyperformula.full.min.js:825:2833)
    at hyperformula.full.min.js:1004:503
    at o.measure (hyperformula.full.min.js:786:1166)
    at t.Evaluator.partialRun (hyperformula.full.min.js:1004:441)
```

adrianszymanski89:
Hi @qcurtemanjc 

It looks like there is an issue in this part:

`return new HyperFormula.ArraySize(data?.length, 1);`

As the first parameter doesn't resize the array correctly. If you change it to the fixed value larger than the initial data in the first cell, then it works correctly.

qcurtemanjc:
Hey @adrianszymanski89 

Thanks for the reply. I'm not sure that setting the first parameter to a larger fixed value will always work. If there is a value in the arrays "path" it will cause a `#SPILL!` error.

Example:
I changed it to this `return new HyperFormula.ArraySize(100, 1);`
[...]

PR #1261, opened 2023-05-06
Describe the limitation of the `ArraySize` method in the docs

sequba:
### Context

Our user asked [a question](https://github.com/handsontable/hyperformula/issues/1253) about this problem so it is beneficial to add this info to documentation.

I described this limitation in the `custom-functions` guide. it could be added to `known-limitations` page also/instead ?

### Types of changes

### Related issues:
1. https://github.com/handsontable/hyperformula/issues/1253
2.
3.

### Checklist:

- [x] I have reviewed the guidelines about [Contributing to HyperFormula](https://hyperformula.handsontable.com/guide/contributing.html) and I confirm that my code follows the code style of this project.
- [ ] I described my changes in the [CHANGELOG.md](https://github.com/handsontable/hyperformula/blob/master/CHANGELOG.md) file.


docs/guide/custom-functions.md
```diff
@@ -262,6 +262,11 @@ declare a companion method for the array size. To do that, provide the
 function arguments and returns an instance of the
 [`ArraySize` class](../api/classes/arraysize.md).
 
+::: tip
+The `arraySizeMethod` is called everytime the cell formula changes, but not when its dependencies change.
+This may cause unexpected behavior when the size of the result array depends on the values of the cells referenced in the formula.
+:::
```
sequba, line 268: @kirszenbaum Do you think it is better to keep it here or in the `known-limitations` guide?

kirszenbaum: I would do both and link between them: `custom-functions` should include this information because it's important to implementing a custom range function. At the same time, `known-limitations` should list all known limitations.

In `custom-functions`, I'd add sth like (If I understand it correctly):

> When you use your custom function in a formula, `arraySizeMethod` is triggered every time the formula changes, but not when the dependencies of the formula change. This can cause unexpected results if the size of the result array depends on the values in the referenced cells.

I'd also change "A function may return multiple values in the form of an array." to "A function **can** return multiple values in the form of an array."

In `known-limitations` (perhaps in a separate "Custom functions" section?), I'd add sth like:

> Custom functions don't automatically recalculate the size of the result arrays when formula dependencies change.

WDYT?

sequba: I changed it. WHYT now?

kirszenbaum (approved):

PR #1564, opened 2025-10-15
Configure eslint for spec files

sequba:
### Context

Install and configure rules for eslint-plugin-jest and eslint-plugin-jasmine

unit tests, linter


AMBudnik:
Fixed in v3.1.1

Issue #1635, opened 2026-03-10
[Bug]: `ParserWithCaching` cache can grow without limit


sequba:
### Description

See parent task

### Demo

See parent task

### HyperFormula version

See parent task

### Your environment

See parent task

PR #1638, opened 2026-03-17
Fix #1629: resolve memory leaks in LazilyTransformingAstService and UndoRedo

### LINKED ISSUE #1629: [Bug]: Memory leaks in UndoRedo.oldData, LazilyTransformingAstService.transformations, and Parser.Cache when undoLimit is 0
### Description

# Memory leak: `LazilyTransformingAstService.transformations` and `UndoRedo.oldData` grow unboundedly when `undoLimit` is 0

## Description

When `undoLimit` is set to `0`, two internal data structures grow without bound, causing memory leaks in long-running applications. The `undoLimit` configuration only controls the size of `undoStack` (via `addUndoEntry`), but does **not** prevent accumulation in `transformations` or `oldData`.

### 1. `LazilyTransformingAstService.transformations` (Critical)

**`transformations: FormulaTransformer[]`** grows without bound.

- **Write path**: `addTransformation()` pushes to the array on every structural change (add/remove rows/columns, move cells).
- **Cleanup**: None. The array is only initialized to `[]` in the constructor and is never truncated.
- **Version coupling**: `version()` returns `transformations.length`, and both `FormulaVertex.version` and `ColumnIndex.ValueIndex.version` store indices into this array for lazy transformation application.

After each operation, `recomputeIfDependencyGraphNeedsIt()` triggers `FormulaVertex.ensureRecentData()` which consumes all pending transformations. **Once consumed, old transformer objects are never needed again**, but they remain in the array permanently.

### 2. `UndoRedo.oldData` (Critical)

**`oldData: Map<number, [SimpleCellAddress, string][]>`** grows without bound.

- **Write path**: `storeDataForVersion()` is called from `applyTransformations()` for every irreversible transformation (`RemoveRows`, `RemoveColumns`, `MoveCells`, `CombinedTransformer`).
- **Cleanup**: None. `restoreOldDataFromVersion()` only **reads** from the map — it never calls `this.oldData.delete(version)`.
- **Problem with `undoLimit=0`**: `storeDataForVersion()` has no awareness of `undoLimit` and continues accumulating data that can never be consumed (since undo is disabled).

### 3. `ParserWithCaching` cache (Minor)

**`cache: Map<string, CacheEntry>`** has no eviction mechanism. When formula references shift due to structural changes, new AST hashes are generated and cached indefinitely.

## Reproduction

```javascript
const { HyperFormula } = require('hyperformula');

const hf = HyperFormula.buildFromArray(
  [
    ['=SUM(A2:A5)', '=SUM(B2:B5)'],
    [1, 2],
    [3, 4],
    [5, 6],
    [7, 8],
  ],
  { licenseKey: 'gpl-v3', undoLimit: 0 }
);

const ltas = hf._lazilyTransformingAstService;

console.log('=== Before operations ===');
console.log('transformations.length:', ltas.transformations.length);
console.log('oldData.size:', ltas.undoRedo.oldData.size);
console.log('cache.size:', ltas.parser.cache.cache.size);

const ITERATIONS = 1000;
for (let i = 0; i < ITERATIONS; i++) {
  // Insert row at index 1 (INSIDE the SUM range, forces formula recalc)
  hf.addRows(0, [1, 1]);
  // Set values so the row is not empty
  hf.setCellContents({ sheet: 0, col: 0, row: 1 }, [[10, 20]]);
  // Remove the row (irreversible, formulas are dirty and get re-evaluated)
  hf.removeRows(0, [1, 1]);
}

console.log('\n=== After ' + ITERATIONS + ' cycles ===');
console.log('transformations.length:', ltas.transformations.length);
console.log('oldData.size:', ltas.undoRedo.oldData.size);
console.log('cache.size:', ltas.parser.cache.cache.size);

// rebuildAndRecalculate forces all lazy transformations to be applied,
// which triggers storeDataForVersion for irreversible transformers
hf.rebuildAndRecalculate();

console.log('\n=== After rebuildAndRecalculate (old LTAS) ===');
console.log('transformations.length:', ltas.transformations.length);
console.log('oldData.size:', ltas.undoRedo.oldData.size);
console.log('cache.size:', ltas.parser.cache.cache.size);

// rebuildAndRecalculate creates a new LTAS, so the new one is clean
const newLtas = hf._lazilyTransformingAstService;
console.log('\n=== After rebuildAndRecalculate (new LTAS) ===');
console.log('transformations.length:', newLtas.transformations.length);
console.log('oldData.size:', newLtas.undoRedo.oldData.size);
console.log('cache.size:', newLtas.parser.cache.cache.size);
```

**Output:**

```
=== Before operations ===
transformations.length: 0
oldData.size: 0
cache.size: 1

=== After 1000 cycles ===
transformations.length: 2000
oldData.size: 0
cache.size: 1

=== After rebuildAndRecalculate (old LTAS) ===
transformations.length: 2000
oldData.size: 1000
cache.size: 2

=== After rebuildAndRecalculate (new LTAS) ===
transformations.length: 0
oldData.size: 0
cache.size: 1
```

**Key observations:**

1. `transformations` grows by 2 per iteration (one for `addRows`, one for `removeRows`) — **2000 consumed but never freed**
2. `oldData` is 0 after the loop because lazy transformations on formulas outside the dirty range haven't been triggered yet. After `rebuildAndRecalculate()` forces evaluation, `oldData` jumps to **1000** (one entry per `removeRows`, which is irreversible)
3. `rebuildAndRecalculate()` creates a new LTAS internally, so the leaked data from the old instance lingers until GC
4. The old LTAS with 2000 transformations and 1000 oldData entries is **pure waste** — `undoLimit=0` means this data will never be used for undo

### Option 1: `versionOffset` + compaction for `transformations`

Introduce a `versionOffset` so the array can be truncated while maintaining correct version semantics:

```typescript
private versionOffset: number = 0

public version(): number {
  return this.versionOffset + this.transformations.length
}

public compact(): void {
  // Call after forceApplyPostponedTransformations()
  this.versionOffset += this.transformations.length
  this.transformations = []
}
```

### Option 2: Skip `storeDataForVersion` when `undoLimit === 0`

```typescript
public storeDataForVersion(version: number, address: SimpleCellAddress, astHash: string) {
  if (this.undoLimit === 0) return  // No undo data needed
  // ... existing logic
}
```

### Option 3: Expose a public `compact()` / `clearTransformations()` API

At minimum, provide a method for users who have disabled undo to manually reclaim memory.

## Environment

- **HyperFormula version**: 3.2.0
- **Runtime**: Node.js (server-side, long-running process)
- **Configuration**: `undoLimit: 0`

## Expected Behavior

When `undoLimit` is set to `0`, consumed transformations and undo-only data structures (`oldData`) should not accumulate indefinitely, or there should be a public API to compact/clear them.

### Video or screenshots

_No response_

### Demo

https://stackblitz.com/edit/node-oqngupra?file=index.js

### HyperFormula version

3.2.0

### Your framework

NodeJS v24.12.0

### Your environment

Ubuntu 22.04.4 LTS

### LINKED ISSUE #1633: [Bug]: `LazilyTransformingAstService.transformations` grows without bound until recalculation is triggered
### Description

See parent task

### Video or screenshots

_No response_

### Demo

See parent task

### HyperFormula version

See parent task

### Your framework

_No response_

### Your environment

See parent task

### LINKED ISSUE #1634: [Bug]: `UndoRedo.oldData` is not bounded by `undoLimit`
### Description

See parent task

### Video or screenshots

_No response_

### Demo

See parent task

### HyperFormula version

See parent task

### Your framework

_No response_

### Your environment

See parent task

marcin-kordas-hoc:
## Summary

Fixes #1629. Closes #1633. Closes #1634.

Two unbounded memory leaks in long-running HyperFormula instances:

- **`LTAS.transformations[]`** grew linearly with every structural operation (addRows, removeRows, moveCells, etc.) and was never cleaned up. Fixed by introducing **threshold-based compaction** with `versionOffset` — once 50+ transformations accumulate, all consumers (FormulaVertex, ColumnIndex) are force-updated, then the array is released while the logical version remains monotonically increasing.

- **`UndoRedo.oldData`** grew linearly even when entries were evicted from the undo stack. Fixed by tracking which LTAS versions each `UndoEntry` references (`getReferencedOldDataVersions()`), cleaning up on eviction/clear, guarding against writes when `undoLimit === 0`, and running orphan cleanup after compaction to handle a race condition where lazy-apply re-inserts already-evicted keys.

### Changed files

| File | Change |
|------|--------|
| `LazilyTransformingAstService.ts` | `versionOffset`, `compact()`, `needsCompaction()` with threshold=50, offset-aware iteration |
| `UndoRedo.ts` | `getReferencedOldDataVersions()` on interface + 7 subclasses, eviction cleanup, `undoLimit===0` guard, `cleanupOrphanedOldData()`, `forceApply` parity in `undoMoveRows`/`undoMoveColumns` |
| `HyperFormula.ts` | Compaction trigger in `recomputeIfDependencyGraphNeedsIt()` |
| `ColumnIndex.ts` | `forceApplyPostponedTransformations()` — iterates all ValueIndex entries |
| `ColumnBinarySearch.ts` | No-op `forceApplyPostponedTransformations()` |
| `SearchStrategy.ts` | New method on `ColumnSearchStrategy` interface |
| `Operations.ts` | Added `columnSearch.forceApply` to undo path (no compact — centralized in HyperFormula.ts) |

### What is NOT fixed here

- **Parser cache** (`ParserWithCaching`) — unbounded growth tracked separately in #1635

## Test plan

- 18 dedicated tests in `hyperformula-tests` — see companion PR in that repo
- Full test suite: **480 suites / 5396 tests passed**
- Benchmark validated threshold=50 as optimal (eager compaction is ~18× slower on a 2.5k formula sheet)

---

> [!NOTE]
> **Medium Risk**
> Touches core recalculation/transform and undo/redo paths; while aimed at memory safety, compaction/cleanup timing could affect formula correctness or undo behavior in edge cases.
> 
> **Overview**
> Prevents unbounded memory growth in long-running engines by **adding threshold-based compaction** of lazy formula transformations and by **cleaning up undo snapshot (`oldData`) entries** when undo/redo stack entries are cleared or evicted.
> 
> Introduces new config `maxPendingLazyTransformations` (default `50`) and wires it into engine construction; when the threshold is reached, `HyperFormula` forces pending transformations to be applied (dependency graph + column search), compacts the transformation history, and prunes orphaned `UndoRedo.oldData`. Column search strategies now expose `forceApplyPostponedTransformations()` (real implementation for `ColumnIndex`, no-op for binary search), and undo for move operations ensures postponed transformations are applied before restoring old data. Documentation and changelog are updated accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d8ebe1d1f59ca7e9de9493154b6d36c211a4b3d2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/Lookup/ColumnBinarySearch.ts
```diff
@@ -53,6 +53,9 @@ export class ColumnBinarySearch extends AdvancedFind implements ColumnSearchStra
   public removeValues(range: IterableIterator<[RawScalarValue, SimpleCellAddress]>): void {
   }
 
+  public forceApplyPostponedTransformations(): void {
```
sequba, line 63: Please, add a comment to explain why we don't need to do anything here

src/LazilyTransformingAstService.ts
```diff
@@ -16,7 +16,10 @@ export class LazilyTransformingAstService {
   public parser?: ParserWithCaching
   public undoRedo?: UndoRedo
 
+  private static readonly COMPACTION_THRESHOLD = 50
```
sequba: This value should be configurable. Unit tests should use lower value for `COMPACTION_THRESHOLD` to produce scenarios simpler and easier to analyze

src/LazilyTransformingAstService.ts
sequba, line 47: Please, add a jsdoc comment to `LazilyTransformingAstService`. The jsdocs should contain a dev documentations of the contraction feature - a complete, detailed description for the technical members of the HF team to get the idea of this mechanism:
- what it is?
- why was it introduced? what problem solve?
- how does it work

Also, do the similar thing for the optimization mechanizm that deals with `UndoRedo.oldData`

src/Lookup/SearchStrategy.ts
```diff
@@ -51,6 +51,13 @@ export interface ColumnSearchStrategy extends SearchStrategy {
   moveValues(range: IterableIterator<[RawScalarValue, SimpleCellAddress]>, toRight: number, toBottom: number, toSheet: number): void,
 
   removeValues(range: IterableIterator<[RawScalarValue, SimpleCellAddress]>): void,
+
+  /**
+   * Forces all lazily-tracked ValueIndex entries to apply any pending transformations,
+   * bringing every entry's version up to the current LTAS version.
```
sequba: Do not use `LTAS` shortcut in the code. It is not clear to people lacking the context of this issue.

src/HyperFormula.ts
```diff
@@ -4610,6 +4610,13 @@ export class HyperFormula implements TypedEmitter {
       const verticesToRecomputeFrom = this.dependencyGraph.verticesToRecompute()
       this.dependencyGraph.clearDirtyVertices()
 
+      if (this._lazilyTransformingAstService.needsCompaction()) {
+        this._dependencyGraph.forceApplyPostponedTransformations()
+        this._columnSearch.forceApplyPostponedTransformations()
+        this._lazilyTransformingAstService.compact()
+        this._lazilyTransformingAstService.undoRedo?.cleanupOrphanedOldData()
+      }
```
sequba: Please wrap this part of the code into a helper function with a name that explains what it does.

sequba (changes requested):
It seems that you came up with an elegant and efficient solution even though it was a complex issue. Good job! I need more dev-oriented in-code documentation so I and the future contributors understand the optimization mechanisms introduced with this PR.

Also, please add 2 changelog entries describing both fixes.

src/HyperFormula.ts
```diff
@@ -4595,6 +4595,20 @@ export class HyperFormula implements TypedEmitter {
...
 
+  /**
+   * When enough transformations have accumulated, forces all formula vertices and
+   * column index entries to apply pending lazy transformations, then compacts the
+   * transformation history and cleans up orphaned undo oldData entries.
+   */
+  private compactLazyTransformationsIfNeeded(): void {
+    if (this._lazilyTransformingAstService.needsCompaction()) {
+      this._dependencyGraph.forceApplyPostponedTransformations()
+      this._columnSearch.forceApplyPostponedTransformations()
```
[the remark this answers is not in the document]
marcin-kordas-hoc: **Note on duplicated force-apply calls** (`compactLazyTransformationsIfNeeded` vs `Operations.forceApplyPostponedTransformations`):

The two `forceApply` calls in `HyperFormula.compactLazyTransformationsIfNeeded()` intentionally duplicate the pair in `Operations.forceApplyPostponedTransformations()`. Delegating would create coupling — `Operations` does not and should not know about `compact()` or `cleanupOrphanedOldData()`. The two call sites serve different purposes:

- **Operations**: synchronizes state *before* undo move operations
- **HyperFormula**: synchronizes state *before* compaction + cleanup

src/ConfigParams.ts
```diff
@@ -402,6 +402,18 @@ export interface ConfigParams {
...
+   * of the LazilyTransformingAstService. When the number of pending transformations
+   * reaches this threshold, all formula vertices and column indexes are forced to
+   * apply their postponed transformations, and the transformation history is cleared.
+   *
+   * Lower values cause more frequent compaction (useful for testing), while higher
+   * values reduce overhead at the cost of more memory usage.
+   * @default 50
+   * @category Engine
+   */
+  compactionThreshold: number,
```
sequba: I don't like this name. Suggesting something more descriptive. E.g.:

```suggestion
  maxPendingLazyTransformations: number,
```

src/ConfigParams.ts
```diff
@@ -402,6 +402,18 @@ export interface ConfigParams {
...
    */
   undoLimit: number,
+  /**
+   * Sets the number of accumulated formula transformations that triggers compaction
+   * of the LazilyTransformingAstService. When the number of pending transformations
+   * reaches this threshold, all formula vertices and column indexes are forced to
+   * apply their postponed transformations, and the transformation history is cleared.
+   *
+   * Lower values cause more frequent compaction (useful for testing), while higher
+   * values reduce overhead at the cost of more memory usage.
```
sequba:
```suggestion
   * Controls memory usage for long-running instances by limiting the number of
   * pending lazy transformations before cleanup occurs.
   *
   * Structural operations (adding/removing rows/columns, moving cells) create
   * transformations that are applied lazily to formulas. This setting determines
   * how many can accumulate before they are flushed and memory is reclaimed.
   *
   * Lower values reduce peak memory usage but may slightly increase CPU overhead.
   * Higher values reduce overhead but allow more memory accumulation.
   * 
```

This description is a little less enigmatic IMHO

CHANGELOG.md
```diff
@@ -9,10 +9,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Added
 
+- Added `compactionThreshold` configuration option to control how often the engine compacts accumulated formula transformations. [#1629](https://github.com/handsontable/hyperformula/issues/1629)
```
```suggestion
- Added `maxPendingLazyTransformations` configuration option to control memory usage by limiting accumulated transformations before cleanup. [#1629](https://github.com/handsontable/hyperformula/issues/1629)
```

src/ConfigParams.ts
sequba: Also, this option should be described in the performance guide in the docs

sequba (changes requested):
Generally great fix. Just a few naming changes and documentation improvements

sequba (approved):
Super!

PR #1661, opened 2026-04-27
HF-24: add stringifyCurrency config callback for TEXT

marcin-kordas-hoc:
## Summary

Adds a `stringifyCurrency` config option mirroring the existing `stringifyDateTime` / `stringifyDuration` callbacks. When set, the `TEXT` function consults the callback before falling through to the built-in number formatter, so users can plug in locale-aware currency formatting (for example via `Intl.NumberFormat` or a third-party library) without bringing currency data into the HyperFormula core.

The default implementation returns `undefined` so existing TEXT behavior is preserved bit-for-bit.

## Linked

- Spec: `agents/hyperformula/docs/specs/2026-04-21-hf-24-currency-in-text.md`
- Tech rationale: `agents/hyperformula/docs/specs/2026-04-24-hf-24-tech-rationale.md`
- Implementation plan: `agents/hyperformula/docs/specs/2026-04-27-hf-24-stringify-currency-plan.md`

## Tests

Tests added in the matching `feature/hf-24-stringify-currency` branch of `handsontable/hyperformula-tests`. Coverage:

- default callback returns `undefined`
- custom callback intercepts currency formats
- callback opts out (returns `undefined`) → fall-through to `numberFormat`
- date / duration formats are not intercepted by `stringifyCurrency`
- six representative Excel format strings handled by the docs `Intl.NumberFormat` adapter (USD, EUR via LCID, JPY via LCID, PLN via LCID, accounting two-section, fall-through)

## Notes

- **PLN format string change**: the spec example originally used `#,##0.00 "zł"` (trailing quoted symbol). HF's formula parser does not accept embedded quotes inside TEXT format strings, so the docs example and the corresponding test were swapped to use `[$zł-415] #,##0.00` (LCID-tagged symbol). The adapter still recognizes the trailing-quote pattern for users invoking the callback outside HyperFormula.
- **Pre-existing lint errors** (9, all in `test/hyperformula-tests/unit/memory-management.spec.ts`) are unrelated to this PR — they originate from PR #4 of the private repo and should be fixed in a separate ticket.

## Test plan

- [ ] CI green on `handsontable/hyperformula` PR
- [ ] CI green on `handsontable/hyperformula-tests` PR (matching branch)
- [ ] Manual: build docs locally, render the new "Currency integration" section under Date and time handling, copy-paste the adapter into a fresh sandbox to confirm it runs against `Intl.NumberFormat` on Node 14+

---

> [!NOTE]
> **Medium Risk**
[...]

PR #1665, opened 2026-04-29
HF-24: add stringifyCurrency config callback for TEXT

marcin-kordas-hoc:
## Summary

Adds a `stringifyCurrency` config option mirroring the existing `stringifyDateTime` / `stringifyDuration` callbacks. When set, the `TEXT` function consults the callback before falling through to the built-in number formatter, so users can plug in locale-aware currency formatting (for example via `Intl.NumberFormat` or a third-party library) without bringing currency data into the HyperFormula core.

The default implementation returns `undefined` so existing TEXT behavior is preserved bit-for-bit **for non-LCID-tagged formats**. LCID-tagged currency strings (`[$SYMBOL-LCID]`) now correctly skip the date/time parser (see **Added** in CHANGELOG); that is a separate behavioral fix, not a regression.

> **Note on PR routing**: this PR replaces #1661 which was opened from a fork branch (`marcin-kordas-hoc`). Fork-side PRs cannot access the `DEPLOY_TOKEN` secret needed to clone the private `hyperformula-tests` repo, so 3 of the matrix checks (`Test performance`, `unit-tests`, `browser-tests`) failed structurally there. Same content, same SHA (`0246ce0bcbbed09ac9bf5e24af38b0d8aa1ac099`), now from upstream branch — CI will have full access. Closing #1661 in favor of this one.

## Linked

- Spec: `agents/hyperformula/docs/specs/2026-04-21-hf-24-currency-in-text.md`
- Tech rationale: `agents/hyperformula/docs/specs/2026-04-24-hf-24-tech-rationale.md`
- Implementation plan: `agents/hyperformula/docs/specs/2026-04-27-hf-24-stringify-currency-plan.md`
- Supersedes: #1661

## Tests

Tests added in the matching `feature/hf-24-stringify-currency` branch of `handsontable/hyperformula-tests`. Coverage:

- default callback returns `undefined`
- custom callback intercepts currency formats
- callback opts out (returns `undefined`) → fall-through to `numberFormat`
- date / duration formats are not intercepted by `stringifyCurrency`
- five Excel format strings actively handled by the docs `Intl.NumberFormat` adapter (USD shorthand, EUR via LCID, JPY via LCID, PLN via LCID, accounting two-section), plus a fall-through case demonstrating opt-out

## Notes

- **PLN format string change**: the spec example originally used `#,##0.00 "zł"` (trailing quoted symbol). HF's formula parser does not accept embedded quotes inside TEXT format strings, so the docs example and the corresponding test were swapped to use `[$zł-415] #,##0.00` (LCID-tagged symbol). The adapter still recognizes the trailing-quote pattern for users invoking the callback outside HyperFormula.
- **EUR / JPY assertion shapes reflect ICU output, not the original plan**: tests assert `'1.234,50 €'` (symbol-trailing) for `[$€-2]` and `'￥1,235'` (full-width yen sign, no space) for `[$¥-411]` because that is what `Intl.NumberFormat('de-DE'/'ja-JP', ...)` actually produces on modern Node ICU. NBSP normalization in tests covers both `\u00A0` and `\u202F` variants for ICU build robustness.
- **No-LCID `[$SYMBOL]` boundary**: the example regex requires the `-LCID` segment. A bare `[$USD]` pattern is not handled by the adapter and falls through to the built-in `numberFormat`, whose handling of `[$...]` in HyperFormula is implementation-defined. Test `docs adapter does not handle [$SYMBOL] without LCID segment` documents the boundary.
- **Additional minor changes** (not in summary): (a) two broken anchor links fixed in `docs/guide/built-in-functions.md`; (b) `docs/guide/i18n-features.md` replaces a stale `currencySymbol` code example with a cross-reference; (c) `tsconfig.test.json` adds `test-utils` to `include`; (d) `docs/guide/known-limitations.md` — new bullet for the TEXT embedded-quote limitation (can't use `""` escape in format strings; use LCID-tagged form or `stringifyCurrency` callback); (e) `docs/guide/date-and-time-handling.md` — adds cross-reference paragraph to `currency-handling.md`; (f) `docs/guide/list-of-differences.md` — updates the "No currency formatting in TEXT" row with callback instructions; (g) `.eslintignore` — adds `test-utils/snippets` exclusion; (h) `test-utils/snippets/*.generated.ts` — generated from the docs and **not committed**; regenerated before every test run by inlining `npm run snippets:extract &&` into the `test:jest`/`test:ci`/`test:browser` scripts (`npm-run-all` does not fire `pre*` hooks) and git-ignored, so the docs stay the single source of truth. (i) `script/extract-doc-snippets.js` (257 LOC) — new codegen script that walks the docs for `<!-- snippet:NAME -->` markers and syncs them into `test-utils/snippets/*.generated.ts`. Heavy for one snippet but designed for future docs expansion; see inline docstring for design rationale.
- **Performance benchmark shows +1.9% to +8.7% across metrics** vs the post-merge base (`456adddff` = develop with HF-85 DatabasePlugin). HF-24's runtime impact is one extra dispatcher call in `format()` per `TEXT` invocation, which the Sheet A/B/T benchmarks don't exercise. The variance is most likely benchmark noise or HF-85 import overhead carried in via the develop merge, not HF-24-specific.
- **`Maybe<T>` in callback types**: `stringifyCurrency`, `stringifyDateTime`, and `stringifyDuration` all declare return type as `Maybe<string>` (= `string | undefined`). This is consistent public API surface shared by all three siblings; the alias is re-exported from HyperFormula's type definitions. Changing only `stringifyCurrency` to bare `string | undefined` would create asymmetry. A follow-up can align all three if desired.

## Test plan

- [x] CI green on `handsontable/hyperformula` PR
- [x] CI green on `handsontable/hyperformula-tests` PR (matching branch)
- [x] Manual: built docs locally — `vuepress build docs` returns EXIT 0 with 215 sitemap entries; new `currency-handling.md` guide renders with the `Currency input` + `Currency output` sections (sidebar wired under Internationalization) per Kuba's review feedback

---
**Private tests PR:** handsontable/hyperformula-tests#10

---

> [!NOTE]
> **Medium Risk**
> Every `TEXT` invocation goes through an extra formatter dispatch step, and LCID-tagged currency format strings change behavior (intentional fix); misconfigured callbacks can alter formula output broadly.
> 
> **Overview**
> Introduces **`stringifyCurrency`**, a config callback parallel to `stringifyDateTime` / `stringifyDuration`, so `TEXT` can delegate currency output to custom formatters (e.g. `Intl.NumberFormat`) while the default no-op preserves existing behavior for ordinary formats.
> 
> **Runtime:** `format()` now calls `stringifyCurrency` first. `defaultStringifyDateTime` and `defaultStringifyDuration` skip Excel **`[$SYMBOL-LCID]`** currency tags so those strings are no longer mangled by date/time parsing and can reach the number formatter or a user callback. Config wiring lives in `Config.ts`, `ConfigParams.ts`, and `defaultStringifyCurrency` in `format.ts`.
> 
> **Docs & DX:** New **Currency handling** guide, sidebar entry, compatibility/differences updates, changelog entries, and expanded `DEV_DOCS` testing guidance. **`script/extract-doc-snippets.js`** generates `test-utils/snippets/*.generated.ts` from `<!-- snippet:NAME -->` blocks in docs; `test:jest` / `test:ci` / `test:browser` run `snippets:extract` first; generated files are gitignored.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9cc8d6ceb5f2462b160064113aa504e1990a6203. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [HF-24 Support popular currency formats in TEXT function](https://app.clickup.com/t/9015210959/HF-24)

src/format/format.ts
```diff
@@ -143,6 +153,20 @@ export function defaultStringifyDuration(time: SimpleTime, formatArg: string): M
 }
 
 export function defaultStringifyDateTime(dateTime: SimpleDateTime, formatArg: string): Maybe<string> {
+  // Skip date/time interpretation for Excel currency formats tagged with
```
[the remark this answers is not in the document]
marcin-kordas-hoc: Fixed in commit `80fd34e3e` — applied the same `/\[\$[^\-\]]+-/` LCID-tagged guard to `defaultStringifyDuration` so currency symbols with duration-token letters (H in CHF/HUF, M in AMD/HMD) are no longer interpreted as time tokens when a partial `stringifyCurrency` callback returns undefined. Guard sits in the same position relative to `parseForDateTimeFormat` for sibling parity with `defaultStringifyDateTime`.

src/format/format.ts
```diff
@@ -143,6 +161,20 @@ export function defaultStringifyDuration(time: SimpleTime, formatArg: string): M
...
+  // the output when a user-supplied stringifyCurrency callback opts out by
+  // returning undefined.
+  //
+  // The guard intentionally requires at least one character between `[$`
+  // and the `-` to distinguish currency tags (`[$USD-409]`, `[$€-2]`) from
+  // Excel's locale-only modifier (`[$-409]`, `[$-F800]`), which is valid
+  // on date/time formats and must continue to flow through this function.
+  if (/\[\$[^\-\]]+-/.test(formatArg)) {
+    return undefined
+  }
```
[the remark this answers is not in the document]
marcin-kordas-hoc: Fixed in commit `19ce06a63` — comment reworded to acknowledge that for LCID-tagged currency formats (`[$SYMBOL-LCID] ...`) output does change post-HF-24: pre-fix the date parser would mangle the symbol (e.g. USD → US9), post-fix the value falls through to numberFormat. The 'bit-for-bit' claim is now correctly scoped to non-currency formats (dates, durations, $#,##0.00 etc.) where it does hold.

qunabu:
Task linked: [HF-85 Implement function DCOUNT](https://app.clickup.com/t/9015210959/HF-85)

docs/guide/known-limitations.md
```diff
@@ -38,3 +38,4 @@ you can't compare the arguments in a formula like this:
 * The INDEX function doesn't support returning whole rows or columns of the source range – it always returns the contents of a single cell.
 * The FILTER function accepts either single rows of equal width or single columns of equal height. In other words, all arrays passed to the FILTER function must have equal dimensions, and at least one of those dimensions must be 1.
 * Array-producing functions (e.g., SEQUENCE, FILTER) require their output dimensions to be determinable at parse time. Passing cell references or formulas as dimension arguments (e.g., `=SEQUENCE(A1)`) results in a `#VALUE!` error, because the output size cannot be resolved before evaluation.
+* The TEXT function does not accept embedded double-quote literals in the format string (e.g., `=TEXT(A1, "#,##0.00 ""zł""")` fails at parse time). Use the LCID-tagged form (`[$zł-415] #,##0.00`) or supply a custom [`stringifyCurrency`](configuration-options.md#stringifycurrency) callback that handles such formats outside the parser.
```
sequba: What is the LCID-tag? If I want to display polish currencies in a format "1234,56 zł", can I do it without `stringifyCurrency`?

docs/guide/list-of-differences.md
```diff
@@ -28,7 +28,7 @@ See a full list of differences between HyperFormula, Microsoft Excel, and Google
...
+| Formatting inside the TEXT function                | TEXT(A1,"dd-mm-yy")<br><br>TEXT(A1,"###.###”)                             | Not all formatting options are supported,<br>e.g., only some date formatting options: (`hh`, `mm`, `ss`, `am`, `pm`, `a`, `p`, `dd`, `yy`, and `yyyy`).<br><br>Currency formatting is opt-in via the [`stringifyCurrency`](date-and-time-handling.md#currency-integration) callback; without it, currency format strings fall through to the built-in number formatter.<br><br>Embedded double-quote literals (e.g. `#,##0.00 "zł"`) are not accepted by the parser; use the LCID-tagged form (`[$zł-415] #,##0.00`) instead. | A wide variety of options for string formatting is supported.                                                                  | Same as Google Sheets.                                                                                                         |
```
sequba: Since we added `stringifyCurrency`, I think we can remove this line from List of Runtime Differences. Instead in the Compatibility with Ms Excel and Compatibility with Google Sheets guides, we should mention that the user need to provide both `stringifyDateTime` and `stringifyCurrency` to support all formats in the TEXT function.

docs/guide/date-and-time-handling.md
```diff
@@ -96,6 +96,172 @@ const data = [["31st Jan 00", "2nd Jun 01", "=B1-A1"]];
 
 And now, HyperFormula recognizes these values as valid dates and can operate on them.
 
+## Currency integration
```
sequba: This section should be made into a separate guide `currency-handling`

docs/guide/date-and-time-handling.md
```diff
@@ -96,6 +96,172 @@ const data = [["31st Jan 00", "2nd Jun 01", "=B1-A1"]];
...
+
+By default, the `TEXT` function renders only the simplest currency-looking formats — `"$0.00"`, `"$0"`, or `"$#.00"` (no thousands separator). Common Excel patterns such as `"$#,##0.00"` (with comma grouping), `"[$€-2] #,##0.00"` (EUR with German grouping), `"[$zł-415] #,##0.00"` (PLN), or accounting two-section formats like `"$#,##0.00;($#,##0.00)"` are **not** rendered correctly by the built-in number formatter; provide a [`stringifyCurrency`](../api/interfaces/configparams.md#stringifycurrency) callback to handle them.
+
+HyperFormula itself ships with **no currency data** and **no currency library dependency**. You choose how to format: native `Intl.NumberFormat`, a third-party library, or a hand-rolled lookup table.
```
sequba: IMHO these paragraph sound to negative. They say a lot about things that HyperFormula does not support. I'd rather say it in more positive tone like `Out of the box HF supports all currency symbols through the `currencySymbol` config option and simple currency formats e.g.: "$0.00", "$0", or "$#.00". If your app needs more formats, you can define them using the `stringifyCurrency` configuration option.` (do not use my exact wording; it's just an example of the more positive-sounding tone).

docs/guide/date-and-time-handling.md
```diff
@@ -96,6 +96,172 @@ const data = [["31st Jan 00", "2nd Jun 01", "=B1-A1"]];
...
+
+The callback contract:
```
sequba: Before discussing the callback, give a very simple example of the behavior with just `currencySymbol` provided and no custom `stringifyCurrency`.

sequba (changes requested):

docs/guide/known-limitations.md
marcin-kordas-hoc: Good catch — clarified inline in [`known-limitations.md`](https://github.com/handsontable/hyperformula/blob/feature/hf-24-stringify-currency/docs/guide/known-limitations.md): defined the LCID tag as a Microsoft Locale ID with a link to the MS-LCID spec, and explicitly noted that the Polish `"1234,56 zł"` pattern (decimal-comma) requires the `stringifyCurrency` callback because the built-in number formatter always emits `.` as the decimal separator. Fixed in f1eb4efb2.

docs/guide/list-of-differences.md
marcin-kordas-hoc: Done. The TEXT-formatting row in [`list-of-differences.md`](https://github.com/handsontable/hyperformula/blob/feature/hf-24-stringify-currency/docs/guide/list-of-differences.md) is now a one-liner that points at both `stringifyDateTime` and `stringifyCurrency`. Added a new "TEXT function formats" subsection to both [`compatibility-with-microsoft-excel.md`](https://github.com/handsontable/hyperformula/blob/feature/hf-24-stringify-currency/docs/guide/compatibility-with-microsoft-excel.md) and [`compatibility-with-google-sheets.md`](https://github.com/handsontable/hyperformula/blob/feature/hf-24-stringify-currency/docs/guide/compatibility-with-google-sheets.md) noting that both callbacks together cover the full TEXT format range. Fixed in f1eb4efb2.

docs/guide/date-and-time-handling.md
marcin-kordas-hoc: Done. The whole "Currency integration" section moved into a new top-level guide [`docs/guide/currency-handling.md`](https://github.com/handsontable/hyperformula/blob/feature/hf-24-stringify-currency/docs/guide/currency-handling.md), wired into the Internationalization sidebar section. `date-and-time-handling.md` keeps a single-paragraph pointer to the new guide. Fixed in f1eb4efb2.

docs/guide/date-and-time-handling.md
marcin-kordas-hoc: Reframed. The new [`currency-handling.md`](https://github.com/handsontable/hyperformula/blob/feature/hf-24-stringify-currency/docs/guide/currency-handling.md) opens by stating what HF handles out of the box (simple `$`-prefixed formats — `"$0.00"`, `"$0"`, `"$#.00"`) and frames `stringifyCurrency` as the additive extension point for locale-aware grouping, non-`$` symbols, decimal-comma patterns, and accounting two-section formats. The reference table is preserved further down for users who want the side-by-side comparison, but no longer leads the page. Fixed in f1eb4efb2.

docs/guide/date-and-time-handling.md
marcin-kordas-hoc: Done. The new [`currency-handling.md`](https://github.com/handsontable/hyperformula/blob/feature/hf-24-stringify-currency/docs/guide/currency-handling.md) now opens with a "Default behavior" subsection that demonstrates simple `$`-prefixed formats (`"$0.00"`, `"$#.00"`) producing `"$1234.50"` out of the box, *before* introducing the callback contract. The `stringifyCurrency` discussion follows for users who need richer formats. Fixed in f1eb4efb2.

docs/guide/date-and-time-handling.md
marcin-kordas-hoc: Refined per your follow-up on Slack: `currency-handling.md` is now the single authority for currency (both input parsing via `currencySymbol` and output formatting via `stringifyCurrency`), with explicit input-vs-output framing at the top. `i18n-features.md` drops its standalone Currency symbol subsection and links here instead — no duplication. Done in e1756701c.

src/format/format.ts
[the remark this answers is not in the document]
marcin-kordas-hoc: Confirming on the new HEAD — the JSDoc enrichment in `d11cfd9ca` codifies the LCID-guard design intent at-source on both `defaultStringifyDateTime` and `defaultStringifyDuration`, so the rationale (currency tags like `[$AMD-…]`, `[$CHF-…]`, `[$HUF-…]` deliberately short-circuit to `undefined` rather than getting consumed by duration tokens) lives at the function definition rather than only in the original commit message. Verified at [src/format/format.ts:112-150](https://github.com/handsontable/hyperformula/blob/de4d6fea6/src/format/format.ts#L112-L150).

src/format/format.ts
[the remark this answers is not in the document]
marcin-kordas-hoc: Adding to the May 11 thread — the comment fix in `19ce06a63` reworded the dispatcher rationale (dropped the bit-for-bit-compatibility claim for LCID-tagged currency formats). `d11cfd9ca` extends that with full JSDoc on `defaultStringifyDateTime` and `defaultStringifyDuration`, so the design intent (LCID currency tags return `undefined` so the value falls through to a user-supplied `stringifyCurrency` callback or `parseForNumberFormat`) is documented at the function definition, not only inline at the dispatcher call site. See [src/format/format.ts:197-256](https://github.com/handsontable/hyperformula/blob/de4d6fea6/src/format/format.ts#L197-L256).

docs/guide/list-of-differences.md
sequba: I'd rather write something like "To support all date, time and currency formats, set `stringifyDateTime` and `stringifyCurrency` configuration options."

docs/guide/currency-handling.md
```diff
@@ -0,0 +1,235 @@
...
+console.log(hf.getCellValue({ sheet: 0, col: 1, row: 2 })) // "($1,234.50)"
+```
+
+::: tip
+The output values above contain non-breaking spaces (U+00A0 or U+202F depending on locale and ICU/CLDR version) as locale-appropriate separators. The comments show them as regular spaces for readability. When comparing programmatically, normalize with `.replace(/[  ]/g, ' ')` if you need ASCII-space output.
+:::
+
+#### What is an LCID tag?
+
+Excel can mark a currency format with a [Microsoft Locale Identifier](https://learn.microsoft.com/openspecs/windows_protocols/ms-lcid) (LCID) so the symbol carries locale context. The syntax is `[$SYMBOL-LCID]` followed by the number template — for example `[$zł-415] #,##0.00` means *"Polish złoty, hex LCID `415` = `pl-PL`"*, and `[$€-2] #,##0.00` means *"euro, generic"*. The adapter above parses the LCID to pick the matching `Intl.NumberFormat` locale and ISO 4217 currency code.
```
sequba: How does Excel handle LCID tags?

package.json
```diff
@@ -60,6 +60,8 @@
     "docs:code-examples:generate-js": "bash docs/code-examples-generator.sh",
     "docs:code-examples:generate-all-js": "bash docs/code-examples-generator.sh --generateAll",
     "docs:code-examples:format-all-ts": "bash docs/code-examples-generator.sh --formatAllTsExamples",
+    "snippets:extract": "node script/extract-doc-snippets.js",
+    "snippets:check": "node script/extract-doc-snippets.js && git diff --exit-code -- test-utils/snippets/",
```
sequba: What is the purpuse of these scripts?

docs/guide/known-limitations.md
```diff
@@ -38,6 +38,7 @@ you can't compare the arguments in a formula like this:
...
 * The FILTER function accepts either single rows of equal width or single columns of equal height. In other words, all arrays passed to the FILTER function must have equal dimensions, and at least one of those dimensions must be 1.
 * Array-producing functions (e.g., SEQUENCE, FILTER) require their output dimensions to be determinable at parse time. Passing cell references or formulas as dimension arguments (e.g., `=SEQUENCE(A1)`) results in a `#VALUE!` error, because the output size cannot be resolved before evaluation.
+* The TEXT function does not accept embedded double-quote literals in the format string (e.g., `=TEXT(A1, "#,##0.00 ""zł""")` fails at parse time). Use Excel's LCID-tagged form — `[$SYMBOL-LCID]` where LCID is a hex [Microsoft Locale ID](https://learn.microsoft.com/openspecs/windows_protocols/ms-lcid), e.g. `[$zł-415] #,##0.00` for Polish złoty — or supply a custom [`stringifyCurrency`](currency-handling.md) callback that handles such formats outside the parser. For locale-specific patterns like the Polish `"1234,50 zł"` (decimal comma), the callback is required because the built-in number formatter always emits `.` as the decimal separator.
```
sequba: How does Excel handle formats like `=TEXT(A1, "#,##0.00 ""zł""")`?

docs/guide/currency-handling.md
```diff
@@ -0,0 +1,235 @@
+# Currency handling
```
sequba, line 1: Nice, solid guide explaining everything very well 👍

docs/guide/list-of-differences.md
marcin-kordas-hoc: Done — updated to your suggested wording: "To support all date, time and currency formats, set `stringifyDateTime` and `stringifyCurrency` configuration options." Implemented in [8cbb139d4](https://github.com/handsontable/hyperformula/commit/8cbb139d4).

docs/guide/currency-handling.md
marcin-kordas-hoc: Excel resolves LCID tags natively — no extra configuration is required. For example, `[$€-2] #,##0.00` automatically uses European grouping and decimal separators and produces `1.234,50 €`; `[$zł-415] #,##0.00` uses `pl-PL` and produces `1 234,50 zł`. HyperFormula's built-in formatter does not resolve LCID tags; the adapter replicates that behavior via `Intl.NumberFormat`.

Added a clarifying paragraph to the "What is an LCID tag?" section in [8cbb139d4](https://github.com/handsontable/hyperformula/commit/8cbb139d4).

docs/guide/known-limitations.md
marcin-kordas-hoc: In Excel, `""` inside a format string is an escape sequence for a literal `"` character — so `#,##0.00 ""zł""` is equivalent to `#,##0.00 "zł"` (a number pattern with `zł` as a quoted literal suffix) and `=TEXT(1234.5, "#,##0.00 ""zł""")` returns `"1,234.50 zł"`. HyperFormula's parser does not support this escape sequence and fails at parse time.

Updated the limitation entry to document what Excel produces before stating the HF restriction, in [8cbb139d4](https://github.com/handsontable/hyperformula/commit/8cbb139d4).

package.json
marcin-kordas-hoc: These two scripts maintain sync between the docs code examples and the generated test utilities in `test-utils/snippets/`:

- `snippets:extract` — scans markdown docs for `<!-- snippet:NAME --> … <!-- /snippet:NAME -->` fences and writes each extracted block to `test-utils/snippets/<NAME>.generated.ts`, so the same code that appears in the docs can be imported directly in tests.
- `snippets:check` — runs the same extraction, then asserts via `git diff --exit-code` that the generated files match what is committed. Used in CI to ensure docs snippets and generated test utilities never drift apart.

docs/guide/known-limitations.md
```diff
@@ -38,6 +38,7 @@ you can't compare the arguments in a formula like this:
...
+* The TEXT function does not accept embedded double-quote literals in the format string. In Excel, `""` inside a format string is an escape sequence for a literal `"` character — so `#,##0.00 ""zł""` is equivalent to `#,##0.00 "zł"` (a number pattern with `zł` as a quoted literal suffix) and `=TEXT(1234.5, "#,##0.00 ""zł""")` returns `"1,234.50 zł"`. HyperFormula's parser does not support this escape sequence and fails at parse time. Use Excel's LCID-tagged form — `[$SYMBOL-LCID]` where LCID is a hex [Microsoft Locale ID](https://learn.microsoft.com/openspecs/windows_protocols/ms-lcid), e.g. `[$zł-415] #,##0.00` for Polish złoty — or supply a custom [`stringifyCurrency`](currency-handling.md) callback that handles such formats outside the parser. For locale-specific patterns like the Polish `"1234,50 zł"` (decimal comma), the callback is required because the built-in number formatter always emits `.` as the decimal separator.
```
sequba:
```suggestion
* The TEXT function does not accept embedded double-quote literals in the format string. In Excel, `""` inside a format string is an escape sequence for a literal `"` character — e.g. `=TEXT(1234.5, "#,##0.00 ""zł""")` returns `"1,234.50 zł"`. If your application requires this escape sequence, supply a custom [`stringifyCurrency`](currency-handling.md) callback.
```

I think this is sufficient

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Added
 
 - Added an Indonesian (Bahasa Indonesia) language pack. [#1674](https://github.com/handsontable/hyperformula/pull/1674)
+- Added a `stringifyCurrency` config option that lets you plug in a custom currency formatter for the `TEXT` function. [#1145](https://github.com/handsontable/hyperformula/issues/1145)
+
+### Fixed
+
+- Fixed the `TEXT` function mangling LCID-tagged currency format strings (e.g. `[$USD-409] #,##0.00`) — pre-fix, the date-time parser greedily consumed letter tokens inside the currency code (`D` in USD, `H` in CHF/HUF, etc.), producing strings like `[$US9-409]`. The default dispatch now short-circuits LCID-tagged currency formats so they fall through to the number formatter (or the user-supplied `stringifyCurrency` callback). Applies to every LCID-tagged currency format regardless of whether `stringifyCurrency` is configured. [#1665](https://github.com/handsontable/hyperformula/pull/1665)
```
sequba: I would rather put this in the `Added` section as it is a new feature rather than a fix for something that was broken.

.github/workflows/lint.yml
```diff
@@ -46,3 +46,8 @@ jobs:
 
       - name: Run linter
         run: npm run lint
+
+      - name: Check docs snippets are in sync with extracted source-of-truth
```
sequba: This does not belong in `lint` workflow. The test utils should be generated by unit-test and browser-test workflows before running the actual tests

test-utils/snippets/currency-adapter.generated.ts
```diff
@@ -0,0 +1,71 @@
+// Auto-generated by script/extract-doc-snippets.js — DO NOT EDIT.
```
sequba: The `.generated.ts` files should not be committed. Add it to `.gitignore`.

package.json
sequba: Instead of checking it with `git diff`, we can generate these snippet files every time before running tests and not keep them in the repository (it's unnecessary duplication of code)

docs/guide/currency-handling.md
```diff
@@ -0,0 +1,237 @@
...
+```
+
+A non-`$` symbol used purely as a suffix (no thousands grouping, no decimal-comma) also passes through unchanged:
+
+```javascript
+const hf = HyperFormula.buildFromArray([[1234.5, '=TEXT(A1, "0.00 zł")']]);
+console.log(hf.getCellValue({ sheet: 0, col: 1, row: 0 })); // "1234.50 zł"
+```
+
+Configure `stringifyCurrency` when your formula corpus needs:
```
sequba:
```suggestion
Configure `stringifyCurrency` when your formula corpus needs more advanced currency formats. E.g.:
```

sequba (changes requested):
Ok, now that I understand the purpose of `snippets:extract` and `snippets:check`, I can finish the review. There are a few more cosmetic things to fix, but we're almost there.

docs/guide/known-limitations.md
marcin-kordas-hoc: Applied in cd024e7a0; the bullet now uses the shorter wording.

docs/guide/currency-handling.md
marcin-kordas-hoc: Applied in cd024e7a0; the lead-in now reads "needs more advanced currency formats. E.g.:".

CHANGELOG.md
marcin-kordas-hoc: Done in cd024e7a0. Moved the entry to `Added` and reframed it as a new capability — LCID-tagged currency formats now resolve correctly regardless of whether `stringifyCurrency` is configured. The old `Fixed`-section wording is gone.

.github/workflows/lint.yml
marcin-kordas-hoc: Done in cd024e7a0. Removed the `snippets:check` step from the lint workflow. Generation now runs via the `pretest:ci` / `pretest:jest` / `pretest:browser` hooks — produced by the test scripts themselves, in both CI test jobs and locally, before jest/karma. I also added a comment at each test job in `test.yml` documenting the pre-generation, so it is discoverable in the workflow without duplicating the step.

test-utils/snippets/currency-adapter.generated.ts
marcin-kordas-hoc: Done in cd024e7a0. Untracked the file and added `test-utils/snippets/*.generated.ts` to `.gitignore`; it is regenerated before every test run.

package.json
marcin-kordas-hoc: Done in cd024e7a0. Dropped `snippets:check` and the `git diff` gate; the files are no longer committed — the `pretest:*` hooks run `snippets:extract` before every test run, so the docs stay the single source of truth with no duplicated snippet code. The consumer spec already skips its currency block gracefully if the artifact is absent.

package.json
```diff
@@ -77,14 +78,17 @@
     "verify:typings": "tsc --noEmit",
     "test": "npm-run-all lint test:jest test:browser",
     "test:setup-private": "bash test/fetch-tests.sh",
+    "pretest:jest": "npm run snippets:extract",
     "test:jest": "cross-env NODE_ICU_DATA=node_modules/full-icu jest",
```
sequba:
```suggestion
    "test:jest": "npm run snippets:extract && cross-env NODE_ICU_DATA=node_modules/full-icu jest",
```

package.json
```diff
@@ -77,14 +78,17 @@
...
     "test:watch": "npm run test:jest -- --watch",
     "test:tmp": "npm run test:jest -- --watch function-irr",
     "test:coverage": "npm run test:jest -- --coverage",
     "test:logMemory": "npm run test:jest -- --runInBand --logHeapUsage",
     "test:performance": "npm run benchmark:basic && npm run benchmark:cruds",
     "test:compatibility": "bash test/compatibility/test-compatibility.sh",
+    "pretest:ci": "npm run snippets:extract",
     "test:ci": "cross-env NODE_ICU_DATA=node_modules/full-icu node --expose-gc ./node_modules/jest/bin/jest --forceExit",
```
sequba:
```suggestion
    "test:ci": "npm run snippets:extract && cross-env NODE_ICU_DATA=node_modules/full-icu node --expose-gc ./node_modules/jest/bin/jest --forceExit",
```

package.json
```diff
@@ -77,14 +78,17 @@
...
+    "pretest:browser": "npm run snippets:extract",
     "test:browser": "cross-env-shell BABEL_ENV=dist env-cmd -f ht.config.js karma start",
```
sequba:
```suggestion
    "test:browser": "npm run snippets:extract && cross-env-shell BABEL_ENV=dist env-cmd -f ht.config.js karma start",
```

.github/workflows/test.yml
```diff
@@ -44,6 +44,8 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      # Doc snippets are regenerated before the tests by the `pretest:ci` npm
+      # hook (runs `snippets:extract`); the generated files are not committed.
```
sequba: But `pretest:ci` is never run. I suggested the change to the npm scripts to run it before actual tests

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Added
 
 - Added an Indonesian (Bahasa Indonesia) language pack. [#1674](https://github.com/handsontable/hyperformula/pull/1674)
+- Added a `stringifyCurrency` config option that lets you plug in a custom currency formatter for the `TEXT` function. [#1145](https://github.com/handsontable/hyperformula/issues/1145)
+- Added support for LCID-tagged currency format strings (e.g. `[$USD-409] #,##0.00`) in the `TEXT` function. Previously the date-time parser greedily consumed letter tokens inside the currency code (`D` in USD, `H` in CHF/HUF, etc.), producing corrupted output like `[$US9-409]`; such formats now fall through to the number formatter (or a user-supplied `stringifyCurrency` callback), regardless of whether `stringifyCurrency` is configured. [#1665](https://github.com/handsontable/hyperformula/pull/1665)
```
```suggestion
- Added support for LCID-tagged currency format strings (e.g. `[$USD-409] #,##0.00`) in the `TEXT` function. [#1665](https://github.com/handsontable/hyperformula/pull/1665)
```

sequba (changes requested):
Good. The only thing is that the `pretest:*` scripts are never run. Look at my suggestions.

.github/workflows/test.yml
marcin-kordas-hoc: Good catch — the root cause was that the `pretest:*` hooks never fired through `npm-run-all`. Fixed by inlining `npm run snippets:extract &&` directly into `test:jest` / `test:ci` / `test:browser` and dropping the `pretest:*` hooks, so the generated snippets are produced before every test run, in CI and locally. CHANGELOG trimmed to the one-liner you suggested.

marcin-kordas-hoc:
Addressed the CHANGES_REQUESTED round — where things stand on the current HEAD:

- **CHANGELOG** → moved to `Added`. Two entries: the new `stringifyCurrency` config option, and the LCID-tag dispatch change (which alters `[$…-409]` output even with no callback). Happy to collapse to one line if you prefer.
- **Test-utils generation** → removed from the `lint` workflow; `snippets:extract` now runs inline at the start of `test:jest` / `test:ci` / `test:browser` (the `pretest:*` hooks were never firing), so the adapter is generated before the unit/browser tests run.
- **`.generated.ts`** → no longer committed; gitignored (`/test-utils/snippets/*.generated.ts`) and eslintignored.
- **Docs** → currency split into its own `currency-handling` guide; `date-and-time-handling`, `i18n-features`, `list-of-differences` and the Excel/Sheets compatibility pages point at it; the guide opens with the out-of-the-box `currencySymbol` example before introducing the callback.
- **known-limitations** → embedded double-quote-literal limitation + the LCID-tag explanation added per your suggestion.

One behavior note: a missing generated adapter now **fails** the consumer spec (it throws) rather than skipping, so the artifact can't silently disappear — CI always runs `snippets:extract` first.

Known follow-up (out of scope for this round): the LCID guard lives in the default `stringifyDateTime`/`stringifyDuration` callbacks, so a user who fully replaces one of those would lose it for currency-letter masks — can hoist it if you think it's worth doing here. Ready for another look.

.github/workflows/test.yml
```diff
@@ -44,6 +44,8 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      # `test:ci` regenerates the doc snippets (runs `snippets:extract`) before
+      # running the tests; the generated files are not committed.
```
sequba: These comments are useless and will likely get outdated, because noone will remember to update them when they modify `npm run test:ci` script

.github/workflows/test.yml
```diff
@@ -82,5 +84,7 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      # `test:browser` regenerates the doc snippets (runs `snippets:extract`)
+      # before running the tests; the generated files are not committed.
```
sequba: Same thing about this comment

.github/workflows/test.yml
marcin-kordas-hoc: Done in 5b9baf68a — removed the comment.

.github/workflows/test.yml
marcin-kordas-hoc: Done in 5b9baf68a — removed the comment.

docs/guide/currency-handling.md
```diff
@@ -0,0 +1,343 @@
...
+
+Side-by-side comparison of the default formatter, the docs adapter from the section below, and Excel:
+
+| Format | `TEXT(1234.5, ...)` without callback | With docs adapter callback | Excel |
+|---|---|---|---|
+| `"$0.00"` | `"$1234.50"` | `"$1234.50"` | `"$1234.50"` |
+| `"$#.00"` | `"$1234.50"` | `"$1234.50"` | `"$1234.50"` |
+| `"$#,##0.00"` | `"$1235,##0.00"` (no grouping) | `"$1,234.50"` | `"$1,234.50"` |
+| `"[$€-2] #,##0.00"` | `"[$€-2] 1235,##0.00"` (no grouping) | `"1.234,50 €"` | `"1.234,50 €"` |
+| `"$#,##0.00;($#,##0.00)"` (value `-1234.5`) | `"$-1235,##0.00;($#,##0.00)"` (no grouping) | `"($1,234.50)"` | `"($1,234.50)"` |
```
sequba:
```suggestion
| Format | Without callback | With adapter callback (section below) |
|---|---|---|
| `"$0.00"` | `"$1234.50"` | `"$1234.50"` |
| `"$#.00"` | `"$1234.50"` | `"$1234.50"` |
| `"$#,##0.00"` | `"$1235,##0.00"` (no grouping) | `"$1,234.50"` |
| `"[$€-2] #,##0.00"` | `"[$€-2] 1235,##0.00"` (no grouping) | `"1.234,50 €"` |
| `"$#,##0.00;($#,##0.00)"` (value `-1234.5`) | `"$-1235,##0.00;($#,##0.00)"` (no grouping) | `"($1,234.50)"` |
```

docs/guide/currency-handling.md
```diff
@@ -0,0 +1,343 @@
...
+  // Not a recognized currency format — let HyperFormula fall through
+  // to the built-in number formatter.
+  return undefined
+}
+```
+<!-- /snippet:currency-adapter -->
+
+#### Limitations of the reference adapter
+
+The adapter above is a compact, illustrative example — not a complete Excel format engine. Two things are worth calling out before you rely on it:
```
sequba:
```suggestion
```

sequba (approved):

PR #1692, opened 2026-06-09
Function metadata API: getAvailableFunctions / getFunctionDetails (HF-249)

marcin-kordas-hoc:
## What & why

A function picker — the Formula Builder's, or any integrator's — needs to answer two questions: *what functions exist?* and *what do this function's arguments mean?* HyperFormula could only answer the first. `getRegisteredFunctionNames()` returned 423 translated names and nothing else; `getAllFunctionPlugins()` exposed `implementedFunctions`, which has coercion rules and arity but no category, description, human-readable parameter names or examples. Everything else a picker needs existed only as prose, in the 551-line hand-maintained `docs/guide/built-in-functions.md`.

And prose with no second copy drifts silently. That page published `FVSCHEDULE` as `FV(Pv, Schedule)`, `RANDBETWEEN` as `RAND(...)`, `COLUMN` as `COLUMNS(...)`, `F.TEST` with `Z.TEST`'s signature, `T.TEST` with two of its four arguments, `DAYS360`'s arguments reversed; it listed `NORMDIST` twice (the second row was really `NORMSDIST`) and omitted `VERSION`, callable and counted in the page's own printed total since 2020. Every one is a copy-paste from a neighbouring row — the signature of data nothing can check against the thing it describes.

**So this PR moves that data into `src/` and reads it back through the engine's own public API, making the published reference just its first consumer.** A new authored catalogue under `src/interpreter/functionMetadata/` (370 entries, one file per category, 730 parameter descriptions and 679 examples) supplies category, description, `snake_case` parameter names and descriptions, examples and a docs link; `implementedFunctions` still supplies arity and optionality. Two methods expose the join, static and per-instance:

- **`getAvailableFunctions(code)`** → one cheap entry per function (`localizedName`, `canonicalName`, `category`, `shortDescription`, plus `aliasOf` on the 53 aliases), sorted by localized name under a collator built from that language — enough to paint all 423 picker rows without a second call.
- **`getFunctionDetails(canonicalName, code)`** → adds the ordered `parameters` (name, description, `optional`), `repeatLastArgs`, `documentationUrl` and `examples`. Deliberately **no pre-rendered syntax string**; the caller composes `SUMIF(range, criteria, [sum_range])` itself, and `script/formatFunctionSyntax.ts` is a reference implementation of that renderer, kept out of `src/` on purpose.

`docs/guide/built-in-functions.md` becomes a build product of *the public API* rather than of the catalogue directly — so generating the page exercises the same alias resolution, listability gate and optionality derivation a customer's picker will. It leaves git, is gitignored, and is regenerated from `built-in-functions.tmpl.md` as the first step of `docs:dev`/`docs:build`. "The docs are wrong" and "the API is wrong" are now the same bug.

The regenerated page adds `VERSION`, gives `NORMSDIST` its own row, adds a table of contents and a per-function anchor, and corrects optionality on **27 functions** the old page showed as required (`IF`, `LOG`, `ROUND`, `SUMIF`, `VLOOKUP`, the `*2*` conversions, …). Nine more change only *notation*: a repeating argument group is now rendered as `...` against `repeatLastArgs` instead of the old hand-written `[Range2, Criterion2 [, ...RangeN, CriterionN]]`. `SWITCH` was also semantically wrong — its parameters were `expression, value1, expression2`, but the third argument is the *result* returned on a match.

## Design decisions worth a second opinion

- **The catalogue is authored data, not derived**, so it must be kept in step with `implementedFunctions` by hand. Parameter *count* is cross-checked, and on a mismatch **the implementation wins**: `getFunctionDetails` reports one parameter per implemented argument under positional names (`Arg1`, `Arg2`, …), discards the authored names and descriptions, and warns on the console naming the function. Category, description, examples and URL still come from the entry, and the function stays listed in both tiers — drift costs the parameter prose, never the availability. `DEV_DOCS.md` documents this and the remaining silent-failure mode (an entry left behind after a rename describes nothing and merely ships in the bundle).
- **Optionality is deliberately not authored or cross-checked** — `optional` comes only from `optionalArg`/`defaultValue`. Hence the single production edit outside the new module: `optionalArg: true` on `SHEET`/`SHEETS`, which have always accepted `=SHEET()` while declaring the argument required. Metadata-only and behaviour-neutral (`runFunctionWithReferenceArgument` returns before argument-count validation), and a sweep of all 423 ids found no other function with this mismatch.
- **One rule decides how a function is described: does the catalogue hold an entry for its id?** The catalogue is keyed by id, not by implementation, so a user plugin registered over `SUMIF` is described with `SUMIF`'s authored category and description, over its own signature. An earlier revision gated this on a snapshot of built-in plugin ownership, so a shadow reported as `'Custom'`; that is gone. It bought little — a plugin re-implementing `SUMIF` is usually still a `SUMIF` — and cost a module-init hook in `index.ts` (the plugin identities can only come from the plugin barrel, and importing it from the registry creates a load-order cycle that breaks the bundled build) plus a second way for the whole built-in set to silently degrade to `'Custom'` if that hook ever failed to run. **This is the bullet I'd most like a second opinion on.**
- **Both tiers describe every registered function, custom ones included.** `registerFunctionPlugin` is global, so a custom function is callable everywhere and the static methods list it; the instance methods list that instance's own registry, which differs when it was built with the `functionPlugins` option. An id with no translation entry for the active language is omitted from both, because the interpreter refuses to evaluate it.
- **A custom function omits the fields it cannot author**, rather than reporting an empty one: `shortDescription`, `documentationUrl` and `examples` are absent (and optional in the public types, as `aliasOf` already was), so a consumer can tell "no authored description" from "an empty one" and the object survives `JSON.stringify` unchanged. Built-ins are unaffected — the catalogue authors all three for every entry.
- **An instance describes its functions under the translation package it was built with**, not under whatever is registered globally for that code today. Otherwise re-registering a language could make the API advertise a localized name that instance refuses to evaluate.
- **Exported:** `FunctionListEntry`, `FunctionDetails`, `FunctionParameterDescription`, `FunctionCategory`. `FUNCTION_CATEGORIES`, `FunctionDoc` and `CUSTOM_FUNCTION_CATEGORY` stay internal — so a TS consumer cannot enumerate the categories to build a filter and must compare against `'Custom'` as a string. Worth confirming that is the right line.
- **`canonicalName` is matched exactly**: case-sensitive (`'sumif'` → `undefined`, though `=sumif(...)` evaluates) and canonical English only. Likeliest integration pitfall for a picker holding translated names.

## Already reviewed

Roughly three-quarters of the develop-diff is already reviewed and merged, as sub-PRs into this branch: **#1699** (page generated from the API), **#1705** (HF-300: examples, docs URLs, parameter descriptions), **#1709** (`snake_case` parameter names), **#1710** (invalid-locale collator guard). New here: the metadata API itself, the catalogue-keyed resolution rule, custom functions in the static tier, `SHEET`/`SHEETS`, the generated table of contents, and ~55 descriptions rewritten because they documented Excel rather than HyperFormula — with the deviations added to `list-of-differences.md` (`INT` truncates toward zero, `MOD` takes the dividend's sign, `ISEVEN`/`ISODD` don't truncate, `CEILING.MATH`/`FLOOR.MATH` honour only `mode` = 1).

## How I tested

Paired suite: [handsontable/hyperformula-tests#14](https://github.com/handsontable/hyperformula-tests/pull/14) (branch `feature/hf-249-function-metadata-api`), 134 tests for this API alone, green with the full repository suite (502 files, 6,214 tests). It covers the static/instance split, i18n across all 18 packs, aliases, custom functions, plugins shadowing a built-in id or a built-in alias id, locale-aware ordering, and prototype-key ids (`toString`, `__proto__`) — plus the two invariants most worth protecting: **every canonical id declared by a registered built-in plugin resolves to details**, so a missing catalogue entry fails CI instead of silently dropping a function, and **the list and the details always agree on which ids exist**. Each guard was mutation-tested: broken deliberately, confirmed red, reverted. Assertions avoid jest-only matchers and never rely on jest ignoring a key valued `undefined`, so they fail under the jasmine/karma browser job too. Separately, all 679 authored examples parse and name their own function, and a sampled slice is pinned to Excel-cross-checked values.

## Known trade-offs

- **`examples` are English-spelled and `OFFSET` is lexed from its translated name**, so `getFunctionDetails('OFFSET','deDE').examples` yields `#NAME?` in all 16 non-English packs — and `ISREF`'s example embeds `OFFSET`, returning `true` in enGB but `false` in plPL with no error. The one item I'd want accepted with eyes open.
- **`SWITCH` publishes `repeatLastArgs: 1`**, understating its (value, result) pair group. It cannot simply become `2`: the field also drives runtime arity validation, and the optional trailing default needs a step of 1.
- **`documentationUrl` is the same page for all 423 ids.** Per-function anchors now exist on the generated page, so `#${canonicalName}` is a follow-up, not a redesign.
- **The catalogue ships in the bundle** (~25 KB gzipped) and is not tree-shakeable — `HyperFormula` and `FunctionRegistry` both import it eagerly.

## Open for the reviewer

- **`CHANGELOG.md`** names only the two methods; it should also name the four exported types, and needs a `### Changed` line for `SHEET`/`SHEETS` now reporting their argument as optional.
- **`DEV_DOCS.md` carries general engineering policy** unrelated to HF-249 (a `## Performance` section, six code-style bullets, and additions to Definition of Done, Automatic tests and Documentation) — which the atomic-PR rule added in this same PR says belongs elsewhere. Split them out, or accept them explicitly.

Source: https://app.clickup.com/t/9015210959/HF-249

---

> [!NOTE]
> **Medium Risk**
> Introduces a large, permanent public API and ships the full catalogue in the bundle (~25 KB gzipped), while changing how published function reference docs are produced; runtime formula evaluation is largely unchanged aside from metadata alignment (e.g. optional reporting for zero-arg reference functions).
> 
> **Overview**
> Adds **`getAvailableFunctions`** and **`getFunctionDetails`** (static and instance) so integrators can build function pickers from engine data instead of scraping docs. Metadata is authored in a new per-category catalogue under `src/interpreter/functionMetadata/` (joined with `implementedFunctions` for arity, optionality, and `repeatLastArgs`); custom functions appear with category `'Custom'` and positional `ArgN` names unless a plugin shadows a built-in id, in which case the catalogue entry for that id still applies.
> 
> The hand-maintained **`docs/guide/built-in-functions.md`** is removed from version control and **regenerated** at build time from `built-in-functions.tmpl.md` plus the same API (`npm run docs:generate-function-docs`, wired into `docs:dev` / `docs:build`). VuePress excludes the template from routes and disables “edit this page” on the generated guide.
> 
> Also exports **`FunctionListEntry`**, **`FunctionDetails`**, **`FunctionParameterDescription`**, and **`FunctionCategory`**; documents the catalogue workflow in **`DEV_DOCS.md`**; expands **`repeatLastArgs`** guidance in the custom-functions guide; and records additional Excel vs HyperFormula differences in **`list-of-differences.md`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f22560e555fe449e8bb810a220b800ad93b3fd9c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


marcinkordas (commented):
all perfect

src/HyperFormula.ts
```diff
@@ -649,6 +652,76 @@ export class HyperFormula implements TypedEmitter {
...
+   * @category Static Methods
+   */
+  public static getAvailableFunctions(code: string): FunctionListEntry[] {
+    validateArgToType(code, 'string', 'code')
+    const language = this.getLanguage(code)
+    const translate = (id: string) => language.getMaybeFunctionTranslation(id)
+    return getCanonicalFunctionIds(FunctionRegistry.getPlugins())
+      .filter(id => FUNCTION_DOCS[id] !== undefined)
+      .map(id => buildFunctionListEntry(id, FUNCTION_DOCS[id], translate))
+      .sort((a, b) => a.category === b.category ? a.canonicalName.localeCompare(b.canonicalName) : a.category.localeCompare(b.category))
```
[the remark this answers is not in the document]
marcin-kordas-hoc: Fixed in d86f8696f. `getAvailableFunctions` now gates each id on `FunctionRegistry.getFunctionPlugin(id) !== undefined` — the same resolution `getFunctionDetails` performs — so the list can never advertise a function the details lookup can't resolve (e.g. after `unregisterFunction`). Added a regression test (`unregisterFunction('SUM')` → absent from the list and `getFunctionDetails` returns `undefined`).

Note on the restricted-`functionPlugins` instance case: instance methods intentionally delegate to the static ones against the global registry (a deliberate spec decision so static and instance return the *same* catalogue and avoid a registered-count discrepancy). List and details stay mutually consistent; surfacing an instance-specific subset is out of scope for this MVP.

src/HyperFormula.ts
```diff
@@ -4358,6 +4433,51 @@ export class HyperFormula implements TypedEmitter {
...
+   * const hfInstance = HyperFormula.buildEmpty();
+   *
+   * // get the details of the SUMIF function, translated for the configured language
+   * const details = hfInstance.getFunctionDetails('SUMIF');
+   * ```
+   *
+   * @category Helpers
+   */
+  public getFunctionDetails(canonicalName: string): FunctionDetails | undefined {
+    return HyperFormula.getFunctionDetails(canonicalName, this._config.language)
```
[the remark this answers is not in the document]
marcin-kordas-hoc: Intentional, per the spec (v2.2): the instance methods delegate to the static ones against the global registry so the static and instance catalogues are identical and language-independent for default-built instances. Because the list is filtered by FUNCTION_DOCS, protected ids (OFFSET/VERSION) and aliases never surface, so the global-vs-instance registry difference does not change the output for built-ins; surfacing an instance-specific restricted subset is deliberately out of scope for this MVP. Resolving as by-design.

src/HyperFormula.ts
```diff
@@ -649,6 +652,78 @@ export class HyperFormula implements TypedEmitter {
...
+   */
+  public static getAvailableFunctions(code: string): FunctionListEntry[] {
+    validateArgToType(code, 'string', 'code')
+    const language = this.getLanguage(code)
+    const translate = (id: string) => language.getMaybeFunctionTranslation(id)
+    return getCanonicalFunctionIds(FunctionRegistry.getPlugins())
+      // documented AND still registered — gate on the same resolution `getFunctionDetails` uses, so the list never
+      // advertises a function the details lookup cannot resolve (e.g. after `unregisterFunction`).
+      .filter(id => FUNCTION_DOCS[id] !== undefined && FunctionRegistry.getFunctionPlugin(id) !== undefined)
+      .map(id => buildFunctionListEntry(id, FUNCTION_DOCS[id], translate))
```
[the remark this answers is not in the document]
marcin-kordas-hoc: Fixed in c98e9b167. Extracted a shared `resolveListableMetadata` (documented + registered + catalogue-arity == implementation-arity) used by BOTH getAvailableFunctions (filter) and getFunctionDetails (gate), so the list and details apply identical inclusion logic — a drifted function is now excluded from both. Added a parity regression test.

qunabu:
Task linked: [HF-249 HF function documentation available via API](https://app.clickup.com/t/9015210959/HF-249)

src/interpreter/functionMetadata/FunctionDescription.ts
```diff
@@ -0,0 +1,86 @@
...
+  /** Ordered; length MUST equal the function's `implementedFunctions.parameters` length (implementation arity). */
+  parameters: ParameterDoc[],
+}
+
+/**
+ * Public: a single entry of the short function list returned by `getAvailableFunctions`.
+ */
+export interface FunctionListEntry {
+  /** Function name, translated for the active language. */
+  name: string,
```
sequba:
```suggestion
  localizedName: string,
```

src/interpreter/functionMetadata/FunctionDescription.ts
```diff
@@ -0,0 +1,86 @@
...
+  /** `true` for the last `repeatLastArgs` parameters. */
+  repeatable: boolean,
+}
+
+/**
+ * Public: the full details for one function returned by `getFunctionDetails`.
+ */
+export interface FunctionDetails {
+  /** Function name, translated for the active language. */
+  name: string,
```
sequba:
```suggestion
  localizedName: string,
```

src/interpreter/functionMetadata/FunctionDescription.ts
```diff
@@ -0,0 +1,86 @@
...
+ * Copyright (c) 2025 Handsoncode. All rights reserved.
+ */
+
+/**
+ * The function categories, matching the `### <Category>` headers in `docs/guide/built-in-functions.md`.
+ */
+export const FUNCTION_CATEGORIES = [
+  'Array manipulation', 'Database', 'Date and time', 'Engineering',
+  'Information', 'Financial', 'Logical', 'Lookup and reference',
+  'Math and trigonometry', 'Matrix functions', 'Operator', 'Statistical', 'Text',
```
sequba, line 25: Please, make sure that these are the same categories, as official Excel docs use

src/interpreter/functionMetadata/FunctionDescription.ts
```diff
@@ -0,0 +1,86 @@
...
+ */
+export interface FunctionParameterDescription {
+  /** Human-readable parameter name. */
+  name: string,
+  /** What the argument does. Present but empty (`''`) in the MVP; populated in a later phase. */
+  description: string,
+  /** `true` when the argument may be omitted. */
+  optional: boolean,
+  /** `true` for the last `repeatLastArgs` parameters. */
+  repeatable: boolean,
```
sequba: `repeatLastArgs` is a number, so a boolean value here is not sufficient to express it - for some functions, the last 2 or 3 arguments need to be repeated

src/interpreter/functionMetadata/FunctionDescription.ts
```diff
@@ -0,0 +1,86 @@
...
+export interface FunctionDetails {
+  /** Function name, translated for the active language. */
+  name: string,
+  /** Language-independent function id, e.g. `'SUMIF'`. */
+  canonicalName: string,
+  category: FunctionCategory,
+  /** One-liner description (English in the MVP). */
+  shortDescription: string,
+  /** Generated from the parameter names, optionality and `repeatLastArgs`. */
+  syntax: string,
```
sequba: I'm not sure if we should generate the syntax string. `getFunctionDetails` returns the parameter list and all the details about them (`optional`, `repeatLastAgrs` etc) so I think FormulaBuilder should be responsible for rendering the syntax string

docs/guide/custom-functions.md
```diff
@@ -538,3 +538,42 @@ MyCustomPlugin.translations = {
...
+This metadata covers **built-in functions only**. Custom functions are not
+included: they don't appear in the `getAvailableFunctions()` list, and
+`getFunctionDetails()` returns `undefined` for them.
+
+```js
+HyperFormula.registerFunctionPlugin(MyCustomPlugin, MyCustomPlugin.translations);
+
+// a custom function is not part of the built-in metadata
+const details = HyperFormula.getFunctionDetails('MY_FUNCTION', 'enGB'); // undefined
+```
```
sequba: It SHOULD include custom functions

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Added
 
+- Added the `getAvailableFunctions()` and `getFunctionDetails()` methods (both static and instance) for retrieving built-in function metadata: category, translated name, short description, generated syntax, and parameters. [#1692](https://github.com/handsontable/hyperformula/pull/1692)
```
```suggestion
- Added the `getAvailableFunctions()` and `getFunctionDetails()` methods (both static and instance) for retrieving function metadata. [#1692](https://github.com/handsontable/hyperformula/pull/1692)
```

sequba (changes requested):

src/interpreter/functionMetadata/FunctionDescription.ts
marcin-kordas-hoc: Done — renamed the field to `localizedName` on `FunctionListEntry` and updated the builder, the API, docs and tests.

src/interpreter/functionMetadata/FunctionDescription.ts
marcin-kordas-hoc: Done — `localizedName` here too on `FunctionDetails`.

src/interpreter/functionMetadata/FunctionDescription.ts
marcin-kordas-hoc: Checked against the official Excel docs (https://support.microsoft.com/en-us/office/excel-functions-by-category-5f91f4e9-7b42-46d2-9bd1-63f26a86c0eb). The ten categories that have an Excel equivalent already use the same names as that page — it spells them out in full ("Math and trigonometry", "Lookup and reference", "Date and time"), not the abbreviated ribbon labels ("Math & Trig" etc.), so no change was needed there. The remaining three — `Array manipulation`, `Matrix functions` and `Operator` — are HyperFormula-specific and have no Excel equivalent. Added a note documenting this on the `FUNCTION_CATEGORIES` constant.

src/interpreter/functionMetadata/FunctionDescription.ts
marcin-kordas-hoc: Done — dropped the per-parameter `repeatable` boolean and now expose `repeatLastArgs` as a number on `FunctionDetails` (straight from the implementation metadata), so the cases where the last 2/3 args repeat are expressed correctly.

src/interpreter/functionMetadata/FunctionDescription.ts
marcin-kordas-hoc: Agreed — dropped the generated `syntax` string entirely. `getFunctionDetails` now returns the parameter list (`name`, `description`, `optional`) plus `repeatLastArgs`, and the consumer (FormulaBuilder) renders the syntax string from that.

docs/guide/custom-functions.md
marcin-kordas-hoc: Done. Custom functions are instance-scoped, so the instance methods now read the instance own registry and include both custom functions and aliases; the static methods cover the globally-registered built-ins and their aliases. Custom functions ship no catalogue entry, so their `category` is `undefined` and `shortDescription` is empty, with positional parameter names. Rewrote this section of the guide accordingly.

CHANGELOG.md
marcin-kordas-hoc: Applied your suggested wording.

docs/guide/custom-functions.md
```diff
@@ -538,3 +538,60 @@ MyCustomPlugin.translations = {
...
+
+`getAvailableFunctions()` returns entries sorted alphabetically by their
+localized name, each with `localizedName`, `canonicalName`, `category`, and
+`shortDescription`. `getFunctionDetails()` returns the same fields plus the
+ordered `parameters` list (each with `name`, `description`, and `optional`) and
+`repeatLastArgs` — the number of trailing parameters that repeat for functions
+with a variable number of arguments (`0` for a fixed argument list). The
+methods do not return a pre-rendered syntax string; build it from `parameters`
+and `repeatLastArgs` as your UI needs.
```
sequba:
```suggestion
with a variable number of arguments (`0` for a fixed argument list).
```

We don't need to describe what is NOT there

sequba (changes requested):

docs/guide/custom-functions.md
marcin-kordas-hoc: Agreed, dropped that sentence.

docs/guide/custom-functions.md
```diff
@@ -538,3 +538,60 @@ MyCustomPlugin.translations = {
 };
 ```
 :::
+
+## Function metadata
```
sequba: I think it's better to describe `getAvailableFunctions` and `getFunctionDetails` functions in the built-in-functions page instead of custom-functions

sequba (changes requested):

sequba:
@cursor fix the review comment

src/HyperFormula.ts
```diff
@@ -649,6 +652,151 @@ export class HyperFormula implements TypedEmitter {
...
+    const collator = new Intl.Collator(HyperFormula.toBcp47Locale(languageCode))
+    return functionIds
+      .map(id => {
+        const resolved = HyperFormula.resolveFunctionMetadata(id, getPlugin(id))
+        if (resolved === undefined) {
+          return undefined
+        }
+        return resolved.doc !== undefined
+          ? buildFunctionListEntry(id, resolved.doc, translate)
+          : buildCustomFunctionListEntry(id, translate)
```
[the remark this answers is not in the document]
sequba: ignore it, it will be changed soon

docs/guide/custom-functions.md
sequba: @cursor fix it

src/HyperFormula.ts
```diff
@@ -649,6 +652,151 @@ export class HyperFormula implements TypedEmitter {
...
+   * ```
+   *
+   * @category Static Methods
+   */
+  public static getFunctionDetails(canonicalName: string, code: string): FunctionDetails | undefined {
+    validateArgToType(canonicalName, 'string', 'canonicalName')
+    validateArgToType(code, 'string', 'code')
+    const language = this.getLanguage(code)
+    return HyperFormula.buildFunctionDetailsFor(canonicalName, FunctionRegistry.getFunctionPlugin(canonicalName), language)
+  }
```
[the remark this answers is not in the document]
sequba: @cursoragent update the guides if they are not correct

sequba (approved):

docs/guide/custom-functions.md
marcin-kordas-hoc: Done — the Function metadata section now lives in built-in-functions.md (#function-metadata). custom-functions.md only keeps the custom-function caveat (custom functions don't appear in the static API) and links to that section.

sequba (approved):

src/interpreter/functionMetadata/buildFunctionDescriptions.ts
```diff
@@ -0,0 +1,138 @@
...
+export const DEFAULT_DOCUMENTATION_URL = 'https://hyperformula.handsontable.com/docs/guide/built-in-functions.html'
+
+/**
+ * Returns whether a parameter may be omitted: it declares `optionalArg`, or it has a `defaultValue`.
+ *
+ * @param {FunctionArgument | undefined} arg - the structural argument metadata, or `undefined`
+ */
+export function isParameterOptional(arg: FunctionArgument | undefined): boolean {
+  return arg?.optionalArg === true || arg?.defaultValue !== undefined
+}
```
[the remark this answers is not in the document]
marcin-kordas-hoc: Confirmed and fixed in d0d3279.

`SHEET` and `SHEETS` now declare `optionalArg: true` on their single parameter, so `getFunctionDetails()` reports `optional: true`.

Root cause: zero-argument support for these two never reaches argument-count validation. `runFunctionWithReferenceArgument` serves `args.length === 0` straight from `noArgCallback()`, so `runFunction` (and with it `isNumberOfArgumentValuesValid`) is bypassed and optionality is only expressible through the flag. `ROW` and `COLUMN` use the same helper and already carried it. `ISFORMULA` correctly does not, because its no-arg callback returns `#N/A!` — so it stays `optional: false`.

Worth noting the flag was the only place that disagreed. The plugin evaluates `=SHEET()` / `=SHEETS()` (covered by `function-sheet.spec.ts` and `function-sheets.spec.ts`), the catalogue copy in this PR says "An optional sheet name…" with `=SHEET()` as the first example, and the published docs already list the syntax as `SHEET([Value])` / `SHEETS([Value])`.

No effect on evaluation: `optionalArg` has exactly two readers, `isParameterOptional` (metadata) and `isNumberOfArgumentValuesValid`. The latter is reachable only from `runFunction`, which for these two functions only ever runs with exactly one argument — `0` short-circuits above it and `> 1` returns `WrongArgNumber` earlier — so `numberOfArgumentValuesPassed === argumentsMetadata.length` regardless of the flag.

Test guard in handsontable/hyperformula-tests#14: pins `SHEET`/`SHEETS`/`ROW`/`COLUMN` as optional and `ISFORMULA` as required. With the engine change reverted, exactly the `SHEET` and `SHEETS` cases fail and the other three still pass.

No changelog entry — the metadata API is unreleased, so the flag was never wrong in a shipped version.

All 33 checks on this PR are green (3 skipped).

qunabu:
Task linked: [HF-300 Params descriptions, links and examples](https://app.clickup.com/t/9015210959/HF-300)

src/interpreter/functionMetadata/FunctionDescription.ts
```diff
@@ -0,0 +1,169 @@
...
+  /**
+   * Function category: one of the documented categories for a built-in function, `'Custom'` for a custom
+   * (user-registered) one, which ships no catalogue entry. Always present.
+   */
+  category: FunctionCategory,
+  /**
+   * One-liner description (English in the MVP). Empty (`''`) for custom functions.
+   * May contain inline markdown links (absolute URLs) and `<br>` line breaks.
+   */
+  shortDescription: string,
```
sequba: this should be optional in the return type of `getAvailableFunctions`

src/interpreter/functionMetadata/FunctionDescription.ts
```diff
@@ -0,0 +1,169 @@
...
```
sequba: This should be optional in the return type of `getFunctionDetails`

src/interpreter/functionMetadata/FunctionDescription.ts
```diff
@@ -0,0 +1,169 @@
...
+   * Always an integer in `[0, parameters.length]`, for a custom (user-registered) function too: a declared value the
+   * interpreter would not honour (negative, fractional, `NaN`, `Infinity`) is reported as `0`, and a valid count
+   * larger than the parameter list is clamped to it, so "the last N of M parameters repeat" is never nonsense.
+   */
+  repeatLastArgs: number,
+  /**
+   * Link to the function's documentation. Always a string: the link authored in the catalogue for built-ins, `''`
+   * for custom functions.
+   */
+  documentationUrl: string,
```
sequba: This should be optional in the return type of `getFunctionDetails`

src/interpreter/functionMetadata/FunctionDescription.ts
```diff
@@ -0,0 +1,169 @@
...
+  /** Usage examples (English). At least one per built-in function; `[]` for custom functions. */
+  examples: string[],
```
sequba: This should be optional in the return type of `getFunctionDetails`

.eslintignore
```diff
@@ -21,7 +25,6 @@ doc
 es
 languages
 lib
-script
```
sequba: Revert this change

sequba (changes requested):

src/index.ts
```diff
@@ -109,13 +111,23 @@ const defaultLanguage = Config.defaultConfig.language
...
+    builtinPlugins.push(plugin)
   }
 }
+// Snapshot the original built-in id -> plugin ownership, so the function-metadata API can tell a built-in's own
+// implementation from a user plugin that shadows its id. Captured here because the plugin identities can only come
+// from the plugin barrel, and this is the only place that already imports it — importing it from the registry or the
+// metadata builders creates a module-load-order cycle that crashes the bundled build ("Class extends value
+// undefined"). Skipping this call costs the built-ins their categories and descriptions, not their presence: which
+// ids the metadata API describes comes from the catalogue (`FunctionRegistry.isBuiltinFunctionId`), not from here.
+FunctionRegistry.captureBuiltinFunctionOwners(builtinPlugins)
```
sequba: Remove this mechanism. Let's not care about shadowing of built-in functions. We have changes already planned that will solve this issue in different way.

sequba:
@cursor address all unresolved code review comments.

docs/guide/custom-functions.md
```diff
@@ -142,6 +142,10 @@ method:
 HyperFormula.registerFunctionPlugin(MyCustomPlugin, MyCustomPluginTranslations);
 ```
 
+::: tip
+Custom functions registered with `registerFunctionPlugin()` are available in all HyperFormula instances, and both the static metadata API methods (`HyperFormula.getAvailableFunctions()` and `HyperFormula.getFunctionDetails()`) and their instance counterparts (`hfInstance.getAvailableFunctions()` and `hfInstance.getFunctionDetails()`) describe them. Custom functions ship no metadata catalogue entry, so they are reported with `category: 'Custom'`, an empty `shortDescription` and `documentationUrl`, no `examples`, and positional parameter names (`Arg1`, `Arg2`, ...). A plugin passed to a single engine through the [`functionPlugins`](../api/interfaces/configparams.md#functionplugins) configuration option is not registered globally, so only that instance's methods describe it. See [`getAvailableFunctions`](../api/classes/hyperformula.md#getavailablefunctions) and [`getFunctionDetails`](../api/classes/hyperformula.md#getfunctiondetails) in the API reference for more details.
+:::
+
```
sequba: Remove this section

src/index.ts
```diff
@@ -109,11 +111,14 @@ const defaultLanguage = Config.defaultConfig.language
 HyperFormula.registerLanguage(defaultLanguage, enGB)
 HyperFormula.languages[enGB.langCode] = enGB
 
+const builtinPlugins: FunctionPluginDefinition[] = []
```
sequba: This var seems to be unused. If so, remove it please

src/interpreter/functionMetadata/buildFunctionDescriptions.ts
sequba, line 1: What is this file about? It's called `buildFunctionDescriptions.ts`. but function description is alread present in catalogue, and doesn't need to be build? Perhaps this file should be renamed

sequba (changes requested):

src/interpreter/functionMetadata/FunctionDescription.ts
sequba: still not changed

src/interpreter/functionMetadata/FunctionDescription.ts
sequba: still not changed

src/interpreter/functionMetadata/FunctionDescription.ts
sequba: still not changed

src/interpreter/functionMetadata/FunctionDescription.ts
sequba: still not changed

PR #1697, opened 2026-06-22
fix(lookup): skip empty cells in approximate MATCH/VLOOKUP/HLOOKUP/XLOOKUP (HF-223)

marcin-kordas-hoc:
## What & why
Approximate `MATCH`, `VLOOKUP`, `HLOOKUP` and `XLOOKUP` returned `#N/A` (or, in descending mode, the wrong position) when the sorted search range contained genuinely empty cells. Excel 2021 and Google Sheets skip empty cells when computing the lower/upper bound; HyperFormula instead landed on an empty cell during binary search (its `EmptyValue` Symbol never matched the key). [HF-223]

## Root cause
`findLastOccurrenceInOrderedRange` (`src/interpreter/binarySearch.ts`): `compare()` ranks `EmptyValue` below every value, breaking the sort invariant the binary search relies on, and the `typeof foundValue !== typeof searchKey` guard then turned a landing-on-empty into `#N/A`. Shared by approximate `MATCH(±1)`, sorted `VLOOKUP`/`HLOOKUP`, and `XLOOKUP(searchMode ±2)`.

## Fix
- The binary search runs directly over the original range while tracking whether the descent ever probes an empty cell. Empty cells are the only source of non-monotonicity in the search predicate, so a descent that never touches one is provably equivalent to a search over the range with empty cells removed and its result is trusted as-is — the common case stays `O(log n)`, including ranges that contain empty cells the descent happens not to touch.
- Only when the descent probes an empty cell does the search fall back to an `O(n)` compaction: the non-empty cell indices are collected, the binary search re-runs over the compacted list, and the result maps back to the original index space, so empty cells keep their slots and the matched non-empty cell's original 1-based position is reported unchanged. (In exact-match mode, a hit found by the direct descent is accepted after an equality re-check, skipping the fallback.)
- The "no match at all" bound cases return the position of the first **non-empty** cell (never the position of a leading empty cell), and the approximate-bound "next" position steps to the next *non-empty* index, so skipped empty slots never shift the reported position.
- `AdvancedFind.findNormalizedValue` skips `EmptyValue` on its in-memory ordered path for the same reason, keeping the linear and binary search modes consistent.
- A matched result cell that is empty is returned as `0` (`zeroIfEmptyResult`), matching Excel, for `VLOOKUP`/`HLOOKUP`/`XLOOKUP` including multi-cell `XLOOKUP` return arrays.
- Empty strings are unaffected (text ranks above numbers, so they still terminate a numeric run).

## Performance & edge cases
- **Complexity:** `O(log n)` whenever the binary-search descent probes no empty cell (always the case for gap-free ranges — the typical sorted-lookup workload); `O(n)` only when an empty cell actually interferes with the descent. A new `Sorted lookup` benchmark in the performance suite guards this fast path.
- **All-empty range:** returns `NOT_FOUND` directly (the `if_not_found` result / `#N/A`), never row 1.
- **Leading empty cells with the key outside the range:** the bound modes return the first non-empty position, consistent with the linear search modes on the same data.
- **Duplicates caveat:** in exact-match binary modes, when the range contains both duplicates of the key and interspersed empty cells, which duplicate is reported is unspecified (Excel's binary modes likewise leave this unspecified); documented in the function's JSDoc.

## Excel / Google Sheets parity
Behaviour verified against the latest Excel and Google Sheets, including exact-vs-blank, empty-string-not-skipped, and the descending early-stop case. For the leading-empty bound cases, HyperFormula's binary modes are consistent with its own linear modes (Excel documents binary search over blank-containing ranges as unreliable).

## Tests
Public test suite: handsontable/hyperformula-tests#17 (matching branch `task/hf-223-match-empty-cells`), including regression locks for the stepping and leading-empty edge cases and a `Sorted lookup` performance benchmark.

## Definition of Done
- [x] Production code + JSDoc
- [x] Tests (hyperformula-tests#17, matching branch)
- [x] Changelog (empty-cell search fix + empty-result → 0 coercion)
- [x] i18n — N/A (shared-logic fix, no function add/rename)
- [x] Docs — list-of-differences entry for the remaining Excel divergence

---

> [!NOTE]
> **Medium Risk**
> Changes shared lookup/binary-search logic used by MATCH and all major lookup functions, so incorrect edge-case handling could affect many formulas; scope is limited to lookup parity and is behavior-fix oriented rather than new surface area.
> 
> **Overview**
> Fixes **HF-223**: approximate lookups no longer break when the search range has **genuinely empty cells** interspersed among sorted values.
> 
> **Search behavior:** `findLastOccurrenceInOrderedRange` now treats `EmptyValue` as non-participating in ordering. It keeps an **O(log n)** binary search when the descent never hits an empty cell; if it does, it **compacts non-empty indices** and re-searches, with exact-match hits still accepted after an equality re-check. Lower/upper bound paths step to the **first/next non-empty** index instead of landing on blanks or returning row 1 on all-empty ranges. The in-memory path in `AdvancedFind.findNormalizedValue` skips empty cells the same way so linear and binary modes stay aligned.
> 
> **Return values:** `VLOOKUP`, `HLOOKUP`, and `XLOOKUP` coerce an **empty matched result cell to `0`**, matching Excel.
> 
> Changelog and **list-of-differences** document the fix and remaining Excel divergence on binary `XLOOKUP` over blank-heavy ranges.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d3a6f2138468daebf6454ad5bb9cab8eeb8f162f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/interpreter/binarySearch.ts
```diff
@@ -80,8 +98,8 @@ export function findLastOccurrenceInOrderedRange(
     }
 
     // orderingDirection === 'asc'
-    const nextIndex = foundIndex+1
-    return nextIndex <= end ? nextIndex - start : NOT_FOUND
```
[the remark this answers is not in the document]
marcin-kordas-hoc: Good catch — confirmed and fixed. With the empty-skip change, an all-empty range left `nonEmptyIndices` empty, so `foundIndex === NOT_FOUND` fell through to the `ifNoMatch` offset-0 branches (`returnLowerBound`+desc / `returnUpperBound`+asc) and approximate XLOOKUP returned row/column 1 instead of `#N/A`. `findLastOccurrenceInOrderedRange` now returns `NOT_FOUND` immediately when there are no non-empty cells, so an all-empty range yields `#N/A` for every direction/bound. Regression test added in the matching hyperformula-tests branch (`XLOOKUP (binary approximate) on an all-empty range returns if_not_found, not row 1`).

src/interpreter/binarySearch.ts
```diff
@@ -35,12 +41,36 @@ export function findLastOccurrenceInOrderedRange(
...
+  // This pre-scan is O(n) over the range, which trades away the binary search's O(log n) guarantee.
+  // It is required for correctness: with empty cells interspersed the search predicate is no longer
+  // monotonic, so the binary search cannot run directly on the original range. A future optimization
+  // could skip the compaction (and keep O(log n)) when the range is statically known to be empty-free.
+  const nonEmptyIndices: number[] = []
+  for (let index = start; index <= end; index++) {
+    if (getValueFromIndexFn(index) !== EmptyValue) {
+      nonEmptyIndices.push(index)
+    }
+  }
```
sequba: This step is slow. If I'm correct, it's not necessary in exact match mode (`ifNoMatch === 'returnNotFound'`), so let's skip it to keep this mode O(log n)

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Added an Indonesian (Bahasa Indonesia) language pack. [#1674](https://github.com/handsontable/hyperformula/pull/1674)
 
+### Fixed
+
+- Fixed approximate `MATCH`, `VLOOKUP`, `HLOOKUP`, and `XLOOKUP` incorrectly returning `#N/A` (or a wrong position) when the search range contained empty cells; empty cells are now skipped during approximate matching, consistent with Excel and Google Sheets. [HF-223]
```
```suggestion
- Fixed approximate `MATCH`, `VLOOKUP`, `HLOOKUP`, and `XLOOKUP` incorrectly returning `#N/A` (or a wrong position) when the search range contained empty cells. [#1697](https://github.com/handsontable/hyperformula/pull/1697)
```

sequba (changes requested):

qunabu:
Task linked: [HF-223 Fix MATCH incompatibility with Excel](https://app.clickup.com/t/9015210959/HF-223)

src/interpreter/binarySearch.ts
marcin-kordas-hoc: Done. Exact mode (`returnNotFound`) now runs the binary search directly over the original range, so it stays O(log n). It also keeps exact match from skipping blanks: with the shared compaction, `XLOOKUP(…, 0, 2)` over a range with an interspersed empty could land past the blank and report a match — now it does not. Added a regression test in hyperformula-tests#17.

CHANGELOG.md
marcin-kordas-hoc: Applied, thanks.

marcin-kordas-hoc:
**Live-Excel (M365) cross-check.** I ran every non-spill HF-223 case through Microsoft Graph against real Excel (one oracle workbook, 30 sheets mirroring the #17 specs, 34 assertions): **32/34 match Excel exactly.** The 2 that differ are both intentional "HF is stricter" cases, now in `list-of-differences.md`:

- **Exact match, binary mode, value *after* an interspersed empty** — HF finds it; Excel returns `#N/A`/if-not-found.
- **`XLOOKUP(…, match_mode 1, search_mode 2)` over an all-empty range** — HF returns the `if_not_found` result; **Excel returns row 1's value**. Our spec asserts "not row 1" on purpose.

One open question for you on that second one: is HF's "return `if_not_found`, never row 1" the contract we want for an all-empty binary-approx lookup, or should we mirror Excel's row-1 behavior? Happy either way — flagging since it's a deliberate divergence.

(All `#N/A` cases confirmed too; the M365 account is PL-locale so Excel returns `#N/D`.)

src/interpreter/plugin/LookupPlugin.ts
```diff
@@ -284,7 +295,8 @@ export class LookupPlugin extends FunctionPlugin implements FunctionPluginTypech
     }
 
     const returnValues: InternalScalarValue[][] = isVerticalSearch ? [returnRange.data[indexFound]] : returnRange.data.map((row) => [row[indexFound]])
-    return SimpleRangeValue.onlyValues(returnValues)
+    const coerced = returnValues.map((row) => row.map((value) => this.zeroIfEmptyResult(value)))
+    return SimpleRangeValue.onlyValues(coerced)
```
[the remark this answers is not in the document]
marcin-kordas-hoc: The empty-result → `0` coercion is intentional and Excel-verified, not a regression:

- **By design**: agreed in the HF sync (2026-07-07) — a reference to an empty matched cell coerces to `0` in Excel, so `VLOOKUP`/`HLOOKUP`/`XLOOKUP` mirror that on the return value (`MATCH` is untouched — it returns an index).
- **Confirmed against live Excel (M365)** this cycle: the empty-return cases (`VLOOKUP`/`HLOOKUP`/`XLOOKUP`) all return `0` in real Excel. `ISBLANK` on such a result being `FALSE` is therefore also Excel-consistent (Excel's lookup of an empty cell *is* `0`, not blank).
- **Changelog**: added an explicit entry for this behavior change in `5413f8da6`.

Fair point on **`INDEX`**: it isn't coerced, so `INDEX(...)` and `VLOOKUP(...)` differ for an empty target cell. Excel coerces both to `0`, so it's actually `INDEX` that's out of step — but aligning `INDEX` is outside HF-223's scope. @sequba — do you want a separate follow-up to coerce `INDEX` empty results too (for full Excel/consistency), or leave it as-is? Leaving this thread open for your call.

CHANGELOG.md
```diff
@@ -12,6 +12,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Added an Indonesian (Bahasa Indonesia) language pack. [#1674](https://github.com/handsontable/hyperformula/pull/1674)
 - Added a `stringifyCurrency` config option that lets you plug in a custom currency formatter for the `TEXT` function. [#1145](https://github.com/handsontable/hyperformula/issues/1145)
 
+### Fixed
+
+- Fixed approximate `MATCH`, `VLOOKUP`, `HLOOKUP`, and `XLOOKUP` incorrectly returning `#N/A` (or a wrong position) when the search range contained empty cells. [#1697](https://github.com/handsontable/hyperformula/pull/1697)
+- Changed `VLOOKUP`, `HLOOKUP`, and `XLOOKUP` to return `0` (matching Excel) when the matched cell is empty, instead of a blank value. [#1697](https://github.com/handsontable/hyperformula/pull/1697)
```
sequba: Let's consolidate them into a single entry.

```suggestion
- Fixed the behavior of `MATCH`, `VLOOKUP`, `HLOOKUP`, and `XLOOKUP` functions when the search range contained empty cells. [#1697](https://github.com/handsontable/hyperformula/pull/1697)
```

sequba (changes requested):

sequba (approved):

Issue #834, opened 2021-10-07
Add support for reversed ranges (previously: "...and multiple range references of cells")


AMBudnik:
### Discussed in https://github.com/handsontable/hyperformula/discussions/832

<sup>Originally posted by **AMBudnik** October  6, 2021</sup>
As discussed [here](https://forum.handsontable.com/t/error-when-using-a-reversible-range/5570) we should consider adding support for **reversed ranges** of cells and support **multiple range references** to match the standard of OpenFormula.

**Examples**
1. Reversed ranges - `=SUM(A10:A1)` should be equal to `=SUM(A1:A10)`
2. Multiple range references - `=SUM(B4:B5:C5)` should be equal to `=SUM(B4:C5)`

warpech:
We decided today that point 1 from the OP will be done in scope of this issue and point 2 from the OP will be extracted into a separate issue, for the future.

AMBudnik:
Thank you for adding this functionality @sequba 
I updated the Client who requested this functionality.

PR #1707, opened 2026-07-13
feat: implement SORT dynamic array function (HF-69)

marcin-kordas-hoc:
## What & why

Implements **SORT** (`HF-69`, child of HF-28 "Modern dynamic array functions", sibling of the shipped SEQUENCE and of VSTACK/HSTACK). Adds `SORT(array, [sort_index], [sort_order], [by_col])` as a dynamic array function.

Tests: handsontable/hyperformula-tests#24 (paired).
Sibling PR: #1708 (UNIQUE / HF-68).
ADR: `docs/adr/2026-07-13-sort-unique-array-functions.md`.

## Behavior

- Returns an array the **same shape** as the input.
- `sort_index` (default 1): 1-based index into the sort dimension.
- `sort_order`: `1` ascending (default) or `-1` descending.
- `by_col`: `FALSE` (default) reorders rows; `TRUE` reorders columns.
- Ordering reuses `ArithmeticHelper` (mixed types: numbers < text < logical; empties; locale collation via `caseSensitive`/`accentSensitive`) and is **stable** — ties keep input order.

## Design

Mirrors the SEQUENCE/FILTER machinery: `sizeOfResultArrayMethod` + `vectorizationForbidden: true`, runtime via `runFunction` returning `SimpleRangeValue`/`CellError`, parse-time size method returning a **fresh** `ArraySize` (the input's `isRef` flag is dropped — a ref-flagged size is treated as scalar and would collapse the spill).

## Notes — divergences from Excel (surfaced here + inline + in tests)

- **`sort_order` is strictly `{1, -1}`**; any other value → `#VALUE!`. Excel documents only `{1,-1}`; the reported "`sort_order=0` does not error" quirk is undocumented and **could not be re-verified against live Excel in this environment**, so the strict documented contract was chosen (see ADR `dec_2`, `con_1`). Flagged for live-Excel/Kuba confirmation.
- **Multi-key array-constant `sort_index`** (e.g. `{1,2}`) is **not supported in v1** (documented in `known-limitations.md`; ADR `dec_6`).
- In-range errors propagate (first error found; ADR `dec_7`).

## Error-type map

`sort_order ∉ {1,-1}` → `#VALUE!` (BadMode) · `sort_index < 1` → `#VALUE!` (LessThanOne) · `sort_index >` dimension → `#VALUE!` (ValueLarge) · in-range error → propagate · wrong arity → `#N/A`.

## Definition of Done

- [x] Production code (`SortPlugin.ts`, registered via `plugin/index.ts`)
- [x] i18n — all 17 language packs (authoritative MS Functions Translator names; enUS inherits enGB)
- [x] Tests (paired tests PR) — across the standard array-function groups, dual-env safe
- [x] Docs — `built-in-functions.md`, `known-limitations.md`
- [x] JSDoc on all methods
- [x] CHANGELOG entry
- [x] ADR with audit-verified citations

Source: https://app.clickup.com/t/86c89q1tt

---

> [!NOTE]
> **Low Risk**
> Self-contained new array function following existing SEQUENCE/FILTER patterns; no changes to auth, persistence, or core recalculation beyond registering one plugin.
> 
> **Overview**
> Adds the **SORT** dynamic array function: `SORT(Array, [SortIndex], [SortOrder], [ByCol])` returns the input range reordered by row (default) or column, same dimensions as the source.
> 
> Implementation lives in new `SortPlugin.ts`, wired like other array functions (`sizeOfResultArrayMethod`, `vectorizationForbidden`, spill size copied from input without propagating `isRef`). Sort keys use `ArithmeticHelper` (with empty cells forced last); invalid `sort_order` (not `1` or `-1`), bad `sort_index`, in-range errors, and empty ranges get the documented `#VALUE!` / `#N/A` / error propagation behavior.
> 
> Docs and changelog are updated; **known-limitations** documents single-key only, strict sort order, and HF comparison rules. **SORT** is added to all 17 language packs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f081be57e9651d13d7228fa33eea3a75ab4aaa28. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [HF-28 Modern dynamic array functions](https://app.clickup.com/t/9015210959/HF-28)
Task linked: [HF-68 Implement UNIQUE function](https://app.clickup.com/t/9015210959/HF-68)
Task linked: [HF-69 Implement SORT function](https://app.clickup.com/t/9015210959/HF-69)

src/interpreter/plugin/SortPlugin.ts
```diff
@@ -0,0 +1,144 @@
...
+          return SimpleRangeValue.onlyValues(result)
+        }
+
+        // Reorder rows; the key of row r is data[r][keyIndex].
+        const rows: InternalScalarValue[][] = data.map(row => row.slice())
+        rows.sort((r1, r2) => compare(
+          r1[keyIndex] as InternalNoErrorScalarValue,
+          r2[keyIndex] as InternalNoErrorScalarValue,
+        ))
+        return SimpleRangeValue.onlyValues(rows)
```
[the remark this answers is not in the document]
marcin-kordas-hoc: This doesn't reproduce — the premise (a zero-height/zero-width `range.data` reaching `SortPlugin.sort`) can't occur for any range literal.

**Why:** HyperFormula normalizes inverted range endpoints in the parser, before the ast ever reaches the interpreter. `ParserWithCaching.parse()` unconditionally runs every parsed ast through `convertReversedRangesToRegularRanges` (`src/parser/ParserWithCaching.ts:106`), which reorders `CELL_RANGE`/`COLUMN_RANGE`/`ROW_RANGE` start/end independently per axis (`ParserWithCaching.ts:112-136`, `orderCellRangeEnds` at `:174-184`). So `B2:B1`, `A2:B1`, `B1:A2` all normalize to the same ordered range (e.g. `A1:B2`) before `AbsoluteCellRange.fromCellRange` (`src/AbsoluteCellRange.ts:101-106`) ever builds the range object — that constructor does no swapping itself, it trusts the input is already ordered, and it always is for a parsed literal. This is intentional, documented behavior added in v2.0.0 ("Added support for reversed ranges", #834) and locked in by `test/hyperformula-tests/unit/parser/parser.spec.ts:598-689` (`describe('reversed range', ...)`), which includes the exact `B2:B1`-style case.

Result: for any range literal, `range.height() >= 1` and `range.width() >= 1` always hold by the time `SortPlugin.sort` runs, so `data`/`rows` can never be `[]`, and the `byCol` path (`data[keyIndex][c1]`) is equally safe for the same reason.

I also checked the other way a range could theoretically collapse to empty: if the referenced cells are deleted at runtime (e.g. `=SORT(B2:B5)` then deleting rows 2-5), the range doesn't degrade into an empty array either — `FunctionPlugin.coerceToType`'s `RANGE` branch (`src/interpreter/plugin/FunctionPlugin.ts:374-378`) short-circuits any `CellError` (here `#REF!`) before `SortPlugin.sort`'s body runs at all.

**Empirically verified** (local Jest run against this branch, `feature/HF-69-sort`): `=SORT(B2:B1)`, `=SORT(D1:A1)`, `=SORT(D2:A1)` (both axes inverted), `=SORT(A1:A1)`, `=SORT(D1:A2,1,1,TRUE())` (byCol path), a range collapsed via `removeRows`, and nested `=SORT(SORT(...))` — all evaluate cleanly (correct sorted/spilled output, or a graceful `#REF!` for the deletion case). No `TypeError`, no crash in any case.

Given this, I'm treating it as a false positive and not adding a guard — happy to revisit if you have a concrete repro that gets past the parser's normalization.

src/interpreter/plugin/SortPlugin.ts
```diff
@@ -0,0 +1,144 @@
...
+      method: 'sort',
+      sizeOfResultArrayMethod: 'sortArraySize',
+      parameters: [
+        {argumentType: FunctionArgumentType.RANGE},
+        {argumentType: FunctionArgumentType.NUMBER, defaultValue: 1, emptyAsDefault: true},
+        {argumentType: FunctionArgumentType.NUMBER, defaultValue: 1, emptyAsDefault: true},
+        {argumentType: FunctionArgumentType.BOOLEAN, defaultValue: false, emptyAsDefault: true},
+      ],
+      vectorizationForbidden: true,
+    },
```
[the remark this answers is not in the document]
marcin-kordas-hoc: Fixed in f2f43fe. Confirmed: without the flag `=SORT(A1:A3*2)` computed only the first cell. Added `enableArrayArithmeticForArguments` (matching FILTER/VSTACK/HSTACK), so the full computed array now sorts. Same flag applied to UNIQUE in #1708. Regression tests in the paired tests PR.

src/interpreter/plugin/SortPlugin.ts
```diff
@@ -0,0 +1,153 @@
...
+      return ArraySize.error()
+    }
+    const metadata = this.metadata('SORT')
+    const subChecks = ast.args.map((arg) =>
+      this.arraySizeForAst(arg, new InterpreterState(state.formulaAddress, state.arraysFlag || (metadata?.enableArrayArithmeticForArguments ?? false))))
+    // Return a fresh ArraySize (isRef defaults to false). Propagating the input's
+    // ArraySize verbatim would carry its `isRef` flag, and ArraySize.isScalar()
+    // treats a ref as scalar — which would collapse the spilled result into a
+    // single cell. SORT's output always matches the input shape.
+    return new ArraySize(subChecks[0].width, subChecks[0].height)
```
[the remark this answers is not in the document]
marcin-kordas-hoc: Verified this does not reproduce as a break. `sortArraySize` is the same shape as the shipped `filterArraySize`/`arrayconstrainArraySize` (fresh `ArraySize` from `subChecks`, `isRef` dropped) — the established pattern for every array function, not introduced here.

Empirically, a whole-column SORT at a non-A1 position returns a graceful `#ERROR!` "Invalid range size" (a parse-time guard), not a crash or a broken spill:
- `=SORT(D:D)` at a non-zero row → `#ERROR!` Invalid range size
- `=FILTER(D:D, D:D>1)` at the same cell → identical `#ERROR!` Invalid range size

The runtime empty-guard targets a different case — a whole-column ref to an *empty existing sheet*, which resolves to a finite 0-row range and returns `#N/A`. Unbounded ranges in array functions are a pre-existing HF-wide limitation; special-casing SORT would make it inconsistent with its siblings. No change here.

docs/adr/2026-07-13-sort-unique-array-functions.md
sequba, line 1: Do you think we should keep ADRs in repo? If so, I'd rather not have them in `docs` dir.

sequba: Personally, I think we can keep some files describing the general architectural decisions, but ADR about a specific function implementation won't be useful in implementing future features so it should be removed. If there are any non-obvious decisions or implementation details, they should be noted in the code comments.

sequba: Ok, let's keep it but move it to the hyperformula-tests repo please (`dev_docs` directory)

marcin-kordas-hoc: Done ✅ Moved it to `hyperformula-tests` under `dev_docs/2026-07-13-sort-unique-array-functions.md` and dropped it from the engine repo here. Applied to both PRs — paired tests PRs: SORT handsontable/hyperformula-tests#24, UNIQUE handsontable/hyperformula-tests#25 (paired with #1708).

sequba (approved):

PR #1722, opened 2026-08-05
Feature/take function

Tobiadefami:
### Context

This PR adds the `TAKE` dynamic-array function. `TAKE` returns rows or columns from the beginning or end of an array and supports positive and negative counts, optional columns, syntactically empty argument slots, and array spilling.

A row or column count that truncates to zero returns the existing `#N/A` error with a TAKE-specific message. This is a documented Excel difference because HyperFormula does not expose a `#CALC!` error type. Omitting the required `rows` argument continues to return the existing wrong-argument `#N/A` error.

### Implementation

- Counts are truncated and clamped to the source dimensions; negative counts select from the end.
- Address-backed sources remain lazy through sub-range spans.
- Static spill-size prediction evaluates dependency-free literal percentages and arithmetic, so invalid counts and literal errors are resolved before spill allocation.
- Whole-column sources spill from row 1 and return `#SPILL!` below row 1; finite counts are capped to configured sheet limits before allocation.
- Function metadata, translations, changelog, and compatibility documentation are included.
- The VSTACK/HSTACK localization fix now lives in handsontable/hyperformula#1748 and is not part of this PR.

### Validation

- 86/86 focused `TAKE` tests passed.
- 503/503 Jest suites passed (6,319 tests passed, 3 skipped).
- 152/152 relevant metadata, localization, and optional-parameter tests passed.
- The focused coverage run exercises invalid percentage sizing and empty unbounded source sizing before spill allocation.
- TypeScript compilation, targeted lint, and `git diff --check` passed.
- Excel Online confirmed same-sheet and cross-sheet whole-column behavior and the row-1 spill boundary.

### Types of changes

- [ ] Breaking change
- [ ] Bug fix
- [x] Additional language file or translation change
- [x] Documentation change

### Checklist

- [x] The code follows the HyperFormula contribution guidelines and project style.
- [x] I have signed the Contributor License Agreement.
- [ ] The change is compliant with OpenDocument 1.3.
- [x] The change is compatible with Microsoft Excel, subject to documented differences.
- [ ] The change is compatible with Google Sheets.
- [x] The change is described in `CHANGELOG.md`.
- [x] Documentation has been updated.
- [ ] A migration guide is required.

---

> [!NOTE]
> **Medium Risk**
> New dynamic-array spill logic and size prediction touch array evaluation and sheet limits; behavior is well-tested but mistakes could cause incorrect spills or performance issues on large column references.
> 
> **Overview**
> Adds the Excel-style **`TAKE`** dynamic-array function so formulas can return a sub-rectangle of a range or array from the start or end, with optional row/column counts, empty slots meaning “keep all,” and negative counts selecting from the tail.
> 
> Implementation lives in **`ArrayPlugin`**: counts are truncated and capped to the source size; zero row/column counts yield **`#N/A`** with a new **`ZeroRowOrColumnCount`** message (documented as differing from Excel’s **`#CALC!`**). Address-backed sources stay lazy via **`AbsoluteCellRange`** sub-spans. **`takeArraySize`** statically resolves literal count expressions (including simple arithmetic and **`TRUE`/`FALSE`**) for spill sizing; whole-column/row unbounded cases respect sheet limits and row‑1/column‑1 spill boundaries. **`Interpreter.isFunctionImplementedBy`** supports that static analysis.
> 
> Also updates changelog, built-in function metadata, compatibility notes for **`TAKE`**, and localized names across language packs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8d9e74e808aafc6e8b38ad4019d4e04825ac4630. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Added
 
+- Added the `TAKE` dynamic-array function and the `CALC` error type used for empty-array results.
```
sequba: Link this pull request in the changelog entry

sequba (changes requested):

src/interpreter/plugin/ArrayPlugin.ts
```diff
@@ -166,6 +177,67 @@ export class ArrayPlugin extends FunctionPlugin implements FunctionPluginTypeche
...
+      return ArraySize.error()
+    }
+
+    const metadata = this.metadata('TAKE')
+    const sourceSize = this.arraySizeForAst(
+      ast.args[0],
+      new InterpreterState(state.formulaAddress, state.arraysFlag || (metadata?.enableArrayArithmeticForArguments ?? false)),
+    )
+
+    return new ArraySize(sourceSize.width, sourceSize.height)
```
sequba: `takeArraySize` always returns the full source dimensions instead of the requested result size, so the engine reserves the source's whole footprint and reports a false `#SPILL!` for any content that TAKE's real result never touches.

Verified by running the engine:

```js
Sheet1 = [['=TAKE(Data!A1:C3,1)'], [null], ['untouched']]   // Data is 3x3

// A1 -> {"value":"#SPILL!","type":"SPILL","message":"No space for array result."}
```

The real 1x3 result never reaches row 3. The control `=ARRAY_CONSTRAIN(Data!A1:C3,1,3)` in the identical layout correctly returns `1`.

The same false collision happens on the width axis — `=TAKE(Data!A1:C3,3,1)` false-spills on content in B1/C1, so two TAKE formulas placed side by side collide with each other. Also `getSheetDimensions` reports height 3 for a 1-row result, and `isCellPartOfArray(C3)` is `true`.

`arrayconstrainArraySize` (lines 105-124 of this file) already solves exactly this by inspecting literal `AstNodeType.NUMBER` args:

```ts
if (ast.args[1].type === AstNodeType.NUMBER) { height = Math.min(height, ast.args[1].value) }
```

`SequencePlugin`'s `parseLiteralDimension` does the same.

### On the parse-time constraint

This can't be fixed in full, and that's worth stating explicitly: HyperFormula predicts array sizes statically, before evaluation, so whenever `rows`/`columns` come from a cell reference or a nested formula the result size genuinely isn't knowable at prediction time and an upper bound is the only option. That's the same unavoidable situation FILTER and UNIQUE are in, where the size is data-dependent by nature.

TAKE differs from those in one respect: its result size *is* statically known whenever the counts are numeric literals — the common case, and the form used in TAKE's own documented examples (`=TAKE(A1:C5, 2)`). Tightening the prediction for literal counts, the way `ARRAY_CONSTRAIN` and `SEQUENCE` already do, would remove the false spill for that majority case and leave the upper-bound reservation only where it is genuinely unavoidable.

src/interpreter/plugin/ArrayPlugin.ts
```diff
@@ -166,6 +177,67 @@ export class ArrayPlugin extends FunctionPlugin implements FunctionPluginTypeche
...
+
+    return this.runFunction(ast.args, state, this.metadata('TAKE'),
+      (range: SimpleRangeValue, rows: number, columns: number | undefined) => {
+        const sourceHeight = range.height()
+        const sourceWidth = range.width()
+        const requestedRows = rowsIsEmpty ? sourceHeight : Math.trunc(rows)
+        const requestedColumns = columnsIsMissingOrEmpty || columns === undefined ? sourceWidth : Math.trunc(columns)
+
+        if (requestedRows === 0 || requestedColumns === 0 || sourceHeight === 0 || sourceWidth === 0) {
+          return new CellError(ErrorType.CALC, ErrorMessage.EmptyRange)
```
sequba: The zero-count branch returns `ErrorMessage.EmptyRange` ("Empty range not allowed."), which is factually wrong when the source range is fully populated and only the requested count is zero — and the new spec locks that wrong text in.

```js
=TAKE(A1:C3, 0)   // over a fully populated 3x3 range

// -> {type: 'CALC', value: '#CALC!', message: 'Empty range not allowed.'}
```

This is user-visible through `DetailedCellError.message`, which is what `getCellValue()` returns. The range is not empty; the count argument is.

The condition on line 203 conflates two distinct situations under one message that only fits the second:

- `requestedRows === 0 || requestedColumns === 0` — a zero count against a perfectly good range
- `sourceHeight === 0 || sourceWidth === 0` — a genuinely empty source

`function-take.spec.ts:110/120/130/141/152` assert the wrong wording, so it is pinned in place by the tests.

`ErrorMessage.LessThanOne` ("Argument cannot be less than 1.") is closer but also imprecise, since negative counts are legal for TAKE. A new message along the lines of "Row or column count cannot be zero." would fit, keeping `EmptyRange` for the genuinely-empty-source branch.

src/interpreter/plugin/ArrayPlugin.ts
```diff
@@ -166,6 +177,67 @@ export class ArrayPlugin extends FunctionPlugin implements FunctionPluginTypeche
...
+   * evaluate to zero return a #CALC! error.
+   *
+   * @param ast
+   * @param state
+   */
+  public take(ast: ProcedureAst, state: InterpreterState): InterpreterValue {
+    const rowsArg = ast.args[1]
+    const columnsArg = ast.args[2]
+    const rowsIsEmpty = rowsArg?.type === AstNodeType.EMPTY
+    const columnsIsMissingOrEmpty = columnsArg === undefined || columnsArg.type === AstNodeType.EMPTY
```
sequba: `take` hand-inspects `ast.args` for `AstNodeType.EMPTY` to implement "empty argument keeps everything", bypassing the argument-metadata mechanism the framework already provides and that every comparable function in the codebase uses.

These two lines read the AST directly, then lines 200-201 branch on the captured flags inside the callback.

`FunctionPlugin.ts:236-249` documents `emptyAsDefault` as: "an empty argument is treated as if the argument was not provided at all" — that is, it falls back to `defaultValue`. It is used by `SequencePlugin.ts:72-74`, `SortPlugin.ts:33-35`, `UniquePlugin.ts:32-33`, and — the exact `optionalArg` + `emptyAsDefault` combination TAKE needs — `AddressPlugin.ts:29-30`.

The duplication is literal. `FunctionPlugin.ts:302` already computes:

```ts
const isSyntacticallyEmpty = argAst.type === AstNodeType.EMPTY
```

and threads it through as `syntacticallyEmptyFlags` into `coerceArgumentsToRequiredTypes`, which is the same expression these two lines recompute by hand.

Declaring `rows` as `{defaultValue: Number.POSITIVE_INFINITY, emptyAsDefault: true}` (and likewise for `columns`) removes both flags and both branches, since the existing `Math.min(Math.abs(...), sourceHeight)` clamp already collapses `Infinity` down to the source dimension. `coerceToType` applies `minValue`/`maxValue` only when they are declared, and TAKE declares neither, so `Infinity` passes through the coercion layer untouched.

Cost: two mechanisms now express "empty means keep the default" in one class, against DEV_DOCS.md:94 — "Avoid duplication. Extract shared logic instead of copying it, and reuse the existing helpers and abstractions of the codebase."

src/interpreter/plugin/ArrayPlugin.ts
```diff
@@ -166,6 +177,67 @@ export class ArrayPlugin extends FunctionPlugin implements FunctionPluginTypeche
...
+          return new CellError(ErrorType.CALC, ErrorMessage.EmptyRange)
+        }
+
+        const rowsToTake = Math.min(Math.abs(requestedRows), sourceHeight)
+        const columnsToTake = Math.min(Math.abs(requestedColumns), sourceWidth)
+        const startRow = requestedRows > 0 ? 0 : sourceHeight - rowsToTake
+        const startColumn = requestedColumns > 0 ? 0 : sourceWidth - columnsToTake
+        const result = range.data
+          .slice(startRow, startRow + rowsToTake)
+          .map(row => row.slice(startColumn, startColumn + columnsToTake))
```
sequba, line 337: `range.data` forces materialization of the entire source range before the slices discard almost all of it, making the cost O(source area) rather than O(result area).

`=TAKE(A1:C100000, 2)` reads and boxes all 300,000 cells to keep 6: `range.data` calls `SimpleRangeValue.ensureThatComputed` (`SimpleRangeValue.ts:211-224`), which runs `addressesArrayMap` over the whole range, and only then do these lines slice.

A narrower path already exists in this codebase for the same shape of problem. `LookupPlugin`'s `doVlookup`/`doHlookup` narrow first:

```ts
searchedRange = SimpleRangeValue.onlyRange(AbsoluteCellRange.spanFrom(range.start, 1, range.height()), this.dependencyGraph)
```

(`LookupPlugin.ts:219` and `:249`). `SimpleRangeValue.onlyRange` (`SimpleRangeValue.ts:74-76`) constructs with `_data = undefined`, so materialization is deferred to just that sub-range. When `range.range !== undefined`, TAKE could span the target sub-range directly instead of slicing an already-materialized 2D array; the computed-array case (`=TAKE(SEQUENCE(3,3),2,2)`, where `range.range` is undefined) would still need the present path.

Two caveats on the suggestion, so it is not taken as drop-in:

- LookupPlugin uses `onlyRange` for an *intermediate* it searches, whereas TAKE would be returning it as the formula's result. That result is still materialized downstream by `ArrayValue.fromInterpreterValue` — but over the sub-range only, which is the intended win. Worth confirming the returned-value path behaves.
- For negative counts the span has to start from an offset address rather than `range.start`.

`ARRAY_CONSTRAIN` carries the same cost, so this is not a regression introduced here. But DEV_DOCS.md:103-104 asks to "Consider the computational complexity of every change" and to "Pick the best complexity that still keeps the code readable. When a faster algorithm is harder to follow, explain the trade-off in a JSDoc comment." `take`'s JSDoc has neither the tighter approach nor a note on why it was not taken.

src/Cell.ts
```diff
@@ -23,6 +23,9 @@ import {AddressWithSheet} from './parser/Address'
  * Possible errors returned by our interpreter.
  */
 export enum ErrorType {
+  /** Calculation error. */
+  CALC = 'CALC',
```
sequba: The hand-maintained error reference table was not updated with this new error type, which DEV_DOCS.md's definition of done requires.

`docs/guide/types-of-errors.md` lists exactly 9 rows — `#DIV/0!`, `#N/A`, `#NAME?`, `#NUM!`, `#REF!`, `#VALUE!`, `#CYCLE!`, `#ERROR!`, `#LIC!` — and stops at line 26. `grep -rn CALC docs/guide/*.md` matches only the unrelated `SEQUENCE(0)` row in `list-of-differences.md`.

DEV_DOCS.md:74 lists "Updates to documentation related to the change" among the definition-of-done items, and DOCS_CONTENT_GUIDE.md:5-6 states the core principle: "Every page and every section is retrieved and read in isolation, out of order, by a reader with zero prior context."

A user who sees `#CALC!` from TAKE and opens the page that exists to explain error values finds nothing.

CHANGELOG.md
Tobiadefami: Done. The changelog entry now links to #1722.

src/interpreter/plugin/ArrayPlugin.ts
Tobiadefami: Fixed. A zero row or column count now returns #CALC! with “Row or column count cannot be zero.” 

EmptyRange is kept for a genuinely zero-dimensional source.

src/interpreter/plugin/ArrayPlugin.ts
Tobiadefami: Fixed. TAKE no longer checks EMPTY AST nodes itself. Both count arguments now use `defaultValue: Infinity` with `emptyAsDefault: true`.

This also required separating defaults from optionality. The rows argument needs a default for `TAKE(range,,2)`, but it must remain required for `TAKE(range)`. 

An explicit `optionalArg: false` now takes precedence over the presence of a default.

src/Cell.ts
Tobiadefami: Done. #CALC! is now included in docs/guide/types-of-errors.md.

src/interpreter/plugin/ArrayPlugin.ts
Tobiadefami: Fixed. Address-backed inputs now return a narrowed onlyRange covering the TAKE result. 

Computed arrays keep the existing value-slicing path. 

Negative row and column offsets are handled independently.

src/interpreter/plugin/ArrayPlugin.ts
Tobiadefami: Fixed. takeArraySize() now uses literal row and column counts, including negative literals, to predict the result footprint. 

Dynamic counts still use the source size as an upper bound.

src/interpreter/plugin/ArrayPlugin.ts
```diff
@@ -166,6 +199,87 @@ export class ArrayPlugin extends FunctionPlugin implements FunctionPluginTypeche
...
+        const sourceRange = range.range
+
+        // Keep address-backed ranges lazy to avoid materializing cells outside the TAKE result.
+        if (sourceRange !== undefined) {
+          const resultRange = AbsoluteCellRange.spanFrom(
+            sourceRange.getAddress(startColumn, startRow),
+            columnsToTake,
+            rowsToTake,
+          )
+          return SimpleRangeValue.onlyRange(resultRange, this.dependencyGraph)
```
sequba: Returning a range-backed value here changes the result's *identity*, not just how it is read. `SimpleRangeValue.onlyRange` leaves `_data` undefined and sets `range`, so `isAdHoc()` becomes `false` — and `coerceRangeToScalar` (`src/interpreter/ArithmeticHelper.ts:813-832`) branches on exactly that flag:

```ts
if (arg.isAdHoc()) {
  return arg.data[0]?.[0]
}
const range = arg.range!
if (state.formulaAddress.sheet === range.sheet) {
  if (range.width() === 1) {
    const offset = state.formulaAddress.row - range.start.row
```

So a TAKE result in scalar context no longer yields its first element; it goes through Excel-style implicit intersection against the **source** address. Every other array function returns `onlyValues` and keeps the first-element behaviour.

Verified by running the engine at this commit and at `b509e68c7~1`, default config (`useArrayArithmetic: false`), `Data!A1:C3 = 1..9`:

| formula | this commit | `b509e68c7~1` |
| --- | --- | --- |
| `=TAKE(Data!A1:C3,2)+0` | `#VALUE!` "Cell range not allowed." | `1` |
| `=ABS(TAKE(Data!A1:C3,2))` | `#VALUE!` | `1` |
| `=TAKE(Data!A1:C3,2)&""` | `#VALUE!` | `"1"` |

The controls in the same sheet are unaffected — `=SORT(Data!A1:C3)+0`, `=VSTACK(Data!A1:C3)+0` and `=ARRAY_CONSTRAIN(Data!A1:C3,2,3)+0` all return `1` on both revisions. TAKE now behaves like a bare range reference (`=Data!A1:C3+0` → `#VALUE!`).

Cross-sheet is broken unconditionally, because the `state.formulaAddress.sheet === range.sheet` guard fails and the function falls through to `return undefined`:

```
Sheet1!A1:A2 = =ABS(TAKE(Sheet2!A1:A5,3))
  this commit -> #VALUE!, #VALUE!
  b509e68c7~1 -> 1, 1
```

When source and formula share a sheet and the source is 1-D, the intersection *succeeds* — so there is no error, just the wrong number, and which number depends on where the formula sits. With `A1:A4 = 10,20,30,40`:

```
B1:B4 = =ROUND(TAKE($A$1:$A$4,2),0)
  this commit -> 10, 20, #VALUE!, #VALUE!
  b509e68c7~1 -> 10, 10, 10, 10
```

One more consequence: the behaviour now depends on the *shape of the input* rather than on what TAKE does, because a computed source has `range === undefined` and falls through to the `onlyValues` path on line 248. `=ROUND(TAKE(VSTACK($A$1:$A$4),2),0)` returns `10` in every row — wrapping the source in a no-op changes the answer.

This is the caveat from the earlier thread on this line ("Worth confirming the returned-value path behaves") coming due. The `LookupPlugin` precedent does not carry over: those `onlyRange` values are consumed inside `doVlookup`/`doHlookup` and never escape, so the flag never reaches a coercion site. TAKE returns one as the formula's value.

Nothing in `function-take.spec.ts` uses TAKE as an argument to anything, so all 35 cases stay green.

If the laziness is worth keeping, the narrowing needs to happen without changing the result's identity — read only the sub-range's cells and return `SimpleRangeValue.onlyValues`, or let `SimpleRangeValue` carry narrowed provenance separately from `isAdHoc()`. A `subRange(topRow, leftCol, height, width)` method on `SimpleRangeValue` would keep that decision behind the class boundary and let `ARRAY_CONSTRAIN` reuse it.

src/interpreter/plugin/ArrayPlugin.ts
```diff
@@ -166,6 +199,87 @@ export class ArrayPlugin extends FunctionPlugin implements FunctionPluginTypeche
...
+      ast.args[0],
+      new InterpreterState(state.formulaAddress, state.arraysFlag || (metadata?.enableArrayArithmeticForArguments ?? false)),
+    )
+
+    const literalRows = ArrayPlugin.parseTakeLiteralDimension(ast.args[1])
+    const literalColumns = ArrayPlugin.parseTakeLiteralDimension(ast.args[2])
+    const height = literalRows === undefined ? sourceSize.height : Math.min(sourceSize.height, literalRows)
+    const width = literalColumns === undefined ? sourceSize.width : Math.min(sourceSize.width, literalColumns)
+
+    if (height < 1 || width < 1) {
```
sequba: This guard is missing the half that catches non-finite dimensions, so a whole-column or whole-row source silently loses data.

`ArraySizePredictor.checkArraySizeForAst` returns `new ArraySize(range.width(), range.height(), true)` for `COLUMN_RANGE`/`ROW_RANGE` (`src/ArraySize.ts:58-65`), and those ranges are constructed with an infinite end — `AbsoluteColumnRange` sets `end.row = Number.POSITIVE_INFINITY` (`src/AbsoluteCellRange.ts:445`), `AbsoluteRowRange` sets `end.col = Number.POSITIVE_INFINITY` (`:499`). When the matching count is not a bare numeric literal, lines 273-274 pass that `Infinity` straight through, `Infinity < 1` is `false`, and an `ArraySize` with an infinite axis reaches the vertex.

`arrayconstrainArraySize`, 134 lines above in this same file, has the clause that catches it:

```ts
if (height < 1 || width < 1 || !Number.isInteger(height) || !Number.isInteger(width)) {
  return ArraySize.error()
}
```

Verified by running the engine at this commit, `Data!A1:C3 = 1..9`:

| formula | result | expected |
| --- | --- | --- |
| `=TAKE(Data!1:3,2)` | `[[1],[4]]`, dims `1x2` | `[[1,2,3],[4,5,6]]` |
| `=TAKE(Data!A:C,,2)` | `[[1,2]]`, dims `2x1` | `[[1,2],[4,5],[7,8]]` |
| `=ARRAY_CONSTRAIN(Data!1:3,2,3)` | `[[1,2,3],[4,5,6]]` | correct |

No error is raised — `B1`, `C1`, `B2`, `C2` are simply `null`. The values TAKE computes are right and only the reservation is wrong: in the same sheet `=COLUMNS(TAKE(Data!1:3,2))` returns `3` and `=SUM(TAKE(Data!1:3,2))` returns `21`, while the spilled area is one column wide.

The same formulas placed anywhere other than row 1 / column A fail differently, because `ArrayFormulaVertex.getRange()` → `AbsoluteCellRange.spanFromOrUndef` (`src/AbsoluteCellRange.ts:117-126`) returns `undefined` for an infinite span not anchored at index 0, and `isThereSpaceForArray` then reports no space:

```
Sheet1!B1 = =TAKE(Data!1:3,2)   ->  #SPILL! "No space for array result."
```

There is a second, subtler case on the same lines even *with* a literal count, because the prediction uses the range's declared height while the runtime uses `effectiveHeight`:

```
Data = [[1,2,3]]                 // one row
Sheet1!A1 = =TAKE(Data!A:C,3)    // min(Infinity, 3) = 3 predicted, effectiveHeight = 1
```
The values are correct (`1,2,3` in row 1), but `A2:C3` are reserved as blanks and `isCellPartOfArray(A3)` is `true`; writing anything there turns the formula into `#SPILL!`. Excel returns a 1x3 result. Clamping the predicted size to the effective dimensions — or reusing `ARRAY_CONSTRAIN`'s `Number.isInteger` guard so an unbounded source degrades to `ArraySize.error()` the way it already does for that function — covers both.

No test in `function-take.spec.ts` uses a whole-column or whole-row source, which is why this is green today.

src/interpreter/plugin/ArrayPlugin.ts
```diff
@@ -166,6 +199,87 @@ export class ArrayPlugin extends FunctionPlugin implements FunctionPluginTypeche
...
+      return ArraySize.error()
+    }
+
+    const metadata = this.metadata('TAKE')
+    const sourceSize = this.arraySizeForAst(
+      ast.args[0],
+      new InterpreterState(state.formulaAddress, state.arraysFlag || (metadata?.enableArrayArithmeticForArguments ?? false)),
+    )
+
+    const literalRows = ArrayPlugin.parseTakeLiteralDimension(ast.args[1])
```
sequba: `parseTakeLiteralDimension` matches only `AstNodeType.NUMBER` and a unary `+`/`-` wrapping a `NUMBER`, so every other way of writing a constant count returns `undefined` and lines 273-274 fall back to the full source size. The reservation is then decided by the *spelling* of the count rather than by its value.

Verified at this commit, `Data` 3x3, `Sheet1 = [['=TAKE(Data!A1:C3,<count>)'], [null], ['neighbour']]` — every one of these results is a single row that fits in `A1:C1`:

| `<count>` | result |
| --- | --- |
| `1` | `1, 2, 3` — correct |
| `"1"` | `#SPILL!` |
| `TRUE()` | `#SPILL!` |
| `(1)` | `#SPILL!` |
| `0+1` | `#SPILL!` |
| `2-1` | `#SPILL!` |
| `200%` | `#SPILL!` |
| `--1` | `#SPILL!` |
| `Counts!A1` (= 1) | `#SPILL!` |

Even with nothing in the way the surplus is reserved: `isCellPartOfArray(A3)` is `true`, and a later `setCellContents(A3, 'hello')` flips the formula to `#SPILL!`.

Two of these are inputs the spec explicitly supports — `'coerces numeric text to a count'` (`function-take.spec.ts:293`) and `'coerces TRUE to one'` (`:300`) — and both pass only because they run on an otherwise empty sheet. Note `ArraySizePredictor.checkArraySizeForAst` unwraps `PARENTHESIS` (`src/ArraySize.ts:115`) and `runFunctionWithReferenceArgument` unwraps it in a `while` loop, so skipping it here is not a house convention.

The cell-reference case is genuinely unfixable at parse time and an upper bound is the only option there — same situation FILTER and UNIQUE are in. But the constant forms above are all statically known, and the helper this one was modelled on already handles most of them. `SequencePlugin.parseLiteralDimension` (`src/interpreter/plugin/SequencePlugin.ts:41-61`) covers `NUMBER`, `STRING`, both unary ops, and zero-arg `TRUE()`/`FALSE()`:

```ts
if (node.type === AstNodeType.STRING) {
  const parsed = Number(node.value)
  return Number.isFinite(parsed) ? Math.trunc(parsed) : undefined
}
```

So `=SEQUENCE("2")` predicts its size and `=TAKE(A1:C3,"2")` does not — an inconsistency between two functions in the same category with no reason behind it.

This is now the fourth hand-rolled copy of "extract a static number from an AST node": here, `SequencePlugin.parseLiteralDimension`, `arrayconstrainArraySize` (line 136 of this file, `NUMBER` only), and `FormulaParser.handleOffsetHeuristic` (`src/parser/FormulaParser.ts:766-790`, inlined four times). They disagree on capability, and the next one lands with DROP/CHOOSEROWS/EXPAND. Extracting a single `signedNumberLiteralValue(ast): number | undefined` into `src/parser/Ast.ts` — which already owns `AstNodeType` and the unary-op builders — and calling it from all four would fix this class in one place; the `Math.abs` stays at TAKE's call site since only TAKE wants the magnitude.

Worth pinning whatever the final boundary is with tests, since none of the rows in the table above is covered today.

src/Cell.ts
sequba: We need to take `CALC` out of this PR. Adding a member to `ErrorType` is a breaking change to a public API, and we don't want to cut a major release now.

### Why it's breaking

`TranslationPackage`'s constructor validates error and UI keys **exhaustively** against the enum:

```ts
// src/i18n/TranslationPackage.ts
private checkErrors(): void {
  for (const key of Object.values(ErrorType)) {
    if (!(key in this.errors) && (key !== ErrorType.LIC)) {
      throw new MissingTranslationError(`errors.${key}`)
    }
  }
}
```

So any third-party language pack written against 3.3.0 now fails at registration, before a single formula is evaluated:

```
HyperFormula.registerLanguage('myPack', packBuiltAgainst_3_3_0)
// MissingTranslationError: Translation for errors.CALC is missing in the translation package you're using.
```

`RawTranslationPackage` is public API and custom packs are a documented feature — `docs/guide/localizing-functions.md` even walks users through building one, and the two example packs on that page list the error keys explicitly, so they throw as of this branch.

Worth being precise about the asymmetry, because it is not obvious: adding a **function** is not breaking in the same way. `checkFunctionTranslations` doesn't iterate the registry at all — it only rejects packs that override protected names — so a pack that has never heard of `TAKE` registers fine and `=TAKE(...)` simply resolves to `#NAME?` in that language (`FunctionRegistry.getFunction` gates on `isFunctionTranslated`). Verified against this branch, each pack derived from `enGB` with one key removed:

| pack | `registerLanguage` |
| --- | --- |
| missing `functions.TAKE` | OK — `=TAKE(...)` → `#NAME?` |
| missing `functions.SORT` | OK — `=SORT(...)` → `#NAME?` |
| `functions: {}` (empty) | OK |
| missing `errors.CALC` | **throws** |
| missing `ui.NEW_SHEET_PREFIX` | **throws** |

So the `TAKE` half of this PR is genuinely additive; the `CALC` half breaks packs on its own, even for users who never call the function. For precedent: `#SPILL!`, the last translatable error type we added, shipped in **1.0.0**.

### To be clear — `#CALC!` is the right answer

Microsoft is explicit: *"Excel returns a #CALC! error to indicate an empty array when either rows or columns is 0."* This isn't a case where we should pick something else on the merits. It's purely a release-timing constraint.

### Interim solution

Use `ErrorType.NA` and keep the new message:

```ts
return new CellError(ErrorType.NA, ErrorMessage.ZeroRowOrColumnCount)
```

`#N/A` is the closest fit among the existing types because it is already what this codebase returns for an empty array result — `FILTER` does exactly that 37 lines above in this same file:

```ts
return new CellError(ErrorType.NA, ErrorMessage.EmptyRange)
```

and `UNIQUE` and `SORT` agree. Using `#N/A` keeps TAKE consistent with its own family instead of introducing a third convention (`SEQUENCE(0)` returns `#VALUE!`, which is the other candidate but the odd one out). Keeping `ErrorMessage.ZeroRowOrColumnCount` preserves the accuracy fix from the earlier round — the message stays correct about *what* was zero, only the error type changes.

That leaves us with one migration later rather than two, since we'll likely want to move `FILTER`/`UNIQUE`/`SORT`/`SEQUENCE` to `#CALC!` at the same time.

### What to revert here

- `CALC = 'CALC'` in `src/Cell.ts`
- the `CALC` entry in all 17 language packs (this also makes the "every pack ships the untranslated English `#CALC!`" problem moot for now — when we do add it, the packs need real localized names; Polish Excel uses `#OBL!`)
- the `#CALC!` row in `docs/guide/types-of-errors.md`
- the `CALC` half of the CHANGELOG entry

Keep `ErrorMessage.ZeroRowOrColumnCount`, and record the deviation in `docs/guide/list-of-differences.md` — there's already a row of exactly this shape for `=SEQUENCE(0)` (`VALUE` / `N/A` / `CALC`).

Tracked for the next major in **[HF-350](https://app.clickup.com/t/86cb3cvah)** (tagged `breaking change`), which has the full scope and the reasoning above. We'll do it there.

src/interpreter/plugin/FunctionPlugin.ts
```diff
@@ -249,6 +253,22 @@ export interface FunctionArgument {
...
+/**
+ * Returns whether a function argument may be omitted.
+ *
+ * An explicit `optionalArg` declaration takes precedence. Otherwise, the
+ * presence of a default value preserves the existing implicit-optional
+ * behavior.
+ *
+ * @param {FunctionArgument | undefined} argument - The argument metadata to inspect.
+ */
+export function isFunctionArgumentOptional(argument: FunctionArgument | undefined): boolean {
```
sequba, line 264: Same situation as `CALC`: the problem is real and the approach is sound, but it's breaking and we can't afford a major right now — so let's solve it locally inside TAKE instead of changing the shared model.

### The problem you hit is real

TAKE genuinely needs something the argument-metadata model cannot express: `rows` must accept a **syntactically empty slot** (`=TAKE(range,,2)` → keep all rows) while still rejecting an **omitted** argument (`=TAKE(range)` → error, since Excel marks `rows` Required). `emptyAsDefault` is inert without a `defaultValue`, and a `defaultValue` used to imply optional — so "required, but empty allowed" had no expression.

It's also worth saying why nothing else in the codebase hit this: TAKE is the only function where an `emptyAsDefault` argument is *positionally required*. Every other user of the flag puts it on a genuinely optional trailing argument, where "omitted" and "empty" should mean the same thing:

| function | parameters with `emptyAsDefault` | shape |
| --- | --- | --- |
| `ADDRESS` | `abs_number`, `use_a1` | trailing, `optionalArg: true` |
| `SEQUENCE` | `columns`, `start`, `step` | trailing, optional |
| `SORT` | `sort_index`, `sort_order`, `by_col` | trailing, optional |
| `UNIQUE` | `by_col`, `exactly_once` | trailing, optional |
| **`TAKE`** | **`rows`** | **middle, required** |

So the gap is genuine, and separating defaults from optionality is the right long-term model.

### Why we can't ship it now

`isFunctionArgumentOptional` changes the meaning of an existing combination on a public interface. The old predicate was `optionalArg || defaultValue !== undefined`; the new one returns `optionalArg` verbatim when it is set, so `{optionalArg: false, defaultValue: X}` flips from optional to required.

`FunctionArgument` is exported from `src/index.ts` and custom function plugins are a documented feature. Verified by registering a plugin whose 2nd parameter is `{argumentType: NUMBER, optionalArg: false, defaultValue: 42}`:

```
=OPTFN(1)   ->  43        on ebaa2b28a
=OPTFN(1)   ->  #N/A "Wrong number of arguments."   on this branch
```

`getFunctionDetails('OPTFN').parameters[1].optional` also flips `true` → `false`, so metadata-driven autocomplete starts advertising the wrong arity.

Two things make it worse than a typical semantic tweak. First, no built-in uses that combination except TAKE itself, so the entire blast radius is user code and nothing in-tree would ever catch a regression. Second, `docs/guide/custom-functions.md` actively prescribes the combination — its `MY_FUNCTION` example pairs `defaultValue: 10` with `optionalArg: false`, and the page still states the old rules ("Setting a `defaultValue` for an argument always makes that argument optional"). A plugin copy-pasted from our own published guide changes arity on upgrade.

### Interim solution: check the arity inside TAKE

Revert `isFunctionArgumentOptional` to the previous predicate, drop `optionalArg: false` from the `rows` parameter, and enforce the requirement where it belongs — in the function that has the requirement:

```ts
public take(ast: ProcedureAst, state: InterpreterState): InterpreterValue {
  // `rows` carries a default so that a syntactically empty slot keeps every row, which also makes
  // the argument metadata treat it as omittable. TAKE requires it, so the arity is checked here.
  if (ast.args.length < 2) {
    return new CellError(ErrorType.NA, ErrorMessage.WrongArgNumber)
  }

  return this.runFunction(ast.args, state, this.metadata('TAKE'), ...)
}
```

`runFunction` still rejects the 4-argument call on its own, and `takeArraySize` already guards `ast.args.length < 2 || > 3`, so this is the only gap to close.

I built it and ran it. Every user-visible TAKE behaviour is identical to this branch:

| formula | this branch | with the local check |
| --- | --- | --- |
| `=TAKE(A1:C3)` | `#N/A` | `#N/A` |
| `=TAKE(A1:C3,,2)` | `[[1,2],[4,5],[7,8]]` | same |
| `=TAKE(A1:C3,2)` | `[[1,2,3],[4,5,6]]` | same |
| `=TAKE(A1:C3,)` / `=TAKE(A1:C3,,)` | full source | same |
| `=TAKE(A1:C3,2,)` | `[[1,2,3],[4,5,6]]` | same |
| `=TAKE(A1:C3,-2,-2)` | `[[5,6],[8,9]]` | same |
| `=TAKE(A1:C3,0,2)` | zero-count error | same |
| `=TAKE(A1:C3,1,1,1)` | `#N/A` | same |
| `=OPTFN(1)` (third-party plugin above) | `#N/A` | **`43` — restored** |

All 35 cases in `function-take.spec.ts` pass unchanged.

### The cost, stated honestly

`getFunctionDetails('TAKE').parameters[1].optional` becomes `true` instead of `false`. The public metadata will advertise `rows` as optional even though the function rejects the one-argument call — Excel documents it as Required. That's a real (if cosmetic) regression, and it's the price of not touching the shared predicate. The catalogue can't override it, since optionality is derived entirely from `implementedFunctions`.

Concretely that means three assertions in the tests repo need updating — the full run was 6176 passed / 3 failed, and all three are assertions that encode the new shared-model semantics rather than TAKE's behaviour:

- `optional-parameters.spec.ts` → *"uses a required argument default only for syntactically empty input"* (the `REQUIREDDEFAULTTEST` plugin)
- `function-metadata-api.spec.ts` → *"returns full details for TAKE"* (the `optional` flags)
- `function-metadata-api.spec.ts` → *"derives optionality from the implementation on the fallback path"*

Reverting also means `docs/guide/custom-functions.md` needs no change — it becomes correct again as written.

### For the record: the right fix, later

The underlying issue is that `defaultValue` does two jobs — "value when the argument is omitted" and "value when the slot is empty". Splitting them (a separate `emptyValue`, or letting `emptyAsDefault` carry its own value) lets TAKE declare no `defaultValue` at all: it stays required under the *existing* optionality rule, empty slots still work, no existing flag changes meaning, and nothing third-party breaks. I prototyped that too and it gives `optional: [false, false, true]` with the full suite green apart from the same three assertions. Worth a backlog item alongside the `#CALC!` one — happy to file it if you want.

src/Cell.ts
Tobiadefami: Fixed in c8924aeee. Removed ErrorType.CALC, its entries from all 17 language packs, and the related error documentation and changelog text. TAKE now returns ErrorType.NA with ErrorMessage.ZeroRowOrColumnCount for zero counts.
 

Added regression coverage in a9f3334 confirming that language packs without a CALC translation register successfully, and documented the Excel deviation.

src/interpreter/plugin/FunctionPlugin.ts
Tobiadefami: Restored the established behavior where arguments with a defaultValue are treated as optional, and removed `optionalArg: false` from TAKE’s rows metadata

The required argument-position rule is now localized to TAKE: `TAKE(array)` returns `#N/A` with `WrongArgNumber`, while `TAKE(array, , columns)` remains valid and uses  the default row count.
 
Added coverage for default-value optionality, public function metadata, and TAKE’s argument-count behavior.
Implementation: 6804d04ac
Tests: 9685c61

src/interpreter/plugin/ArrayPlugin.ts
Tobiadefami: Fixed this in takeArraySize(). Bounded cases like `TAKE(A:A, 2)` still work, while results that remain unbounded return `#VALUE!`. 

I also added regression tests and documented the Excel difference.

src/interpreter/plugin/ArrayPlugin.ts
```diff
@@ -3,17 +3,52 @@
...
+  | {kind: 'unresolved'}
+
 export class ArrayPlugin extends FunctionPlugin implements FunctionPluginTypecheck<ArrayPlugin> {
+  /**
+   * Classifies a TAKE count for static result-size prediction.
+   *
+   * @param {Ast | undefined} argument - The count argument to inspect before evaluation.
+   * @returns {TakeLiteralDimension} The literal value, an invalid-literal marker, or an unresolved marker.
+   */
+  private parseTakeLiteralDimension(argument: Ast | undefined): TakeLiteralDimension {
```
marcin-kordas-hoc: `parseTakeLiteralDimension` only recognizes a bare `NUMBER`/`STRING` and a single unary `+`/`-`, so every other way of writing a constant count returns `unresolved` and falls back to reserving the full source size instead of the literal's value. `TAKE(range,TRUE())`, `TAKE(range,(1))` and `TAKE(range,0+1)` therefore produce a false `#SPILL!` when there is content near the true (smaller) result — reproduced against this PR's HEAD.

Real Excel evaluates all three of those as `1` (measured against Excel through the Graph API), so this is a genuine behavioural deviation rather than an HF convention. Worth noting the boundary so you don't over-fix: `200%` is *not* part of this — real Excel rejects that one too.

Could you extend the parser to cover these forms, plus regression tests for each spelling?

src/interpreter/plugin/ArrayPlugin.ts
```diff
@@ -166,6 +212,100 @@ export class ArrayPlugin extends FunctionPlugin implements FunctionPluginTypeche
...
+
+  /**
+   * Calculates the spilled array size of TAKE using the source dimensions as
+   * the upper bound. Unbounded results are rejected so the dependency graph
+   * never reserves an infinite spill range.
+   *
+   * @param ast
+   * @param state
+   */
+  public takeArraySize(ast: ProcedureAst, state: InterpreterState): ArraySize {
```
marcin-kordas-hoc, line 354: `TAKE` with a whole-column source (e.g. `TAKE(A:A,,1)`) returns `#VALUE!` "Cell range not allowed" here, even when the source is on the same sheet as the formula — confirmed live via MS Graph that real Excel spills correctly in exactly that case.

This does not look like an engine-wide limitation: `SORT`, `UNIQUE`, and `FILTER` in this same codebase already handle whole-column sources correctly, same-sheet and cross-sheet alike (checked all three live on current `develop`). Would you be open to reusing whatever `SortPlugin`/`UniquePlugin` do differently in their array-size prediction, so `takeArraySize` does not need a blanket rejection of non-finite dimensions? Happy to point at the exact lines if useful.

Separately: the cross-sheet case (`Data!A:A`, the example already in this PR's own `list-of-differences.md`) does genuinely fail in real Excel too (`#SPILL!`) — so that doc row's "Spills the whole column from row 1" claim is inaccurate for its own example. Worth correcting once the same-sheet case above is fixed.

marcin-kordas-hoc:
Nice catch fixing VSTACK/HSTACK to use Excel's real localized names across 14 languages — spot-checked two directly against Microsoft's own docs (French 'ASSEMB.V', German 'VSTAPELN') and both are correct.

Only ask: could you mention this in the PR description? Right now it only shows up in the auto-generated bot summary, and it's a real (good) breaking change for anyone who was using the placeholder English names in those locales before.

marcin-kordas-hoc:
The description still describes the `CALC` design that was dropped: *"It also introduces the `CALC` error type and its `#CALC!` representation in all built-in language packs. For now, `#CALC!` is produced only by `TAKE` when a row or column count evaluates to zero"*, plus the test-plan bullet *"Zero row and column counts returning `#CALC!`"*. The code returns `ErrorType.NA` with `ZeroRowOrColumnCount` — the right call per the thread above; the description just didn't follow. Worth fixing, since it's the first thing a reviewer reads and it points at behaviour that is deliberately absent.

Two small leftovers from that same removal:

- `docs/guide/types-of-errors.md` is still in the diff, but the only remaining change in it is a trailing newline at EOF. Either drop it from the PR or keep it on purpose.
- Please don't "fix" the `list-of-differences.md` row for TAKE while you're in there — HF `N/A` vs Excel `CALC` is correct as written (I checked it against real Excel).

Tobiadefami:
@marcin-kordas-hoc Thanks for flagging this. I moved the VSTACK/HSTACK localization fix into its own PR so the breaking language-pack migration can be reviewed separately: https://github.com/handsontable/hyperformula/pull/1748 (companion tests: 44). The changes are no longer part of this TAKE PR.

src/interpreter/plugin/ArrayPlugin.ts
```diff
@@ -166,6 +268,127 @@ export class ArrayPlugin extends FunctionPlugin implements FunctionPluginTypeche
...
+    const width = columnDimension.kind === 'value' ? Math.min(sourceSize.width, columnDimension.value) : sourceSize.width
+    const sourceRange = this.takeSourceRange(ast.args[0], state)
+    const canSpillUnboundedHeight = !Number.isFinite(height)
+      && sourceRange !== undefined
+      && sourceRange.sheet === state.formulaAddress.sheet
+      && state.formulaAddress.row === 0
+    const canSpillUnboundedWidth = !Number.isFinite(width)
+      && sourceRange !== undefined
+      && sourceRange.sheet === state.formulaAddress.sheet
+      && state.formulaAddress.col === 0
```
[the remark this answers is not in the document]
Tobiadefami: Fixed in cb5bc3c95 by using the formula address for placement and the source's effective dimensions for sizing; covered in test repo PR 28 by cf89480 and 8047c14.

src/i18n/languages/daDK.ts
```diff
@@ -19,8 +20,8 @@ const dictionary: RawTranslationPackage = {
   },
   functions: {
     FILTER: 'FILTER',
-    VSTACK: 'VSTACK',
-    HSTACK: 'HSTACK',
+    VSTACK: 'LSTAK',
+    HSTACK: 'VSTAK',
```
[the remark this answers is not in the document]
Tobiadefami: Removed from this PR. Corrected to `VSTAK` / `HSTAK` in engine PR 1748, with coverage in test repo PR 44.

src/interpreter/plugin/ArrayPlugin.ts
Tobiadefami: Fixed in 86a6a937d by materializing only the narrowed range and returning `onlyValues`; covered in test repo PR 28 by 131f650 and c424c5f.

src/interpreter/plugin/ArrayPlugin.ts
Tobiadefami: Fixed locally in cb5bc3c95 and 3fd5a702c. Static text, booleans, parentheses, unary forms, arithmetic, percentages, and literal errors are sized before spill allocation; covered in test repo PR 28.

src/interpreter/plugin/ArrayPlugin.ts
Tobiadefami: Fixed in cb5bc3c95 and 3fd5a702c; covered in test repo PR 28 by cf89480, 554ece1, and c781315.

src/i18n/languages/ptPT.ts
```diff
@@ -225,6 +226,7 @@ const dictionary: RawTranslationPackage = {
     SWITCH: '',
     SYD: 'SDA',
     T: 'T',
+    TAKE: 'INCLUIR',
```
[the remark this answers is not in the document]
Tobiadefami: Rechecked in live Excel Online with the PT-PT interface: both the Consulta e Referência menu and Insert Function dialog use `INCLUIR`. The PT-PT alphabetical list agrees; the individual TAKE support page is inconsistent. No code change is needed.

src/interpreter/plugin/ArrayPlugin.ts
Tobiadefami: Fixed in cb5bc3c95; covered in test repo PR 28 by cf89480 and 8047c14. Direct Excel Online verification showed cross-sheet sources also spill from the sheet edge, so the docs reflect edge-based placement.

sequba:
### `TRUE()`/`FALSE()` are constant-folded by procedure name, so overriding `TRUE` crashes the engine

`parseTakeLiteralNumber` folds zero-arg `TRUE()`/`FALSE()` by matching `ast.procedureName`, without consulting the function registry:

```ts
// src/interpreter/plugin/ArrayPlugin.ts:67-75
if (argument?.type === AstNodeType.FUNCTION_CALL && argument.args.length === 0) {
  if (argument.procedureName === 'TRUE') {
    return {kind: 'value', value: 1}
  }
  if (argument.procedureName === 'FALSE') {
    return {kind: 'value', value: 0}
  }
}
```

Overriding a built-in id is a supported API — `registerFunction`/`registerFunctionPlugin` accept it, and only ids in `_protectedPlugins` (`VERSION`) are refused. `TRUE` is not protected. Once a plugin overrides it, `takeArraySize` still statically predicts height `1` for `=TAKE(Data!A1:C10, TRUE())`, the array vertex is sized 1 row, evaluation produces 5, and `ArrayValue.resize` throws a bare `Error` that escapes uncaught through `Evaluator.recomputeFormulas` and out of `buildFromArray` / `buildFromSheets` / `setCellContents`:

```
### PHASE 1: built-in TRUE intact ###
=TAKE(Data!A1:C10,TRUE())    -> [[1,2,3]]
=SEQUENCE(TRUE())            -> [[1]]

### PHASE 2: TRUE overridden to return 5 ###
=TRUE()                      -> [[5]]
=TAKE(Data!A1:C10,TRUE())    -> THREW Error | Resizing to smaller array
=SEQUENCE(TRUE())            -> [[#VALUE!]]                 <- sibling degrades gracefully
=TAKE(Data!A1:C10,MYFIVE())  -> [[1,2,3],[4,5,6],...]        <- unresolved counts are fine
```

```
Error: Resizing to smaller array
    at ArrayValue.resize (src/ArrayValue.ts:141:13)
    at ArrayFormulaVertex.setCellValue (src/DependencyGraph/FormulaVertex.ts:132:11)
    at Evaluator.recomputeFormulaVertexValue (src/Evaluator.ts:141:21)
    at Evaluator.recomputeFormulas (src/Evaluator.ts:49:34)
```

The direction is what makes this specific to the shortcut. Over-prediction is harmless — a count the folder cannot resolve falls back to the full source dimension and `resize` pads with `EmptyValue`, which is why `MYFIVE()` above returns correctly. Only the hard-coded `TRUE`/`FALSE` branch can *under*-predict, and under-prediction is the throwing path. `SequencePlugin` has the same fold-by-name pattern at `SequencePlugin.ts:56` but returns `#VALUE!` under the identical override rather than throwing, so `TAKE` is currently the only function that turns a legitimate override into an uncaught exception.

After the escaped throw the cell is left at `#REF!` (the formula text survives, and a subsequent write recovers), so it is not silent data loss — but a bare `Error` out of a public API is worth closing before this ships.

Nothing in the test suite catches it: `function-take.spec.ts` pins only the built-in behaviour (`does not reserve unused rows when the count is TRUE()`, `coerces TRUE to one`, `coerces FALSE to zero and returns #N/A`), and no test registers a plugin over the `TRUE` id.

**Two ways to fix it**

1. Gate the shortcut on the id still resolving to the built-in — the engine's instance registry exposes `getFunctionPlugin('TRUE')`, so the fold can be taken only when that is `BooleanPlugin`. This keeps the existing size-prediction behaviour and the test at `function-take.spec.ts:367` passing.
2. Drop the branch entirely and let `TRUE()`/`FALSE()` fall through to `{kind: 'unresolved'}`. Simpler and provably safe, at the cost of reserving the full source footprint for those two counts (and of that one test).

Option 1 looks preferable since it preserves the tighter footprint.

Independently, and not this PR's to fix: `ArrayValue.resize`'s bare `Error` escaping a public API is a pre-existing engine issue. `ArrayValue.ts` and `FormulaVertex.ts` are byte-identical to the base commit, and the same throw reproduces on `develop` via `=SORT(A:A)`, `=UNIQUE(A:A)`, `=VSTACK(A:A)`, `=FILTER(A:A,…)` and plain `=A:A*2` when the source's used area grows. An under-prediction should surface as `#SPILL!`/`#REF!` rather than an exception — probably worth its own ticket.


---

sequba:
### A nested TAKE returns a spurious `#SPILL!` whenever the host cell is not in row 1 / column A

The "no space for an unbounded result" guard in `take()` keys off `state.formulaAddress` — the address of the cell holding the *whole* formula — not off whether this particular TAKE call is the thing being spilled:

```ts
// src/interpreter/plugin/ArrayPlugin.ts:356-364
if (sourceRange !== undefined) {
  const leavesHeightUnbounded = !Number.isFinite(sourceRange.height()) && !Number.isFinite(requestedRows)
  const leavesWidthUnbounded = !Number.isFinite(sourceRange.width()) && !Number.isFinite(requestedColumns)
  const startsBelowFirstRow = leavesHeightUnbounded && state.formulaAddress.row !== 0
  const startsRightOfFirstColumn = leavesWidthUnbounded && state.formulaAddress.col !== 0

  if (startsBelowFirstRow || startsRightOfFirstColumn) {
    return new CellError(ErrorType.SPILL, ErrorMessage.NoSpaceForArrayResult)
  }
```

For a nested call `state.formulaAddress` is the host cell's address — instrumenting the plugin shows `=SUM(TAKE(Data!A:A,,1))` in `A5` and a bare `=TAKE(Data!A:A,,1)` in `A5` both arrive with `sheet0!r4c0`, byte-identical. So the guard cannot distinguish them and fires for both, even though the nested TAKE is consumed as a value and never spills anything.

`Data` is `[[1],[4],[7]]` in all of the following:

```
=SUM(TAKE(Data!A:A,,1))         A1: 12    A5: #SPILL!
=COUNT(TAKE(Data!A:A,,1))       A1: 3     A5: 0
=AVERAGE(TAKE(Data!A:A,,1))     A1: 4     A5: #SPILL!
=MAX(TAKE(Data!A:A,,1))         A1: 7     A5: #SPILL!
=ROWS(TAKE(Data!A:A,,1))        A1: 3     A5: #SPILL!
=INDEX(TAKE(Data!A:A,,1),1,1)   A1: 1     A5: #SPILL!
=TAKE(TAKE(Data!A:A,,1),2)      A1: [[1],[4]]   A5: #SPILL!

column twin: =SUM(TAKE(Data2!1:1,1))   A1: 12    C1: #SPILL!
```

Every other array producer over the same range is correct at every row, so this is not a general engine limitation:

```
=SUM(ARRAY_CONSTRAIN(Data!A:A,3,1))   row0: 12   row2: 12
=SUM(FILTER(Data!A:A,Data!A:A>0))     row0: 12   row2: 12
=SUM(ARRAYFORMULA(Data!A:A))          row0: 12   row2: 12
=SUM(VSTACK(Data!A:A))                row0: 12   row2: 12
=SUM(TRANSPOSE(Data!A:A))             row0: 12   row2: 12
=SUM(Data!A:A)                        row0: 12   row2: 12
```

Two things worth being precise about:

- **The affected shape is narrower than "a whole-column TAKE".** It needs an unbounded source dimension *and* a non-finite count on that axis, i.e. an omitted or empty `rows`/`columns`. A finite count switches the guard off, so `=SUM(TAKE(Data!A:A,3))` is `12` at any row. Bounded sources are unaffected: `=SUM(TAKE(Data!A1:A3,,1))` is `12` in `A5`.
- **`COUNT` → `0` is not a second defect.** `COUNT` ignores error values (`=COUNT(NA())` is also `0`) and `COUNTA` counts them (`1`), so those are correct Excel semantics applied to the one spurious `#SPILL!`. It matters only because it means this can surface as a silently wrong number with no error visible anywhere.

The guard cannot simply be removed — it is load-bearing for the top-level spill tests (`returns #SPILL! for a cross-sheet whole-column result below the first row` and its three siblings). A usable discriminator does exist: `state.formulaVertex.formula === ast` is `false` for the nested call and `true` for both top-level cases. (`state.arraysFlag` and `state.formulaVertex` are otherwise identical between them, so neither works on its own.) Alternatively the position check belongs in `takeArraySize`, where spill placement is actually decided, rather than in the runtime value path where the result may never be spilled.

Not covered by the suite. Every spill assertion in `function-take.spec.ts` places a bare `=TAKE(...)` in the cell (lines ~157, ~178, ~200, ~220). Two tests do consume a nested TAKE — `=TAKE(Data!A1:C3,2)+0` (line 232) and `=ABS(TAKE(A1:A4,2))` (line 243) — but both use bounded sources, so nothing exercises this path and all 69 tests pass. Nesting cases with an unbounded source (`SUM`, `COUNT`, `INDEX`, at least one off row 0) look like the highest-value tests missing.


---

sequba:
### A non-foldable count over an unbounded source returns `#VALUE!` where `FILTER` returns `#SPILL!` — TAKE should match `FILTER`

`=TAKE(Data!A:A, Counts!A1)` placed anywhere below row 1 returns `#VALUE!` "Cell range not allowed." instead of spilling. `FILTER` — whose result size is *less* statically knowable than TAKE's — handles the same situation correctly, and so does TAKE's own literal-count path.

`Data` is `[[1],[4],[7]]`, `Counts!A1` is `2`:

| formula | in `A1` | in `A2` | vertex at `A2` |
|---|---|---|---|
| `=TAKE(Data!A:A,Counts!A1)` | spills `1, 4` | **`#VALUE!` "Cell range not allowed."** | scalar (`isCellPartOfArray` false) |
| `=FILTER(Data!A:A,Data!A:A>0)` | spills | `#SPILL!` "No space for array result." | array |
| `=SORT(Data!A:A)` | spills | `#SPILL!` | array |
| `=UNIQUE(Data!A:A)` | spills | `#SPILL!` | array |
| `=VSTACK(Data!A:A)` | spills | `#SPILL!` | array |
| `=TAKE(Data!A:A,,1)` | spills | `#SPILL!` | — |
| `=TAKE(Data!A:A,2)` | spills `1, 4` | spills `1, 4` | array |

So TAKE is inconsistent with `FILTER`, with three other array siblings, **and with its own literal-count branch**. `#VALUE!` also isn't a considered choice about TAKE's semantics — it is an internal scalar-expected error leaking out, which is why it has two knock-on symptoms `#SPILL!` doesn't: it poisons downstream references (`=SUM(A2:A3)` → `#VALUE!`) and it never recovers on recalculation, even after the count cell changes.

#### Root cause: the predictor cannot tell "unbounded" from "unknown"

`takeArraySize` classifies the count with `parseTakeLiteralDimension`, which has no `AstNodeType.EMPTY` branch — an empty argument and an unfoldable one both fall through to `{kind: 'unresolved'}`, and both then take `height = sourceSize.height` (`Infinity`) and hit the `startsBelowFirstRow` guard. Instrumenting the predictor shows the two producing a byte-identical result:

```
=TAKE(Data!A:A,,1)          args[1..2] astType=["EMPTY","NUMBER"]           -> ArraySize {w:1,h:1,isRef:true} scalar=true
=TAKE(Data!A:A,Counts!A1)   args[1..2] astType=["CELL_REFERENCE","MISSING"] -> ArraySize {w:1,h:1,isRef:true} scalar=true
=TAKE(Data!A:A,2)           args[1..2] astType=["NUMBER","MISSING"]         -> ArraySize {w:1,h:2,isRef:false} scalar=false
```

But the two cases are not the same thing:

- **empty argument** → the count really is `Infinity` (keep all rows), the result really is unbounded, and `#SPILL!` below row 1 is correct and is what the spec pins.
- **non-foldable argument** → the count is *finite*, just not known until evaluation. The result is 2 rows and fits perfectly at `A2`.

The reason they diverge downstream is that `ArraySize.error()` makes the engine build a **scalar** vertex in both cases. For the empty argument `take()`'s own runtime guard (`!Number.isFinite(requestedRows)`, `ArrayPlugin.ts:357-358`) also fires and returns a scalar `#SPILL!` `CellError`, which sits in a scalar vertex happily. For the non-foldable argument the evaluated count is finite, the guard correctly does not fire, and `take()` returns a `SimpleRangeValue` into a scalar vertex — which `Exporter` can only render as `#VALUE!`.

Note that the runtime half is already right: it keys off the *evaluated* count, so it treats these two cases differently, exactly as it should. Only the predictor conflates them.

#### Suggested fix — adopt `FILTER`'s model for the "unknown" case

`filterArraySize` never collapses an unresolvable size into an error; it takes the source's size as an upper bound and lets the engine's generic spill machinery decide whether it fits:

```ts
// filterArraySize — ArrayPlugin.ts:313-314
const width = Math.max(...(subChecks).map(val => val.width))
const height = Math.max(...(subChecks).map(val => val.height))
return new ArraySize(width, height)
```

Concretely, in `takeArraySize`:

1. Give the count classification a third outcome, distinguishing an `AstNodeType.EMPTY` argument (genuinely unbounded) from an unfoldable non-empty one (finite, unknown) — today both are `'unresolved'`. A `kind: 'unbounded'` alongside `'unresolved'` would do it.
2. Restrict `startsBelowFirstRow` / `startsRightOfFirstColumn` (`ArrayPlugin.ts:415-416`) to the genuinely-unbounded case, so the existing `#SPILL!` behaviour for `=TAKE(Data!A:A,,1)` and its three sibling spec tests is unchanged.
3. For the unknown-but-finite case, return `new ArraySize(effectiveWidth, effectiveHeight)` — the source's effective footprint as an upper bound, which the method already computes and already returns for this exact formula when it sits in row 1. That is the `FILTER` model: an honest array vertex, so the engine yields a real spill where it fits and an honest `#SPILL!` where it does not.

That keeps `take()` untouched, leaves every one of the 69 existing TAKE tests passing, and lines TAKE up with `FILTER`/`SORT`/`UNIQUE`/`VSTACK` and with its own literal-count branch in one change.

Worth adding tests for a cell-reference, cross-sheet-reference and function-call count over a whole-column and a whole-row source, at row 1 and below — the spec currently has 12 unbounded-source tests and every one uses a literal or omitted count, which is why this path is unguarded.

Two caveats so this isn't over-read. `|count| == 1` accidentally works today (`Interpreter` unwraps a 1×1 `SimpleRangeValue`), so the bug only bites for `|count| >= 2`. And separately, `FILTER` at row 1 spills only its first row rather than the full used height — a pre-existing quirk in how these siblings realise unbounded footprints, not something this PR should take on.


---

sequba:
### `list-of-differences.md`: the second TAKE row describes behaviour HyperFormula does not have

The row added at `docs/guide/list-of-differences.md:120`:

```
| TAKE | =TAKE(Data!A:A, , 1) | Spills the whole column from row 1; returns SPILL below row 1. | No such function. | Spills the whole column from row 1; returns SPILL below row 1. |
```

with the table's columns being `Function | Example | HyperFormula | Google Sheets | Microsoft Excel`.

**"Spills the whole column from row 1" is not what happens.** The realised spill is only as tall as the *source's used height*:

```
Data = 3 used rows    =TAKE(Data!A:A, , 1)   dims={"width":1,"height":3}   A3inArr=true  A4inArr=false  A40000inArr=false  serializedRows=3
Data = 10 used rows   =TAKE(Data!A:A, , 1)   dims={"width":1,"height":10}  A3inArr=true  A4inArr=true   A40000inArr=false  serializedRows=10
```

With the default `maxRows: 40_000` (`src/Config.ts:52`), "the whole column" would be 40 000 rows; the actual answer is 3. Nothing beyond the used height is claimed by the array — `A4` is outside it for a 3-row source, and `A40000` is outside it in both cases.

The second half of the sentence does reproduce: `=TAKE(Data!A:A, , 1)` in `A2` returns `#SPILL!` "No space for array result."

Three consequences:

1. **The row records no HyperFormula-vs-Excel difference at all.** The HyperFormula and Microsoft Excel cells contain identical text, so as written the row asserts parity with Excel — in a table whose purpose is to list differences. If the intent was to record a Google-Sheets-only difference, that should be explicit.

2. **Excel's actual behaviour is the difference this row should be capturing.** My understanding is that Excel's `=TAKE(A:A,,1)` fills the entire column (all 1 048 576 rows, blanks rendering as `0`), which would make HyperFormula's 3 rows a large and genuine divergence that the current wording hides. I have no way to confirm that against real Excel from this environment, so please verify before rewording the Excel cell — but either way the two columns should not be carrying the same sentence.

3. **HyperFormula isn't internally consistent here, which is worth a mention in whatever wording lands.** An explicitly oversized count *is* honoured and padded, while the omitted count collapses to the used height:

```
Data = 3 used rows    =TAKE(Data!A:A,6,1)    dims={"width":1,"height":6}   A4inArr=true   serializedRows=3
```

So `=TAKE(Data!A:A,6,1)` reserves six rows over a three-row source, but `=TAKE(Data!A:A,,1)` reserves three.

Suggested shape for the HyperFormula cell — something like "Spills only the used height of the source column, starting at row 1; returns SPILL below row 1." — with the Excel cell stating Excel's real behaviour once confirmed.

**Separately, on both new rows:** the Google Sheets cell says "No such function." for TAKE. I believe that is wrong — Google Sheets shipped `TAKE` alongside `CHOOSECOLS`, `CHOOSEROWS`, `HSTACK`, `VSTACK`, `TOROW`, `TOCOL`, `WRAPROWS`, `WRAPCOLS`, `DROP` and `EXPAND` in its 2022 dynamic-array batch. I could not verify this offline and found nothing in the repo that corroborates it either way, so it needs a quick check by someone with a browser; if it is wrong, both rows need that column filled in with Google's actual result. (The first row's other two cells check out: `=TAKE(A1:A3, 0)` does return `#N/A` in HyperFormula, and `CALC` in the Excel column matches the pre-existing `SEQUENCE` row's convention for the same situation.)


---

sequba (changes requested):

Tobiadefami:
Fixed in `5078a8c15`.

The runtime position guard has been removed. For a direct unbounded `TAKE` anchored away from the required sheet edge, `takeArraySize()` now preserves the unbounded predicted dimension so the normal array-space validation returns `#SPILL!` before runtime evaluation. A nested `TAKE` can therefore be consumed as a value regardless of the host formula's row or column.

Regression coverage now verifies both a nested whole-column result below row 1 and a nested whole-row result after column A. The existing four direct top-level `#SPILL!` cases remain green. The focused and full test suites, lint, and TypeScript compilation pass.

src/interpreter/plugin/ArrayPlugin.ts
```diff
@@ -166,6 +314,120 @@ export class ArrayPlugin extends FunctionPlugin implements FunctionPluginTypeche
...
+    const effectiveHeight = !Number.isFinite(height) && sourceRange !== undefined
+      ? sourceRange.effectiveHeight(this.dependencyGraph)
+      : height
+    const effectiveWidth = !Number.isFinite(width) && sourceRange !== undefined
+      ? sourceRange.effectiveWidth(this.dependencyGraph)
+      : width
+
+    if (startsBelowFirstRow || startsRightOfFirstColumn) {
+      return new ArraySize(width, height)
+    }
```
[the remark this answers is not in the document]
Tobiadefami: Fixed in 9bd7be465. takeArraySize() now checks for zero effective dimensions before applying off-edge unbounded spill sizing, so zero row or column counts preserve the zero-count error instead of being masked by #SPILL!. Added regressions for both whole-row and whole-column off-edge placements.

src/interpreter/plugin/ArrayPlugin.ts
```diff
@@ -166,6 +314,121 @@ export class ArrayPlugin extends FunctionPlugin implements FunctionPluginTypeche
...
+      : height
+    const effectiveWidth = !Number.isFinite(width) && sourceRange !== undefined
+      ? sourceRange.effectiveWidth(this.dependencyGraph)
+      : width
+
+    if (startsBelowFirstRow || startsRightOfFirstColumn || effectiveHeight < 1 || effectiveWidth < 1) {
+      return ArraySize.error()
+    }
+
+    return new ArraySize(effectiveWidth, effectiveHeight)
```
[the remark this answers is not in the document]
Tobiadefami: Fixed in `7c5e0b443`. Finite `TAKE` counts are now capped at `maxRows` or `maxColumns` before spill allocation, preventing oversized range traversal while preserving the requested reservation within the configured limits. Added regression coverage for oversized row and column counts, and Bugbot passes on the updated head.

Tobiadefami:
Thanks for flagging this. The HyperFormula behavior issue was valid, though the suggestion that Google Sheets supports `TAKE` was incorrect.

I updated the comparison to reflect the verified behavior:

- HyperFormula spills only to the source sheet's current height when entered in row 1, and returns `#SPILL!` below row 1.
- Google Sheets does not provide `TAKE`.
- Excel spills all 1,048,576 rows from row 1, converts blank source cells to `0`, and returns `#SPILL!` below row 1.

I verified the Excel behavior directly in Excel and pushed the documentation correction in commit `8d9e74e80`.

PR #1726, opened 2026-08-11
HF-307: entitlement model, capability registry, gate B (PR 1/4)

marcin-kordas-hoc:
### Context

HF-307 (feature packages / entitlement gating). This is PR 1 of 4: the entitlement model, the capability registry, and gate B in the interpreter. It only touches `src/`, has zero dependency on the (not-yet-shared) license-key package, and does not change behaviour for any key this library recognizes today (`gpl-v3`, legacy keys) — gate A (existing key-validity check) is untouched.

New:
- `src/license/LicenseEntitlement.ts` — `FeatureId`, `LicenseExpiry`, `LicenseEntitlement`, `unrestrictedEntitlement()`
- `src/license/capabilities.ts` — `CapabilityGrant`, `CAPABILITY_TABLE` (placeholder: every built-in under a single `core` token, refreshed from the static function registry on every `CapabilityRegistry` construction — see the comment on why it can't be built eagerly at module load)
- `src/license/CapabilityRegistry.ts` — `resolve()` (transitive, cycle-safe `implies` expansion), `capabilityOf()`, `allowsFunction()`, `allowsFeature()`

Modified:
- `src/Config.ts` — license-derived state added to the existing private pool, never exposed through `getConfig()`
- `src/interpreter/Interpreter.ts` — gate B added after gate A in the `FUNCTION_CALL` case, with a custom-function exemption (a function not covered by the capability table is treated as instance-registered/custom, not gated) and alias canonicalization (an alias must gate identically to its canonical function name)
- `src/error-message.ts` — `ErrorMessage.LicenseCapability`

This PR ships without a real license-key payload adapter (that's a later PR), so `Config` always resolves an unrestricted entitlement today — gate B is a correct, independently-testable no-op in production until that adapter lands. `FeatureId.CustomFunctions` is kept in the enum as reserved vocabulary (no grant maps to it this release) per team decision to drop function-registration gating from this pass; happy to remove it instead if reviewers prefer.

- `eslint` on all new/changed files: clean.
- `test/smoke.spec.ts`: passes.
- Manual verification against the built engine (not committed, ts-node scratch scripts): `gpl-v3` and missing/invalid-key behaviour unchanged (gate A untouched, exact same `#LIC!` messages); `getConfig()` does not expose the new license-derived state; a function outside a restricted entitlement's granted set returns `#LIC!` with the new message while a function inside it computes normally; `VERSION`/`OFFSET` (protected) bypass the gate regardless of restriction; a custom function (`config.functionPlugins`) is exempt from gate B; calling a gated built-in through an alias (e.g. `VAR` vs. canonical `VAR.S`) gates identically to the canonical name, not the alias string; `CapabilityRegistry.resolve()` handles transitive `implies`, cycles, and unknown tokens correctly.
- **Not yet done, flagging explicitly**: the private test suite (paired branch in `hyperformula-tests`) isn't pushed yet — I don't have that repo checked out in this environment. I also haven't run a full before/after benchmark on the interpreter hot path; the design keeps the common path to a single boolean read (`isLicenseGateActive === false`) replacing today's string-enum comparison, but I want real numbers in before removing draft status.

### Types of changes
- [ ] Breaking change
- [ ] Bug fix

### Related issues:

### Checklist:
- [ ] I have signed the Contributor License Agreement. *(please confirm/attach on your end — I can't verify this from here)*
- [ ] My change is compliant with the OpenDocument standard. *(N/A — no worksheet function behaviour changes in this PR)*
- [ ] My change is compatible with Microsoft Excel. *(N/A — same reason)*
- [ ] My change is compatible with Google Sheets. *(N/A — same reason)*
- [ ] I described my changes in the CHANGELOG.md file. *(intentionally not done — internal-only change, no user-visible behaviour yet; a CHANGELOG entry lands with the PR that actually activates gate B for customers)*
- [ ] My changes require a documentation update. *(no — nothing user-visible yet)*

Opening as a **draft**: the private test suite isn't pushed yet, the hot-path benchmark hasn't been run, and one design question is still open (whether to keep `FeatureId.CustomFunctions` as reserved or drop it entirely). Will mark ready once those are resolved.

### DIFF
--- a/src/Config.ts
+++ b/src/Config.ts
@@ -20,11 +20,24 @@ import {checkLicenseKeyValidity, LicenseKeyValidityState} from './helpers/licens
 import {HyperFormula} from './HyperFormula'
 import {TranslationPackage} from './i18n'
 import {FunctionPluginDefinition} from './interpreter'
+import {CapabilityRegistry, ResolvedCapabilities} from './license/CapabilityRegistry'
+import {unrestrictedEntitlement} from './license/LicenseEntitlement'
 import {Maybe} from './Maybe'
 import {ParserConfig} from './parser/ParserConfig'
 import {ConfigParams, ConfigParamsList} from './ConfigParams'
 
-const privatePool: WeakMap<Config, { licenseKeyValidityState: LicenseKeyValidityState }> = new WeakMap()
+/**
+ * The license-derived state kept off the public `ConfigParams` surface — see
+ * {@link Config.licenseCapabilities}.
+ */
+interface LicensePrivateState {
+  licenseKeyValidityState: LicenseKeyValidityState,
+  licenseCapabilities: ResolvedCapabilities,
+  isLicenseGateActive: boolean,
+  capabilityRegistry: CapabilityRegistry,
+}
+
+const privatePool: WeakMap<Config, LicensePrivateState> = new WeakMap()
 
 export class Config implements ConfigParams, ParserConfig {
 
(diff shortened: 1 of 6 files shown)


qunabu:

marcin-kordas-hoc:
**prep-flip T1 (static) — PASS**

Ran our internal `prep-flip` review-readiness tool's T1 tier (lint / tsc / snippets-drift / fork-guard) against `0a966e35` — the mechanical checks that don't need network or paired-repo access:

```json
{
  "tier": "T1",
  "verdict": "PASS",
  "findings": [
    { "id": "T1-SNIPPETS-DRIFT-O5", "severity": "P1", "evidence": "" }
  ]
}
```

`lint` and `tsc` both pass clean. The one finding is `npm run snippets:check` not existing as a script in this repo (unrelated tooling config, not a defect in this diff) — flagging it here rather than silently dropping it since a P1 finding shouldn't just disappear from the record.

Tiers T2.5 (cross-repo contract, needs a paired `hyperformula-tests` PR — none exists yet for this change), T4a (Bugbot scrape), T4b (xlsx gate), and the LLM-judgement tiers (T5a/T5b/T6) weren't run this session — no network/paired-PR access from this environment. Noting the gap rather than claiming a full run.

marcin-kordas-hoc:
Superseded by #1728 — same branch, pushed directly to this repo instead of the fork so CI can access the private test-suite checkout (fork-triggered runs don't get repo secrets).

PR #1728, opened 2026-08-11
HF-307: entitlement model, capability registry, gate B (PR 1/4)

marcin-kordas-hoc:

### Context

HF-307 (feature packages / entitlement gating). This is PR 1 of 4: the entitlement model, the capability registry, and gate B in the interpreter. It only touches `src/`, has zero dependency on the (not-yet-shared) license-key package, and does not change behaviour for any key this library recognizes today (`gpl-v3`, legacy keys) — gate A (existing key-validity check) is untouched.

New:
- `src/license/LicenseEntitlement.ts` — `FeatureId`, `LicenseExpiry`, `LicenseEntitlement`, `unrestrictedEntitlement()`
- `src/license/capabilities.ts` — `CapabilityGrant`, `CAPABILITY_TABLE` (placeholder: every built-in under a single `core` token, refreshed from the static function registry on every `CapabilityRegistry` construction — see the comment on why it can't be built eagerly at module load)
- `src/license/CapabilityRegistry.ts` — `resolve()`, `capabilityOf()`, `allowsFunction()`, `allowsFeature()`. ~~transitive, cycle-safe `implies` expansion~~ **removed 22.09**: no grant ever set `implies`, and the table is documented as fully expanded, so the recursive walk resolved nothing a flat pass does not. The queue and the cycle guard went with it; the visited set stays, now only to expand each distinct token once.

Modified:
- `src/Config.ts` — license-derived state added to the existing private pool, never exposed through `getConfig()`
- `src/interpreter/Interpreter.ts` — gate B added after gate A in the `FUNCTION_CALL` case, with a custom-function exemption (a function not covered by the capability table is treated as instance-registered/custom, not gated) and alias canonicalization (an alias must gate identically to its canonical function name)
- `src/error-message.ts` — `ErrorMessage.LicenseCapability`

This PR ships without a real license-key payload adapter (that's a later PR), so `Config` always resolves an unrestricted entitlement today — gate B is a correct, independently-testable no-op in production until that adapter lands. `FeatureId.CustomFunctions` is kept in the enum as reserved vocabulary (no grant maps to it this release) per team decision to drop function-registration gating from this pass; happy to remove it instead if reviewers prefer.

- `eslint` on all new/changed files: clean.
- `test/smoke.spec.ts`: passes.
- Manual verification against the built engine (not committed, ts-node scratch scripts): `gpl-v3` and missing/invalid-key behaviour unchanged (gate A untouched, exact same `#LIC!` messages); `getConfig()` does not expose the new license-derived state; a function outside a restricted entitlement's granted set returns `#LIC!` with the new message while a function inside it computes normally; `VERSION`/`OFFSET` (protected) bypass the gate regardless of restriction; a custom function (`config.functionPlugins`) is exempt from gate B; calling a gated built-in through an alias (e.g. `VAR` vs. canonical `VAR.S`) gates identically to the canonical name, not the alias string; `CapabilityRegistry.resolve()` handles unknown tokens correctly. (The `implies` part of this line no longer applies — see the note above.)
**Update:** private test suite pushed — [hyperformula-tests#30](https://github.com/handsontable/hyperformula-tests/pull/30) (`unit/license/{entitlement,capability-registry,config-wiring}.spec.ts` + a gate-B section appended to `unit/licence.spec.ts`, the 2 existing cases untouched). Against the paired branch: the new license specs pass, and the **full existing suite (505 suites / 6204 tests, 0 failures)** is unaffected. CI on this PR now runs the real suite (pushed directly here rather than from a fork, so the private-repo checkout has secrets access) — `build`/`lint`/`unit-tests`/`browser-tests`/`Test performance`/`CodeQL`/audit/Snyk all pass.

**Task 1.5 benchmark**, from this PR's own CI run (`Test performance` job, base `61ead730` vs head `0a966e35`):

```
                                     testName |    base |    head |  change
---------------------------------------------------------------------------
                                      Sheet A |  499.81 |  501.66 |  +0.37%
                                      Sheet B |  156.27 |  167.13 |  +6.95%
                                      Sheet T |  138.85 |  147.87 |  +6.50%
                                Column ranges |  469.57 |  485.16 |  +3.32%
                                Sorted lookup | 14182.7 | 14444.5 |  +1.85%
Sheet A:  change value, add/remove row/column |   15.32 |   17.05 | +11.29%
 Sheet B: change value, add/remove row/column |  132.54 |  139.87 |  +5.53%
                   Column ranges - add column |  147.38 |  156.86 |  +6.43%
                Column ranges - without batch |  459.94 |  464.95 |  +1.09%
                        Column ranges - batch |  114.68 |  112.74 |  -1.69%
```

Reporting this as-is rather than smoothing it over: Sheet A (the highest-resolution signal) is flat (+0.37%), but several rows show a consistent single-digit-percent slowdown on this one run, on shared GitHub Actions runners rather than the codespace the spike's own noise-floor numbers were calibrated against. I have not repeated the run to separate signal from noise — happy to re-run `Test performance` a second time before merge if reviewers want more confidence than a single sample gives. The structural claim I *can* stand behind: the common path (`isLicenseGateActive === false`, checked before gate B does anything) is one boolean read replacing today's string-enum comparison, so there's no new work on the hot path for `gpl-v3`/legacy keys — that part is a code-inspection fact, not a benchmark claim.

Codecov: patch coverage is 58.1% (`CapabilityRegistry.ts` 48.7%, `Interpreter.ts` 41.7%, `Config.ts` 71.4%) and project coverage -0.19%, both **still red**. This is checked against `develop`, but the new private tests exercise the paired branch, not this diff directly through Codecov's own accounting — worth a look before merge, though the "full suite green" result above is the stronger correctness signal.

### Types of changes
- [ ] Breaking change
- [ ] Bug fix

### Related issues:

### Checklist:
- [ ] I have signed the Contributor License Agreement. *(please confirm/attach on your end — I can't verify this from here)*
- [ ] My change is compliant with the OpenDocument standard. *(N/A — no worksheet function behaviour changes in this PR)*
- [ ] My change is compatible with Microsoft Excel. *(N/A — same reason)*
- [ ] My change is compatible with Google Sheets. *(N/A — same reason)*
- [ ] I described my changes in the CHANGELOG.md file. *(intentionally not done — internal-only change, no user-visible behaviour yet; a CHANGELOG entry lands with the PR that actually activates gate B for customers)*
- [ ] My changes require a documentation update. *(no — nothing user-visible yet)*

**Ready for review.** The paragraph that used to sit here said this was opening as a draft pending the private suite, the benchmark and the `FeatureId.CustomFunctions` question — all three are resolved: the suite is pushed (hyperformula-tests#30), the benchmark ran, and Kuba settled the token with decision D1 (dropped from this release, kept as reserved vocabulary).

---

> [!NOTE]
> **Medium Risk**
> Changes the formula evaluation hot path for license checks and introduces entitlement resolution at config init; production is a no-op today but incorrect future wiring could deny functions or leak license state.
> 
> **Overview**
> Introduces **HF-307 license entitlement infrastructure** (PR 1/4): a `LicenseEntitlement` model, a `CapabilityRegistry` that maps capability tokens to function/feature grants, and a placeholder `core` capability table refreshed from the static function registry at `Config` construction.
> 
> **`Config`** now keeps license-derived state in a private pool—resolved capabilities, `isLicenseGateActive`, and the registry—without exposing it via `getConfig()`. Until a real license-key payload adapter lands (later PR), every instance still resolves **`unrestrictedEntitlement()`**, so production behavior for `gpl-v3`/legacy keys stays the same while gate B is testable.
> 
> **Interpreter gate B** runs only when `isLicenseGateActive` is true: invalid keys still get the existing `#LIC!` message; then built-ins covered by the capability table must pass `allowsFunction`, with **alias canonicalization** and an exemption for functions **not** in the table (treated as custom). Adds **`ErrorMessage.LicenseCapability`** for per-function denials.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 40ad18aa56c69c9043eb7f56e2b1a6d0aa597376. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/license/CapabilityRegistry.ts
+++ b/src/license/CapabilityRegistry.ts
@@ -1,0 +1,126 @@
+/**
+ * @license
+ * Copyright (c) 2025 Handsoncode. All rights reserved.
+ */
+
+import {FeatureId, LicenseEntitlement} from './LicenseEntitlement'
+import {CAPABILITY_TABLE, CapabilityGrant, refreshCoreGrant} from './capabilities'
+
+/**
+ * The capabilities a resolved {@link LicenseEntitlement} grants, ready for gate B (the
+ * interpreter) and PR 2's `ensureCapability` to query through {@link allowsFunction} and
+ * {@link allowsFeature}.
+ */
+export interface ResolvedCapabilities {
+  /** `true` short-circuits both {@link allowsFunction} and {@link allowsFeature} to `true`. */
+  unrestricted: boolean,
+  functions: ReadonlySet<string>,
+  features: ReadonlySet<FeatureId>,
+}
+
+/**
+ * Expands a {@link LicenseEntitlement}'s capability tokens against a table of
+ * {@link CapabilityGrant}s into the concrete functions and features they grant, and answers
+ * which token, if any, covers a given function id.
+ */
+export class CapabilityRegistry {
+  private readonly table: ReadonlyMap<string, CapabilityGrant>
+  private readonly reverseIndex: ReadonlyMap<string, string>
+
+  /**
+   * @param {ReadonlyMap<string, CapabilityGrant>} [table] - capability table to resolve
+   * against. Omit to use the production {@link CAPABILITY_TABLE}; tests inject their own so the
+   * suite does not depend on its placeholder content.
+   */
+  constructor(table?: ReadonlyMap<string, CapabilityGrant>) {
+    if (table === undefined) {
+      refreshCoreGrant()
+    }
+    this.table = table ?? CAPABILITY_TABLE
+    this.reverseIndex = CapabilityRegistry.buildReverseIndex(this.table)
+  }
+
+  /**
+   * Inverts a capability table from token → grant into function id → token, so
+   * {@link capabilityOf} is a single lookup instead of a scan. The first token that lists a
+   * given function id wins, in table iteration order.
+   *
+   * @param {ReadonlyMap<string, CapabilityGrant>} table - the table to invert
+   */
+  private static buildReverseIndex(table: ReadonlyMap<string, CapabilityGrant>): ReadonlyMap<string, string> {
+    const index = new Map<string, string>()
+    for (const [token, grant] of table) {
+      for (const functionId of grant.functions) {
+        if (!index.has(functionId)) {
+          index.set(functionId, token)
+        }
+      }
+    }
+    return index
+  }
+
+  /**
+   * Expands an entitlement's capability tokens into the concrete functions and features they
+   * grant. An `unrestricted` entitlement short-circuits to an unrestricted result without
+   * consulting the table at all. Expansion through `implies` is transitive and cycle-safe (a
+   * visited set guards against a token implying itself, directly or through others); an
+   * unrecognized token is skipped without an error.
+   *
+   * @param {LicenseEntitlement} entitlement - the entitlement to resolve, e.g. one built by
+   * hand in a test or produced by PR 3's license-key payload adapter
+   */
+  public resolve(entitlement: LicenseEntitlement): ResolvedCapabilities {
+    if (entitlement.unrestricted) {
+      return {unrestricted: true, functions: new Set<string>(), features: new Set<FeatureId>()}
+    }
+
+    const functions = new Set<string>()
+    const features = new Set<FeatureId>()
+    const visited = new Set<string>()
+    const queue = [...entitlement.capabilities]
+
+    while (queue.length > 0) {
+      const token = queue.shift() as string
+      if (visited.has(token)) {
+        continue
+      }
+      visited.add(token)
+
+      const grant = this.table.get(token)
+      if (grant === undefined) {
+        continue
+      }
+      grant.functions.forEach((functionId) => functions.add(functionId))
+      grant.features.forEach((feature) => features.add(feature))
+      grant.implies?.forEach((impliedToken) => queue.push(impliedToken))
+    }
+
+    return {unrestricted: false, functions, features}
+  }
+
+  /**
+   * Returns the capability token a function id is covered by, or `undefined` if this registry's
+   * table does not cover it. The completeness invariant in
+   * `unit/license/capability-registry.spec.ts` guarantees every built-in registered in the
+   * static function registry is covered by the table, the core token, or the protected list —
+   * so `undefined` for a function known to the current instance's function registry means it is
+   * a custom, instance-registered function rather than an unlisted built-in.
+   */
+  public capabilityOf(functionId: string): string | undefined {
+    return this.reverseIndex.get(functionId)
+  }
+}
+
+/**
+ * Whether a resolved entitlement allows calling the given function.
+ */
+export function allowsFunction(resolved: ResolvedCapabilities, functionId: string): boolean {
+  return resolved.unrestricted || resolved.functions.has(functionId)
+}
+
+/**
+ * Whether a resolved entitlement allows using the given feature area of the public API.
+ */
+export function allowsFeature(resolved: ResolvedCapabilities, feature: FeatureId): boolean {
+  return resolved.unrestricted || resolved.features.has(feature)
+}
--- a/src/license/capabilities.ts
+++ b/src/license/capabilities.ts
@@ -1,0 +1,56 @@
+/**
+ * @license
+ * Copyright (c) 2025 Handsoncode. All rights reserved.
+ */
+
+import {FunctionRegistry} from '../interpreter/FunctionRegistry'
+import {FeatureId} from './LicenseEntitlement'
+
+/** The single capability token every built-in currently falls under, see {@link CAPABILITY_TABLE}. */
+export const CORE_TOKEN = 'core'
+
+/**
+ * Describes what a capability token grants: a set of function ids, a set of {@link FeatureId}
+ * values, and optionally other tokens it implies. `implies` is expanded recursively by
+ * `CapabilityRegistry.resolve`, not by anything in this file.
+ */
+export interface CapabilityGrant {
+  functions: string[],
+  features: FeatureId[],
+  implies?: string[],
+}
+
+const coreGrant: CapabilityGrant = {
+  functions: [],
+  features: [FeatureId.NamedExpressions, FeatureId.Clipboard, FeatureId.Crud, FeatureId.UndoRedo, FeatureId.Batching],
+}
+
+/**
+ * The production capability table.
+ *
+ * Placeholder content pending HF-331/HF-329 (the real per-package token vocabulary): every
+ * built-in function, plus the features already wired for gating in PR 2, fall under the single
+ * {@link CORE_TOKEN}. `FeatureId.CustomFunctions` and `FeatureId.ImportExport` are deliberately
+ * absent — reserved vocabulary with no grant yet (HF-307 decision D1; HF-107 for ImportExport).
+ *
+ * `coreGrant.functions` starts empty and is refreshed on every {@link refreshCoreGrant} call
+ * rather than populated once here: `src/index.ts` registers HyperFormula's built-in plugins as a
+ * side effect of being imported, and it does so AFTER `Config` and `Interpreter` — and so this
+ * module — have already been fully evaluated. Reading the function registry at module-load time
+ * would capture an empty registry.
+ */
+export const CAPABILITY_TABLE: ReadonlyMap<string, CapabilityGrant> = new Map([[CORE_TOKEN, coreGrant]])
+
+/**
+ * Refreshes the placeholder `core` grant with every function currently in the static function
+ * registry. Called from `CapabilityRegistry`'s constructor every time it is constructed without
+ * an explicit table — not just the first time: the static registry can change after the first
+ * engine is built (`HyperFormula.registerFunctionPlugin`/`unregisterFunctionPlugin` are public,
+ * documented APIs), and a one-time snapshot would silently go stale for every engine built
+ * afterward. Cheap (a single array copy from an existing map's keys) and only ever runs once per
+ * `Config`/engine construction, never on the per-formula hot path, so re-running it every time
+ * costs nothing worth guarding against with memoization.
+ */
+export function refreshCoreGrant(): void {
+  coreGrant.functions = FunctionRegistry.getRegisteredFunctionIds()
+}
--- a/src/Config.ts
+++ b/src/Config.ts
@@ -20,11 +20,24 @@ import {checkLicenseKeyValidity, LicenseKeyValidityState} from './helpers/licens
 import {HyperFormula} from './HyperFormula'
 import {TranslationPackage} from './i18n'
 import {FunctionPluginDefinition} from './interpreter'
+import {CapabilityRegistry, ResolvedCapabilities} from './license/CapabilityRegistry'
+import {unrestrictedEntitlement} from './license/LicenseEntitlement'
 import {Maybe} from './Maybe'
 import {ParserConfig} from './parser/ParserConfig'
 import {ConfigParams, ConfigParamsList} from './ConfigParams'
 
-const privatePool: WeakMap<Config, { licenseKeyValidityState: LicenseKeyValidityState }> = new WeakMap()
+/**
+ * The license-derived state kept off the public `ConfigParams` surface — see
+ * {@link Config.licenseCapabilities}.
+ */
+interface LicensePrivateState {
+  licenseKeyValidityState: LicenseKeyValidityState,
+  licenseCapabilities: ResolvedCapabilities,
+  isLicenseGateActive: boolean,
+  capabilityRegistry: CapabilityRegistry,
+}
+
+const privatePool: WeakMap<Config, LicensePrivateState> = new WeakMap()
 
 export class Config implements ConfigParams, ParserConfig {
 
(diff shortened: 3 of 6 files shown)


qunabu:

Tobiadefami (commented):
Reviewed at 0a966e35 together with the paired tests at 28952a33. The entitlement model, capability resolution, private Config wiring, and interpreter gate look good at this layer. The focused paired suite passes (4 suites / 23 tests), the current engine checks are green, and I found no material issue in this PR.

src/license/CapabilityRegistry.ts
```diff
@@ -0,0 +1,126 @@
...
+import {CAPABILITY_TABLE, CapabilityGrant, refreshCoreGrant} from './capabilities'
+
+/**
+ * The capabilities a resolved {@link LicenseEntitlement} grants, ready for gate B (the
+ * interpreter) and PR 2's `ensureCapability` to query through {@link allowsFunction} and
+ * {@link allowsFeature}.
+ */
+export interface ResolvedCapabilities {
+  /** `true` short-circuits both {@link allowsFunction} and {@link allowsFeature} to `true`. */
+  unrestricted: boolean,
```
sequba, line 16: We need encode also license keys that include all functions but not all features, so instead of single `unrestricted` flag I'd rather have
```
functions: ReadonlySet<string> | 'all',
features: ReadonlySet<FeatureId> | 'all',
```

`unrestricted === true` is equivalent to `functions === 'all' & `features === 'all'`

src/license/capabilities.ts
```diff
@@ -0,0 +1,56 @@
...
+export interface CapabilityGrant {
+  functions: string[],
+  features: FeatureId[],
+  implies?: string[],
+}
+
+const coreGrant: CapabilityGrant = {
+  functions: [],
+  features: [FeatureId.NamedExpressions, FeatureId.Clipboard, FeatureId.Crud, FeatureId.UndoRedo, FeatureId.Batching],
+}
```
sequba, line 25: I think you missed the point of `core`. Idea is that `core` grants access to all features except NamedExpressions, Clipboard, Crud, UndoRedo, Batching, CustomFunctions and ImportExport. The code should not validate that grant. It's just a way to logically represent all the other feature inside a license key. The code should just ignore it.

sequba (changes requested):

src/license/capabilities.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Describes what a capability token grants: a set of function ids, a set of {@link FeatureId}
+ * values, and optionally other tokens it implies. `implies` is expanded recursively by
+ * `CapabilityRegistry.resolve`, not by anything in this file.
+ */
+export interface CapabilityGrant {
+  functions: string[],
+  features: FeatureId[],
+  implies?: string[],
```
sequba: What is the point of `implies`?

marcin-kordas-hoc: @sequba None, you are right — nothing used it. No grant in the table ever set `implies`, and the table's own comment says the grants are stored fully expanded, so the recursive walk resolved nothing a flat pass does not. The field, the queue and the cycle guard are gone; the visited set stays, now only to expand each distinct token once.

The specs took two passes, and the first one was wrong in a way worth flagging. It claimed to pin "a repeated token grants no more than a single one" — but it could not fail: an entitlement carries its tokens in a `Set`, so the duplicate never reaches the registry, and set union is idempotent anyway. Deleting the deduplication from the engine left it green. It is gone.

What the single-expansion promise is actually about is several *spellings* of one token, which normalisation collapses. That pin now lives in PR 3, where `normalizeCapabilityToken` exists, and it counts the lookups made against the table rather than the grant returned — the grant is identical either way and would hide a registry expanding one token four times.

src/license/capabilities.ts
marcin-kordas-hoc: @sequba Partly already done, and I want to flag one thing before doing the rest, because taken literally it breaks something.

The premise is out of date — that is my fault for leaving PR 1 reviewable in isolation. In PR 1 `coreGrant` did grant the five features you list. By PR 3 it is `{functions: [...OPERATOR_FUNCTIONS], features: []}`, and `capabilities.ts` says so in a comment: "An earlier revision granted all five features from CORE_TOKEN, which made feature gating inert". So the feature half of your objection is already applied.

What is left is that `core` still grants the operator functions. If the code simply ignores the token, a key naming only `core` loses the explicit callable forms — `=HF.ADD(…)` and friends — because those are covered by single-function entries in the table. (The infix operators themselves are fine either way: gate B only runs in the interpreter's `FUNCTION_CALL` branch, and `=A1+B1` never reaches it.)

So the change is doable, but it has to move the operators out of the token system into the protected built-ins path that already exists in the same file, not just drop the grant. Confirm that is what you want and I will do it in the same pass as the capability reshape.

src/license/CapabilityRegistry.ts
marcin-kordas-hoc: @sequba Agreed, and the gap is real — a key granting every function but not every feature cannot be expressed today. Two things before I do it, because this one is not local.

First, the size. `ResolvedCapabilities` is byte-identical at the head of all four PRs, so this is the shipping shape rather than scaffolding: 26 references across 6 files in `src/`, plus about a dozen spec files. It sits in PR 1, so it propagates through the whole stack and both paired suites.

Second, the part I want to be careful about. `unrestricted` is what carries fail-open: a missing, invalid or expired key resolves to an unrestricted entitlement so that nothing is gated. Split into two axes, fail-open has to set `'all'` on both, in every path that produces one — and missing it on one axis fails *closed*, which is the direction that breaks working installations rather than the one that leaks. So I want a test pinning fail-open per axis written before the change, not after.

Planned unless you say otherwise, but it depends on the scoping question in my reply about the packaging constants, since both reshape the same file.

sequba: Still, let's do it despite all of this.

src/license/capabilities.ts
sequba: operator functions should be controlled by `fun:operator.A` capability as described in https://app.clickup.com/9015210959/v/dc/8cnjcyf-33175/8cnjcyf-47835

src/license/capabilities.ts
sequba: Let's remove `implies`

src/license/CapabilityRegistry.ts
marcin-kordas-hoc: @sequba Done.

src/license/capabilities.ts
marcin-kordas-hoc: @sequba Done. `HF.*` forms now need `fun:operator.A`; infix operators are unaffected. §4 of the doc still says the `HF.*` forms are always available.

src/license/capabilities.ts
marcin-kordas-hoc: @sequba Done.

src/license/handsontable-license-key-parser/PROVENANCE.md
```diff
@@ -0,0 +1,97 @@
...
+|---|---|
+| `generate-key.js`, `build-payload.js`, `build-prose.js` | Mint keys. HyperFormula only ever reads them. |
+| `default-schema.js` | The generator's vocabulary (packages, add-ons, wordings, templates). The reader needs no schema; the only name this library reads is its own product entry, kept as `HYPERFORMULA_PRODUCT_NAME` in `src/license/licenseResolution.ts`. |
+| `create-engine.js`, `resolve-schema.js`, `validate-schema.js`, `validate-record.js` | Bind and verify a caller's schema/record at generation time — generator-side. |
+| `validate-key.js` | A two-line boolean wrapper over `extractEntitlementKeyData`; the extractor is called directly. |
+
+From `utils.js`, the two generation-side helpers `bytesToBase64` and `stringToBase64Url` are
+also left out. Everything else in that file is ported.
+
+## Deliberate divergences from upstream
```
sequba: Let's continue this discussion here 
image

sequba (changes requested):

src/license/handsontable-license-key-parser/PROVENANCE.md
sequba: image

sequba: I'm leaning toward option 1 "Verbatim upstream .js + allowJs + hand-written .d.ts — prototyped". But I want to understand, why is `isIsoDate` modified? What is the reason not to use the upstream version?

src/license/capabilities.ts
```diff
@@ -0,0 +1,289 @@
...
+]
+
+/** The whole implemented catalog of built-in functions, operator callable forms included. */
+export const FUN_ALL_TOKEN = 'fun:all'
+/**
+ * Spreadsheet Bundle add-on (2026-08-12 packages meeting). Grants {@link FeatureId.Crud},
+ * {@link FeatureId.UndoRedo}, {@link FeatureId.Clipboard} and {@link FeatureId.Batching} — see
+ * {@link CAPABILITY_TABLE}.
+ */
+export const SPREADSHEET_ADDON_TOKEN = 'spreadsheet'
```
sequba: Let's continue this discussion here

Image

sequba: Import_export should also become `feat:*` and spreadsheet add-on should not exist in this code. This code should know nothing about any add-ons and what features they include

docs/guide/license-key.md
```diff
@@ -46,19 +46,70 @@ const options = {
...
-* The date in your proprietary license key
+Which versions of HyperFormula a key covers, and for how long, follows from the
+terms of your contract. Your key carries those terms, and HyperFormula applies
+them locally, without any connection to a server.
+
+## Feature packages and add-ons
+
+A proprietary license key may grant the whole library, or only part of it. If your key covers
+everything you buy nothing new to think about, and neither does the GPLv3 key `gpl-v3`, which
+always grants everything.
```
sequba:
```suggestion
A proprietary license key may grant the whole library, or only part of it.
```

docs/guide/license-key.md
```diff
@@ -46,19 +46,68 @@ const options = {
...
+In this release, not having either add-on doesn't restrict anything either: a key that names no
+feature token at all is granted every feature area — CRUD, undo/redo, clipboard, named expressions
+and batching — regardless of whether it names these add-ons. Every key issued today is of that
+shape, so the add-on tokens describe what was sold rather than changing what the engine allows.
+
+::: tip
+To find out which package your key includes, check your order confirmation or
+[contact our team](contact.md). HyperFormula deliberately reports nothing about the contents of
+your key at runtime.
+:::
```
sequba:
```suggestion
```

docs/guide/license-key.md
```diff
@@ -46,19 +46,44 @@ const options = {
...
+
+Arithmetic keeps working: operators such as `=A1+B1` are not function calls, so nothing gates them.
+`VERSION()` and `OFFSET()` are function calls, but they are protected built-ins that sit outside the
+licence system entirely, so they keep evaluating too. A sheet with a key problem therefore does not
+go blank.
+
+A **valid** key can print one notification too: if it expires on a set date and that date is
+within the notice period your license carries, the console names the last day the key covers. It
+is a heads-up only — nothing is restricted while a key is valid, and the message appears once.
+
```
sequba:
```suggestion
```

src/HyperFormula.ts
```diff
@@ -589,7 +596,22 @@ export class HyperFormula implements TypedEmitter {
...
+   * method [[getRegisteredFunctionNames]] instead, which answers for the engine you actually hold —
+   * the same reasoning that made `getAvailableFunctions()` and `getFunctionDetails()` instance-only
+   * from their first release.
+   *
+   * The two are not interchangeable: this one translates into any registered language without
+   * building an engine, while the instance method answers only under its own instance's language.
+   * Migrating means building an engine:
+   * `HyperFormula.buildEmpty({ language: 'plPL' }).getRegisteredFunctionNames()`.
+   *
+   * @deprecated Use the instance method of the same name. This static method is deprecated and
```
sequba: This PR should not deprecate this function. This decision was never made.

CHANGELOG.md
```diff
@@ -7,6 +7,23 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+
+### Deprecated
+
+- Deprecated the static `HyperFormula.getRegisteredFunctionNames()` method; it will be removed in one of the next major releases. Use the instance method of the same name instead — a static method has no engine, and therefore no configuration, in scope, so it answers for the global function registry: an engine configured with its own `functionPlugins` registers only those, and this method can list functions that engine cannot evaluate at all. The two are not interchangeable: the static one translates into any registered language without an engine, so migrating means building one, for example `HyperFormula.buildEmpty({ language: 'plPL' }).getRegisteredFunctionNames()`. [#1743](https://github.com/handsontable/hyperformula/pull/1743)
+
```
sequba: The changelog should be written from the client's perspective, and the client doesn't care about all these details. Just add a single entry under "Added" about support for new-format license keys.

src/HyperFormula.ts
```diff
@@ -4767,6 +4912,45 @@ export class HyperFormula implements TypedEmitter {
...
+   * Note this checks gate B (entitlement) only, never gate A (key validity). That asymmetry
+   * with the interpreter's gate B - which checks key validity first - is deliberate: it keeps
+   * today's behaviour for a missing or invalid key, where formulas yield `#LIC!` but the CRUD
+   * API keeps working. A later PR that resolves an invalid key to a *restricted* entitlement
+   * rather than an unrestricted one would silently turn that into a breaking API change.
+   *
+   * @internal
+   */
+  private ensureCapability(feature: FeatureId): void {
+    if (!this._config.isLicenseGateActive) {
```
sequba: What do we need this flag `isLicenseGateActive` for?

sequba (changes requested):

Issue #73, opened 2019-12-10
Add support for merged cells


wojciechczerniak:
### Description

![Screenshot 2020-02-28 at 12 29 50](https://user-images.githubusercontent.com/4718998/75545251-16714300-5a26-11ea-92f1-5a4f14dca521.png)

Conclusion: Merged cells are transparent to the calculation engine. The integrator has to remember that it's up to her to clear all the cells other than top left corner cell.

PR #1730, opened 2026-08-11
HF-307: license key reader (entitlement key format, rev 6), resolution, and capability table

marcin-kordas-hoc:

## Update 2026-09-18 — folded #1731, #1736, #1737, #1740, #1741 into this PR

So that no PR in this stack shows code a later PR rewrites, this PR now contains, squashed with
no loss of content: the entitlement reader (rev 6, was #1740), the notice-window warning (was
#1736), the two add-on tokens (was #1737), the function-metadata filtering (was #1731), and both
capability-token dialects (was #1741). Those five PRs are closed, folded here. The stack is now
`#1728 → #1729 → this PR → #1743`.

Two things below are stale as a result and are corrected here rather than edited in place, so the
history stays legible: the title above (fixed), and every PR-number reference in the body below
still says `#1740`/`#1741`/etc. — read those as "now #1730". The CHANGELOG entries in the diff are
already repointed to #1730.

Everything below this line is the PR as it was written across 2026-08-11 through 2026-08-26,
before the fold — the decisions and verification it documents did not change.

---

Third of four. Stacked on PR 2 (#1729), so review the commits on top of it.

**Paired tests: handsontable/hyperformula-tests#32 — merge that first.**

## Why this PR exists at all

**A genuine typed license key did not work before this.** `checkLicenseKeyValidity` recognizes three fixed strings and the older 25-character format; a typed key matches none of them and falls through to `INVALID`. I verified this by building an engine with a real, unexpired `[SUB]` key before touching anything:

```
validityState = "invalid"
C1            = #LIC!   "License key is invalid."
```

So a paying customer pasting a valid key got `#LIC!` in every cell plus a console warning. An entitlement adapter alone would not have helped — gate A kills the formulas first.

## What lands here

**1. Vendor the typed-key reader** into `src/license/handsontable-license-key-parser/` (~~`src/license/vendor/`~~, **renamed 22.09** at the reviewer's request so the directory names the repository it mirrors) as TypeScript (`allowJs` is off, `strict` is on, so this is a port, not a copy). This follows the key spec's own recommendation of a vendored copy plus a drift check rather than a shared package: a private dependency would break `npm install` for open-source users of this GPL package.

`PROVENANCE.md` records the upstream commit and a per-file sha256, so drift is detectable by re-cloning and re-hashing, and lists the deliberate divergences.

**2. Resolve typed keys into both gates.** `resolveLicense` reads the key once and answers both gates from that single reading, so they cannot disagree about what the string says. Anything that is not a typed key — `gpl-v3`, an older-format key, an empty string, a typed key with a broken checksum — falls through to `checkLicenseKeyValidity` with its **verdict unchanged**. That function is extracted from rather than rewritten, so both paths share one warning-per-page flag instead of each getting their own. Two additive edits to the file it lives in, both affecting the legacy path too and both deliberate: `formatDate` now reads its date in UTC (it printed the expiry a day early west of UTC — a latent bug on the legacy path as well), and an `@internal` reset for the warn-once flag was added because it made the console path untestable.

**3. The real capability table**, plus reading both payload shapes.

## Please confirm two things

**The gate-A/gate-B invariant.** Only a VALID typed key resolves to a restricted entitlement; missing, invalid and expired all resolve to `unrestrictedEntitlement()`. Gate A already stops formula evaluation on its own — letting a bad key restrict the entitlement *as well* would make PR 2's `ensureCapability` throw from the CRUD API, turning today's "formulas fail, the API still works" into a silent breaking change for every user whose key lapsed. Mutation-tested.

**The capability table is a DRAFT.** Membership is transcribed from the packaging design's own per-function evidence file rather than invented, and the transcript is checked two ways: it reproduces that file's five published counts exactly (370 rows; 17 / 51 / 127 / 355 cumulative plus 15 operators), and all 423 registered function ids resolve into the 370 canonical entries once the 53 declared aliases are canonicalised, with zero uncovered. But the design is still under review — the free tier's exact contents and the placement of several function families are not settled. Landing it now was a deliberate call, not a claim that it is final.

## Judgement calls worth a reviewer's eye

- **Both payload shapes are read**, detected per product entry by the presence of `capabilities`: the shipped shape (`tier`/`addons`/`exp`) and the newer specified shape (`capabilities`/`usage_until`/`release_until`/`notice`/`flags`). They disagree about nearly every field, the newer one is still under review, and only the first can be minted today. The newer spec also contradicts itself on whether its dates are `YYYY-MM-DD` strings or numeric timestamps, so both are accepted.
- **Feature gating is REAL** (updated 12.08, per Kuba's answers on the task). An earlier revision of this PR granted all five feature areas from the `core` token, which made feature gating inert by construction. The five features now live on their own `feat:*` tokens: a rev-5 key grants exactly the `feat:*` tokens it carries; the shipped shape — whose vocabulary predates feature tokens — is granted all five by the adapter, so a shipped-shape key's API behaviour is unchanged; legacy keys resolve unrestricted (Kuba's carve-out). The exits stay ungated (`resumeEvaluation`, teardown), per the never-gate-the-exit rule.
- **An unrecognized token no longer silences the key** (updated 12.08). Silence comes solely from the key's `flags`; the earlier coupling (`silent || unrecognized > 0`) suppressed strictly more than D3 asks for and was confirmed an implementation error by Kuba.
- **The two add-on tokens are recognized but empty.** `spreadsheet` has no agreed content — the packaging design names a *package* "Spreadsheet" and the pricing work names a "Spreadsheet Bundle" *add-on*, and whether those are the same set is unsettled; guessing either way silently sells an empty add-on or duplicates a whole tier. Pinned by a test so filling it has to be deliberate.
- **`noticeDays` is 0 for the shipped shape**, which has no notice field. Inventing a default in the parser would put a product decision in the wrong place. The newer shape carries `notice` and it is used.
- **`PROVENANCE.md` names the upstream private repo and commit.** That is what makes the documented drift check runnable. Ratified by Kuba on the task (D7-A, 12.08): kept as-is.

## Update 13.08 — key spec rev 5 read against the code

Reading rev 5 ([doc](https://app.clickup.com/9015210959/docs/8cnjcyf-31675/8cnjcyf-48155), updated 12.08) and running its own example payload through this branch turned up three things, all fixed in the last commit:

- **Feature tokens are now OPT-IN.** §2.2 lists HyperFormula's entire token vocabulary as `functions_1..4`, `spreadsheet`, `import_export` — there is no `feat:*` entry at all. So a key that names no feature token cannot be saying "no features"; nothing in circulation can express one. Minting the spec's own §2 payload and running it: `setCellContents`, `addRows`, `copy`, `undo`, `addNamedExpression` and `batch` **all threw**. The same hole hit Handsontable-only keys and keys with `hyperformula: null`, which lost the API that `core` used to give them, silently (gate-A VALID = no console warning). A key that *does* name a `feat:*` token still gets exactly what it names, which is the gating Kuba asked for.
- **`no-console-warns` is honoured.** rev 5 spells the flag three ways — §2.3 and the §2 example say `no-console-warns`, §4.3/§5.2 say `silent-console`, earlier revisions said `silent`. Only the last two were recognised, so a doc-conformant SaaS key printed console warnings it had explicitly asked to suppress.
- **An unreadable `capabilities` rejects the key.** Present-but-not-an-array fell through to the shipped-shape branch, which was a free pass twice: every feature granted, and the rev-5 dates never read — a subscription expired in 2020 resolved as perpetual.

**Still divergent from rev 5, deliberately:** §4.1 says a non-trial key never hard-blocks, even past grace ("Trial: block. Non-trial: error only (18.1)"). This branch keeps EXPIRED → `#LIC!`, per Kuba's D5=A (hard blocking stays this release; the notice/soft-stop/hard-stop model is a follow-up).

## Verification

Full unit suite green — **512 suites, 6344 tests** (after the 13.08 update). Each of the three fixes above is mutation-verified: removing the opt-in rule fails 9 tests, removing `no-console-warns` fails 1, removing the `capabilities` guard fails 2, and cutting the console notification fails 1 (it used to fail none). Beyond that, the parts that matter were mutation-tested rather than trusted: a wrong SHA-512 round constant, a one-bit UTF-8 error, a broken invariant and a moved package boundary each fail the tests that are supposed to catch them.

---

> [!NOTE]
> **High Risk**
> Changes core licensing, formula evaluation (`#LIC!`), and API capability gating for all proprietary keys; incorrect resolution or table drift could block paying customers or expose paid features.
> 
> **Overview**
> Introduces **partial-library licensing** for proprietary keys: entitlement keys (rev 6) are parsed locally via a vendored reader, resolved once in `resolveLicense`, and drive two gates—formula validity (`#LIC!` when the key is bad) vs. capability entitlements (missing functions/features).
> 
> **Capability model:** Replaces the placeholder `core` grant with a static table keyed by normalized tokens (`fun:all`, `fun:<family>.<A|B|C>`, per-function `fun:<NAME>`, `feat:*`, and add-ons `spreadsheet` / `import_export`). Keys that name no recognized `feat:*` token still get all five API feature areas (backward compatible for today’s keys). Invalid/missing/expired keys resolve to an **unrestricted** entitlement so the CRUD API keeps working while gated functions fail in cells.
> 
> **Product-facing behavior:** `getAvailableFunctions()` / `getFunctionDetails()` omit functions the **valid** key does not include; bad keys still list the full catalogue. `#LIC!` docs cover “not in your package.” Console warnings distinguish usage vs. release expiry, format dates in UTC, and add a one-time “valid until …” notice inside the usage notice window.
> 
> **Config:** `Config` uses `resolveLicense` instead of legacy-only validation; `mergeConfig` / rebuild can silence license console noise for throwaway configs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 258786ed7354e51510b440a167d15fb18549b24f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## ⛔ Update 18.08 — upstream retired this key format the same day, and one claim above is wrong

A scoped spec-to-ship review of this PR (5 dimensions, every finding adversarially verified) confirmed 5 findings. Four are test-hardening and are fixed in the paired PR — see [hyperformula-tests#32](https://github.com/handsontable/hyperformula-tests/pull/32)'s "Update 18.08". The fifth needs a decision, not a patch, so nothing here has been changed for it yet.

**The vendored reader cannot read the format upstream now ships.** `handsontable/license-key` commit `01eae6530d4f` (2026-08-18 08:44 UTC, DEV-2512, "Replace typed license keys with the entitlement key format (rev 6)"), released as **4.0.0**, **deletes `src/typed-key/` outright** — the exact five files `PROVENANCE.md` pins and hashes. Its own docs are explicit: "The typed key format was **removed**, not deprecated… this package cannot read a typed key." The replacement differs in three ways this reader gates on:

| | this branch | entitlement format (rev 6 / 4.0.0) |
|---|---|---|
| format marker | requires a leading `[SUB]_`/`[FREE]_`/`[TRIAL]_`/`[PERP]_` tag (`extractKeyData.ts:147`) | **no tag** — the key *ends* with a bracketed block, which is the only marker |
| payload version | requires `v: 1` (`extractKeyData.ts:199`) | **no `v` field at all** — payload is exactly `{"products":{…}}` |
| checksum domain | SHA-512 over the whole key body, prose included | over the **encoded payload only** — an explicit reversal, so the prose becomes editable |
| unknown product | invalidates the key (`extractKeyData.ts:105`) | tolerated — "kept, and ignored" |

Measured, not inferred: an authentic 4.0.0-format key built per upstream's own `build-payload.js` gives `extractTypedKeyData(key)` → `null`, `resolveLicense(key).validityState` → `invalid`, and `C1` → `#LIC!` "License key is invalid.". Passing only the `[…]` block behaves the same. The retired tagged format still resolves correctly on the same build — so this is a format gap, not a robustness one.

Independently, Budzio published the byte-level spec for that format this morning as a new page under key-spec rev 6 ("Technical implementation", sections `T1`–`T8`, created 09:45 UTC). It matches upstream and confirms all four rows above. Three of its rules this branch **already satisfies**: checksum verified before any decode or parse ("order is normative"), uppercase hex rejected rather than normalised, and `__proto__`-safe result construction.

**What is genuinely at stake, and what is not.** The invariant holds throughout: a rev-6 key resolves INVALID and therefore to `unrestrictedEntitlement()`, so the CRUD API keeps working and only formulas fail (verified). And myHOT PR #73 (merged to `master` 12:37 UTC today, pinning `license-key#4.0.0`) states **"the typed keys never reached production"**, with the legacy 25-character keys still generated in parallel and validating exactly as before — that path is untouched here. So the correction owed to this PR's own opening claim: no customer ever held a typed key, so "a paying customer pasting a valid key got `#LIC!`" describes a format that was never issued. The live exposure is narrower but real — per that PR's surfacing table, **trial orders and the HubSpot trial webhook now surface the entitlement key**, while commercial orders and transactional emails still show the legacy key only.

**Why this is not fixed in this PR.** Re-porting is the obvious move and upstream made it cheaper than the original port (`sha512.js` and `utils.js` are unchanged; the new `extractEntitlementKeyData` is deliberately schema-free and built to be vendored, so only the extractor, schema and constants need re-porting). But it is not a patch to slip in: adopting the new shape rules verbatim also makes `capabilities` mandatory and **invalidates every payload shape this branch currently reads**, rev 6 is still status "for review", and the work is arguably [HF-329](https://app.clickup.com/t/86cb119tz)'s (unassigned, due 21.08, untouched since 04.08 — it now has concrete content). Raised on the task as **D9**; the merge freeze until the spec owner returns means escalating is the action, not a delay.

Also worth a reviewer's eye, since it sharpens the still-unanswered **D8**: rev 6's own examples use `capabilities: ["functions_1", "spreadsheet"]` / `["core"]` — i.e. **exactly the token names this branch implements** — while the packaging design published 12.08 states its vocabulary is `fun:<family>.<A|B|C>` and that "nothing else about packaging exists at the technical layer". The two documents disagree, and this branch conforms to the key spec rather than to the packaging doc. A key minted in the packaging vocabulary resolves VALID with the whole gated API and **zero functions**, silently (gate A says VALID, so nothing prints, and `unrecognizedCapabilities` is exposed nowhere). That outcome is now pinned by a test in the paired PR — pinned, not endorsed, so whichever way D8 lands has to change it.

---

### Changed in the 2026-09-22 review round

- The guide no longer describes the key formats, and the validation section points at the terms of the contract instead of enumerating the two date axes. The changelog entry for the parser change was reworded the same way, since it carried the same description.
- `PROVENANCE.md`: two paragraphs about the superseded `src/typed-key/` port removed; the divergence list corrected. Its item 3 described *upstream's* date check as if it were this port's, which has not been true since a type guard was added, and that guard — the only divergence that changes which keys are accepted — was missing from the list entirely.
- The expired-key console message no longer blames the installed version when the key ran out on the usage axis. That wording predates this axis and was correct for every key `develop` can see; it is not for a subscription key.
- The `implies` field and its recursive expansion are gone (see #1728), so no capability token refers to another.


qunabu:

marcin-kordas-hoc:
bugbot run

marcin-kordas-hoc:
bugbot run

src/license/capabilities.ts
```diff
@@ -20,37 +38,122 @@ export interface CapabilityGrant {
...
+const SPREADSHEET_FUNCTIONS = [
+  'ADDRESS', 'AVERAGEIF', 'CHOOSE', 'COLUMN', 'COLUMNS', 'COUNTIF', 'COUNTIFS', 'DATEVALUE', 'DAYS360',
+  'EDATE', 'EOMONTH', 'FILTER', 'FORMULATEXT', 'FV', 'HLOOKUP', 'HYPERLINK', 'IFERROR', 'IFNA', 'INDEX',
+  'INTERVAL', 'IPMT', 'IRR', 'ISBINARY', 'ISBLANK', 'ISERR', 'ISERROR', 'ISEVEN', 'ISFORMULA', 'ISLOGICAL',
+  'ISNA', 'ISNONTEXT', 'ISNUMBER', 'ISODD', 'ISOWEEKNUM', 'ISREF', 'ISTEXT', 'MATCH', 'MAXIFS', 'MINIFS',
+  'NA', 'NETWORKDAYS', 'NETWORKDAYS.INTL', 'NPER', 'NPV', 'PERCENTILE.EXC', 'PERCENTILE.INC', 'PMT',
+  'PPMT', 'PV', 'RAND', 'RANDBETWEEN', 'RATE', 'ROW', 'ROWS', 'SORT', 'STDEV.P', 'STDEV.S', 'STDEVA',
+  'STDEVPA', 'SUMIF', 'SUMIFS', 'TIME', 'TIMEVALUE', 'UNIQUE', 'VAR.P', 'VAR.S', 'VARA', 'VARPA',
+  'VLOOKUP', 'WEEKDAY', 'WEEKNUM', 'WORKDAY', 'WORKDAY.INTL', 'XLOOKUP', 'YEARFRAC',
+]
```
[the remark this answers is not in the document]
marcin-kordas-hoc: Checked and not a defect — the gap is exactly the two protected built-ins.

Measured on the current branch: `functions_1` = 16, `functions_2` = 50, `functions_3` = 125, against the documented 17 / 51 / 127. The differences are 1 / 1 / 2, and they are `VERSION` (counted in math engine) and `OFFSET` (counted in spreadsheet) — both `FunctionRegistry.functionIsProtected() === true`.

Protected built-ins sit **outside** the token system: the interpreter never gate-checks them, so listing them in a package would imply a restriction that does not exist. `capability-table.spec.ts` pins this explicitly as `17 - 1`, `51 - 1` and `127 - 2` rather than as raw numbers, so the relationship is asserted rather than assumed.

One part of this finding I have **not** been able to re-verify: the claim that `INT` belongs in the math-engine set. `INT` currently sits in `functions_4`. Which package owns it is settled by the HF-306 evidence CSV, and I read that file as placing it in excel-simulator — but I cannot re-open it right now, so I am flagging this rather than asserting it, and it will be confirmed against the CSV before release.

Apologies for the four-day silence on this thread: I had assessed it but never wrote the assessment here.

src/license/capabilities.ts
```diff
@@ -20,37 +37,152 @@ export interface CapabilityGrant {
...
+const functions3Grant: CapabilityGrant = {
+  functions: [...MATH_ENGINE_FUNCTIONS, ...CALCULATED_FIELDS_FUNCTIONS, ...SPREADSHEET_FUNCTIONS], features: [],
 }
+const functions4Grant: CapabilityGrant = {
+  functions: [
+    ...MATH_ENGINE_FUNCTIONS, ...CALCULATED_FIELDS_FUNCTIONS, ...SPREADSHEET_FUNCTIONS,
+    ...EXCEL_SIMULATOR_FUNCTIONS,
+  ],
+  features: [],
+}
```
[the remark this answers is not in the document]
marcin-kordas-hoc: Same finding as the thread on `capabilities.ts:118` — answering here too so neither is left open.

The arrays really do total 16 / 50 / 125 against a documented 17 / 51 / 127, and the two missing entries really are from the evidence file. They are `VERSION` and `OFFSET`, and both are **protected built-ins** (`FunctionRegistry.functionIsProtected() === true`), which the interpreter never gate-checks. Listing them in a package would imply a restriction that cannot exist, so they are deliberately absent — and `capability-table.spec.ts` asserts the counts as `17 - 1`, `51 - 1`, `127 - 2` so the subtraction is pinned rather than hand-waved.

So: no drift between the table and the packaging design. The file comment could have been clearer that "an exact transcript" means "of the non-protected entries", which is a fair reading of your report.

src/license/licenseResolution.ts
```diff
@@ -0,0 +1,310 @@
...
+
+  if (now === null) {
+    return {state: LicenseKeyValidityState.VALID}
+  }
+
+  const deadline = terms.expiryTimestamp + MILLISECONDS_PER_DAY + (terms.graceDays * MILLISECONDS_PER_DAY)
+
+  return now < deadline
+    ? {state: LicenseKeyValidityState.VALID}
+    : {state: LicenseKeyValidityState.EXPIRED, expiredOn: new Date(terms.expiryTimestamp)}
```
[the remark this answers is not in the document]
marcin-kordas-hoc: Correct at the time, and the contradiction was resolved by changing the documentation rather than the code.

`LicenseExpiry` did describe `kind: 'usage'` as compared against the client's LOCAL calendar date, while `resolveLicense` compared in UTC. The JSDoc was the part that was wrong: key spec rev 5 §1.2 requires the offline check and a future online check to return the same verdict for the same key at the same instant, and any rule reading a local clock breaks that parity. UTC is therefore the intended semantics, and the doc now says so, including the practical cost — a customer far west of UTC loses the tail of their last local day.

GitHub has since marked this thread outdated because the code moved. Resolving it.

src/license/licenseResolution.ts
```diff
@@ -0,0 +1,375 @@
...
+ *
+ * Read from the same `HT_RELEASE_DATE` (`DD/MM/YYYY`) the legacy validator uses, so a perpetual
+ * typed key and a legacy key agree on what "this build" means.
+ */
+function releaseDateTimestamp(): number | null {
+  const [day, month, year] = (process.env.HT_RELEASE_DATE ?? '').split('/')
+  const timestamp = Date.UTC(parseInt(year, 10), parseInt(month, 10) - 1, parseInt(day, 10))
+
+  return isNaN(timestamp) ? null : timestamp
+}
```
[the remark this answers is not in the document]
marcin-kordas-hoc: You are right, and this one had a real defect in it — not in the code, but in the comment that described it. Fixed in 05882e11.

Reproduced at process level, using each path's own parse of the same `HT_RELEASE_DATE=10/08/2026`:

```
                                  legacy (local)   typed (Date.UTC)
  TZ=UTC, TZ=America/Los_Angeles      20675            20675      agree
  TZ=Asia/Tokyo                       20674            20675      differ by a day
  TZ=Pacific/Kiritimati               20674            20675      differ by a day
```

So the comment's promise that "a perpetual typed key and a legacy key agree on what 'this build' means" was simply false east of UTC. It has been replaced with the measurements above and an explicit statement of the consequence: two customers east of UTC, one on a legacy key and one on an equivalent typed key, can disagree by a day about whether this build is covered.

**The code stays as it is**, deliberately. `Date.UTC` is required for the typed path by rev 5 §1.2 (offline/online parity). The legacy path keeps its local parse because legacy behaviour is frozen for this release — switching it would move the expiry verdict of already-issued legacy keys by a day for every customer east of UTC, which is a product decision rather than a refactor. It is flagged as such for the release discussion.

No regression test accompanies it, and that is deliberate rather than an omission: assigning `process.env.TZ` mid-run has no effect once the runtime has resolved its timezone — probed inside Jest, `UTC`, `Asia/Tokyo` and `Pacific/Kiritimati` all returned an identical timestamp — and CI runs in UTC, where both parses agree by definition. I wrote such a test, mutated the source to the legacy parse, watched it still pass, and removed it rather than ship an assertion that cannot fail. Closing it properly needs a timezone-parameterised CI job; the reasoning is recorded next to the release-axis tests.

qunabu:
Task linked: [HF-329 Import license key validator shared code to HF](https://app.clickup.com/t/9015210959/HF-329)

src/license/licenseResolution.ts
```diff
@@ -0,0 +1,456 @@
...
+  // CORE_TOKEN is always granted, but note what it actually grants: the calculation operators -
+  // NOT a usable set of functions. A key whose only tokens this build does not recognize therefore
+  // evaluates operators and protected built-ins and returns #LIC! for every function call,
+  // silently, per HF-307 decision D3. Kuba ratified that cliff as-is on 12.08 (D6-A): "this
+  // situation should never happen. There is no point in issuing a key if empty capabilities."
+  const capabilityTokens = [CORE_TOKEN]
+
+  if (hyperformulaGrant !== undefined) {
+    if (isRev5) {
+      capabilityTokens.push(...readStrings(hyperformulaGrant.capabilities))
```
Tobiadefami: This array is payload-controlled and the format defines no size limit. At this head, a checksum-valid payload with 125,000 capability strings makes `HyperFormula.buildFromArray` throw `RangeError: Maximum call stack size exceeded` at this spread instead of producing a license verdict. The same argument-expansion risk exists for `addons` at line 266. Appending the strings one at a time made the reproducer pass; please change both sites and add a large-array regression in the paired private suite.

Tobiadefami (dismissed):

src/license/licenseResolution.ts
```diff
@@ -0,0 +1,456 @@
...
+    return {state: LicenseKeyValidityState.VALID}
+  }
+
+  const deadline = terms.expiryTimestamp + MILLISECONDS_PER_DAY + (terms.graceDays * MILLISECONDS_PER_DAY)
+
+  return now < deadline
+    ? {state: LicenseKeyValidityState.VALID}
+    // The reported day is the first day NOT covered, which is the convention the legacy validator
+    // already uses for the same message (it reports `keyValidityDays + 1`).
+    : {state: LicenseKeyValidityState.EXPIRED, expiredOn: new Date(terms.expiryTimestamp + MILLISECONDS_PER_DAY)}
```
Tobiadefami: `deadline` correctly includes `terms.graceDays`, but `expiredOn` drops that interval. For `usage_until: 2020-01-01` with `grace: 1`, the key remains covered through January 2 and first becomes invalid on January 3, yet this path reports January 2. The discrepancy grows with the grace window and contradicts this function's “first day not covered” contract. Please include the grace interval when constructing `expiredOn` and add a positive-grace regression.

Tobiadefami (commented):
Reviewed at ab1bba15 together with the paired tests at 9977ea88. The typed-key reader, license resolution, capability mapping, legacy-key invariant, and current automated findings were checked. The six changed paired suites pass (131 tests), and the current engine checks are green. I found one material issue: the expired-key console date ignores the key's grace period, as noted inline.

src/license/licenseResolution.ts
marcin-kordas-hoc: Right, and the asymmetry was the bug: `deadline` added the grace interval, `expiredOn` reported the licensed day itself, so a key inside its grace window described itself as already expired. Fixed in f63b3bde8 — `expiredOn` now reports the first day the key does *not* cover, i.e. `expiry + graceDays + 1`. Your `usage_until: 2020-01-01` case is pinned with `grace: 30` (tests commit 8ee492c), so the two candidate days sit a month apart - the warning must name `February 1, 2020` and must not name `January 2, 2020`.

docs/guide/license-key.md
```diff
@@ -29,25 +40,94 @@ const options = {
...
+Your proprietary license key is in one of two formats, and both work the same way:
+
+* A classic key: 25 characters in five dash-separated groups, for example
+  `1a2b3-4c5d6-7e8f9-0a1b2-3c4d5`.
+* An entitlement key: a short, human-readable license text that ends with a machine-readable
+  block in square brackets. Assign the whole text to the `licenseKey` option, or just the
+  bracketed block — the block is the only part HyperFormula reads, so both work. The text around
+  the block may be re-wrapped on its way to you (for example, by an email client) without
+  affecting the key; the block itself has to arrive character for character.
+
```
sequba: Don't mention it in the public docs.

docs/guide/license-key.md
```diff
@@ -29,25 +40,94 @@ const options = {
...
 ### Proprietary license key validation
```
sequba, line 43: This section should say that the validation mechanism depends on the terms of your contract.

src/helpers/licenseKeyValidator.ts
```diff
@@ -42,6 +43,106 @@ const consoleMessages: ConsoleMessages = {
...
+ * expiry-approaching notice.
+ *
+ * Deliberately keyed per key rather than a single boolean like {@link _notified}
+ * above: that flag reports one of a handful of states that mean the same thing regardless of
+ * which key triggered them ("a key is invalid", "a key is missing"), so once-per-page-load is the
+ * right behaviour for it. Two different keys approaching their OWN expiry are two different
+ * events, and a page that swaps keys (or a test suite that builds one engine per key) must still
+ * warn for the second one even though the first already consumed a shared flag.
+ */
+const _noticedKeys = new Set<string>()
```
sequba, line 69: Why it is per key? Why do we need to manage more keys in single instance of HF?

src/license/handsontable-license-key-parser/PROVENANCE.md
```diff
@@ -0,0 +1,100 @@
+# Vendored entitlement-key reader — provenance and drift control
```
sequba, line 1: Let's rename the directory `vendor` to `handsontable-license-key-parser`

src/license/vendor/PROVENANCE.md
```diff
@@ -0,0 +1,100 @@
...
+| Repository | `handsontable/license-key` (private) |
+| Tag | `4.0.0` |
+| Commit | `c50ef40a6` (the `4.0.0` release commit; on `develop` as `1acddafa8`) |
+| Ported on | 2026-08-20 |
+| Reference docs | the format and design notes kept alongside the upstream sources; the byte-level rules are also specified in the key spec's "Technical implementation" addendum (T1–T14) |
+
+This replaces the earlier port of `src/typed-key/` at `7553d0d1` (2026-08-11). Upstream 4.0.0
+(DEV-2512) **deleted** that directory and replaced the tagged key format with the entitlement key
+format; the tagged format never reached customers, so the old reader was removed here rather than
+kept alongside.
```
sequba: Remove this paragraph

src/license/vendor/PROVENANCE.md
```diff
@@ -0,0 +1,100 @@
...
+|---|---|---|
+| `constants.ts` | `constants.js` | `6e2ad68d1a316abdec3f89bf04260a2cc4098f76525427d919f22cb25fb077d6` |
+| `detectFormat.ts` | `detect-format.js` | `7dc037fd70e7c64078a0fe29b42cb33ecf25e8f4d8963ae9bfb16b69479d267f` |
+| `extractKeyData.ts` | `extract-key-data.js` | `afd0858768879764ea016d2bc4fca692a0cd12214c1dfda6932d7ed9e4f32e45` |
+| `utils.ts` | `utils.js` | `135a8396bb22f424160fc651e899931d4be807df9b94c6dd24bb1cf6526e0541` |
+| `sha512.ts` | `sha512.js` | `668dd1109160b92965a1f9a9c5fb78dfdc1e5b7e93f635a147ae8a6bb2a5d837` |
+
+`utils.js` and `sha512.js` are byte-identical between `src/typed-key/` at the old pin and
+`src/entitlement-key/` at `4.0.0` (same hashes as the previous revision of this table), so their
+ports carried over unchanged apart from this file's path references.
```
sequba: Remove this paragraph

src/license/handsontable-license-key-parser/PROVENANCE.md
```diff
@@ -0,0 +1,100 @@
...
+| `validate-key.js` | A two-line boolean wrapper over `extractEntitlementKeyData`; the extractor is called directly. |
+
+From `utils.js`, the two generation-side helpers `bytesToBase64` and `stringToBase64Url` are
+also left out. Everything else in that file is ported.
+
+## Deliberate divergences from upstream
+
+`allowJs` is off in HyperFormula's `tsconfig.json` and `strict` is on, so these files are a port
+rather than a copy. Beyond adding types, the semantics were kept identical except for the
+following, which a drift review should expect to see:
```
sequba, line 65: Can we add exceptions to tsconfig so these files could be the byte-identical copies of the upstream?

src/license/capabilities.ts
```diff
@@ -20,37 +93,282 @@ export interface CapabilityGrant {
...
+  ['fun:array.c', ['ARRAYFORMULA', 'ARRAY_CONSTRAIN']],
+])
+
+/** The group tokens each package adds, exactly as the packaging doc's §4 table states them. */
+const MATH_ENGINE_GROUPS = ['fun:math.a', 'fun:stat.a', 'fun:logic.a', 'fun:operator.a', 'fun:info.a', 'fun:lookup.a']
+const CALCULATED_FIELDS_GROUPS = ['fun:time.b', 'fun:text.b', 'fun:logic.b', 'fun:math.b', 'fun:stat.b']
+const SPREADSHEET_GROUPS = [
+  'fun:lookup.c', 'fun:math.c', 'fun:stat.c', 'fun:time.c', 'fun:text.c', 'fun:info.c', 'fun:logic.c',
+  'fun:finance.c', 'fun:engineer.c', 'fun:array.c',
+]
```
sequba, line 184: These groups should not be hardcoded. The code should operate on a capabilities level and should know nothing about what capabilities are in what package.

src/license/capabilities.ts
```diff
@@ -20,37 +93,282 @@ export interface CapabilityGrant {
...
+  // moving it into the bundle stays a product call rather than a silent edit.
+  [SPREADSHEET_ADDON_TOKEN, {
+    functions: [],
+    features: [FeatureId.Crud, FeatureId.UndoRedo, FeatureId.Clipboard, FeatureId.Batching],
+  }],
+  [IMPORT_EXPORT_ADDON_TOKEN, {functions: [], features: [FeatureId.ImportExport]}],
+  [FUN_ALL_TOKEN, {functions: [...functions4Grant.functions], features: []}],
+  ...groupEntries,
+  ...singleFunctionEntries,
+])
```
sequba, line 373: In this file there are a lot of things that don't belong in HF repo

sequba (changes requested):

docs/guide/license-key.md
marcin-kordas-hoc: @sequba Done — the whole "Proprietary license key formats" section is out of the guide.

I also applied the same reasoning to the changelog, since it is the same disclosure in another public file: the entry for this PR described the format in the same words ("a human-readable text ending with a machine-readable block in square brackets"). It now names the change and what is unaffected, without describing the format. Say the word if you would rather have that wording back.

docs/guide/license-key.md
marcin-kordas-hoc: @sequba Done. The section no longer enumerates the two date axes; it says that which versions a key covers, and for how long, follows from the terms of your contract, and that HyperFormula applies them locally. I kept the existing tip about no internet connection, since that was not what you objected to.

Worth knowing: this section predates HF-307 on `develop` and this PR had expanded it. The rewrite takes it back past the original wording rather than only undoing the expansion.

src/license/vendor/PROVENANCE.md
marcin-kordas-hoc: @sequba Done — removed.

src/license/vendor/PROVENANCE.md
marcin-kordas-hoc: @sequba Done — removed.

src/license/handsontable-license-key-parser/PROVENANCE.md
marcin-kordas-hoc: @sequba Done. `src/license/vendor/` is now `src/license/handsontable-license-key-parser/`, moved with `git mv` so blame and `--follow` walk through it, and the paired test directory is renamed to match. Only paths changed; the prose about a "vendored reader" stays, since it describes the relationship rather than the location.

src/license/handsontable-license-key-parser/PROVENANCE.md
marcin-kordas-hoc: @sequba I would like this too, and it is worth more than tidiness: byte-identical files turn drift control from "re-read and re-port the changed file" into `sha256sum`, which is the whole point of the hash table in that document.

But it is not just `allowJs`. The files would have to be emitted and bundled as `.js` inside a `strict` build that also emits declarations, which means either `checkJs: false` plus a hand-written `.d.ts` per file, or `@ts-check`-clean upstream sources. I have not measured whether the rollup and declaration steps survive that.

Before I spend that: is the goal (a) a mechanical drift check, or (b) literally no local edits? If (a), there is a cheaper route — keep the port and add a CI job that clones the upstream at the pinned commit and compares. §7.1 of the spec asks for a drift check in CI anyway, and the document currently admits the check is manual.

One thing that collides either way: there is a deliberate local divergence in `extractKeyData.ts` (see my separate comment on the divergence list). A byte-identical copy cannot keep it.

src/helpers/licenseKeyValidator.ts
marcin-kordas-hoc: @sequba Small correction to the premise first: `_noticedKeys` is a module-level `const`, not instance state. It is not one engine holding several keys — it is several engines on one page load.

The case it exists for is a page that swaps the key, or a suite that builds one engine per key. Two keys approaching their own expiry are two different events, and with a single boolean the second one is silently swallowed because the first consumed it. The flag above it stays a single boolean on purpose, because "a key is invalid" means the same thing whichever key caused it.

If you would rather have one notice per page load regardless, say so and it collapses to a boolean — it is a two-line change. I only wanted the reason on record either way.

src/license/capabilities.ts
marcin-kordas-hoc: @sequba On the group constants you are simply right: `MATH_ENGINE_GROUPS` / `CALCULATED_FIELDS_GROUPS` / `SPREADSHEET_GROUPS` are the packaging table transcribed into the engine — the comment above them literally says "exactly as the packaging doc's §4 table states them". That belongs to whatever mints keys, not here.

Your broader point I want to scope before acting on, because it touches a decision we recorded in August (D8): *the HyperFormula token registry belongs to us* — rev 6 §2.2, and the shipped 4.0.0 schema declares `hyperformula: []` with the whitelist off. Two readings:

- **(a)** the engine knows *token → functions* but nothing about *package → tokens*. Then the three constants and the `functions_1..4` grants go, and the `fun:<family>.<tier>` table stays.
- **(b)** the engine knows nothing, and a key carries function ids directly. That retires the group-token vocabulary and reverses D8.

Which one? I would rather reopen D8 deliberately than by accident. This also gates the `unrestricted` reshape, since both rewrite the same file.

src/license/capabilities.ts
marcin-kordas-hoc: @sequba On the group constants you are simply right: `MATH_ENGINE_GROUPS` / `CALCULATED_FIELDS_GROUPS` / `SPREADSHEET_GROUPS` are the packaging table transcribed into the engine — the comment above them literally says "exactly as the packaging doc's §4 table states them". That belongs to whatever mints keys, not here.

Your broader point I want to scope before acting on, because it touches a decision we recorded in August (D8): *the HyperFormula token registry belongs to us* — rev 6 §2.2, and the shipped 4.0.0 schema declares `hyperformula: []` with the whitelist off. Two readings:

- **(a)** the engine knows *token → functions* but nothing about *package → tokens*. Then the three constants and the `functions_1..4` grants go, and the `fun:<family>.<tier>` table stays.
- **(b)** the engine knows nothing, and a key carries function ids directly. That retires the group-token vocabulary and reverses D8.

Which one? I would rather reopen D8 deliberately than by accident. This also gates the `unrestricted` reshape, since both rewrite the same file.

marcin-kordas-hoc:
@sequba One more thing in the vendored parser, found while acting on your comments rather than from your review — flagging it because it is the only divergence from upstream that changes *which keys are accepted*.

`isIsoDate` in `extractKeyData.ts` rejects a `usage_until` that is not a string. Upstream matches `String(value)` against the date shape, so a single-element array of the right string passes there and is carried into a RESTRICTED entitlement; here it takes the invalid-key path instead.

Two problems with how that was documented, both now fixed in `PROVENANCE.md`:

- The "Deliberate divergences" list did not mention this guard at all, even though the document promises a drift review will find every divergence there.
- Item 3 described *upstream's* behaviour as if it were this port's — it said a `usage_until` array "passes", which has not been true here since the guard was added. A false statement about the code in the same directory.

On the substance: the key spec's addendum (T7) makes that field a real calendar date, so the stricter reading is the specified one and it is upstream that deviates. Upstream has not adopted it — checked at `c50ef40a`, `develop` and `master`, and there is no PR or issue proposing it. Worth raising on that side; until then it is a live fork of behaviour rather than a re-port waiting to happen.

CHANGELOG.md
```diff
@@ -7,6 +7,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+
+### Fixed
+
+- Fixed the `AVERAGEIF` function returning a division-by-zero error when the calculated average was `0`. [#1733](https://github.com/handsontable/hyperformula/pull/1733)
+- Fixed the localized names of `VSTACK` and `HSTACK` in 14 language packs to match Microsoft Excel. [#1748](https://github.com/handsontable/hyperformula/pull/1748)
+- Fixed the MAXPOOL and MEDIANPOOL functions throwing an uncaught `TypeError` instead of returning the `#VALUE!` error when the range dimensions are not a whole multiple of the window size and the stride. [#1718](https://github.com/handsontable/hyperformula/pull/1718)
+- Fixed the `MOD` function returning a remainder with the sign of the dividend instead of the sign of the divisor, which made the results differ from Excel and Google Sheets for arguments with opposite signs (e.g. `=MOD(-3, 12)` now returns `9` instead of `-3`). [#1747](https://github.com/handsontable/hyperformula/issues/1747)
```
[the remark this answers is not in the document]
marcin-kordas-hoc: Not applicable to this PR. That `MOD` line is already on `develop` — CHANGELOG.md line 15, introduced by #1748 (`50e7170cd`). It appears in this diff as context, not as a change this PR makes, so the PR-link convention has nothing to correct here.

src/license/licenseResolution.ts
```diff
@@ -0,0 +1,391 @@
...
+    }
+
+    return {validityState: LicenseKeyValidityState.INVALID, entitlement: unrestrictedEntitlement()}
+  }
+
+  const terms = licenseTermsOf(data)
+  const {state, expiredOn} = validityOf(terms)
+
+  if (notifyConsole && !terms.silent) {
+    notifyLicenseKeyState(state, expiredOn)
```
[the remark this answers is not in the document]
marcin-kordas-hoc: Fixed. You are right, and it is a regression this stack introduced rather than an old wording: on `develop` this file does not know the usage axis at all, so the single "…and is not valid for the installed version" message was correct for every key it could see.

Now the axis reaches the message. A maintenance key keeps the old wording — an older build does keep working, so installing one is the fix. A usage-based key gets "The license key for HyperFormula expired on <date>." and nothing about versions, because there is no version to fall back to and pointing at one sends the reader to downgrade for no reason. Classic 25-character keys are release-only, so their message is untouched.

Two paired specs pin it, each failing on the opposite mutation: collapsing both axes onto the maintenance wording, and collapsing both onto the usage wording.

src/license/handsontable-license-key-parser/PROVENANCE.md
sequba: The goal is (b). What options do we have to achieve that?

src/license/capabilities.ts
sequba: a

src/license/capabilities.ts
sequba: a

src/license/handsontable-license-key-parser/PROVENANCE.md
marcin-kordas-hoc: @sequba Options:

1. Verbatim upstream `.js` + `allowJs` + hand-written `.d.ts` — prototyped; builds, and only the `isIsoDate` tests fail.
2. Same, via git subtree.
3. Upstream publishes a public, reader-only npm package.
4. Upstream ships TS or strict-clean JS.

Any of them drops that guard (upstream accepts `usage_until: ["2099-12-31"]`). I'd upstream it first, then do 1.

src/license/capabilities.ts
marcin-kordas-hoc: @sequba Done. license-key's default schema still mints `functions_1..4` for HF — it needs to emit the `fun:*` group tokens instead. Should `spreadsheet` / `import_export` become `feat:*` too?

src/license/capabilities.ts
marcin-kordas-hoc: @sequba Done.

src/license/capabilities.ts
sequba: Discussion will be continued in https://github.com/handsontable/hyperformula/pull/1728

src/license/handsontable-license-key-parser/PROVENANCE.md
sequba: Discussion will be continued in https://github.com/handsontable/hyperformula/pull/1728

PR #1724, opened 2026-08-10
Remove the static function metadata API (HF-349)

sequba:
### Context

[HF-349](https://app.clickup.com/t/9015210959/HF-349) — a blocker for the 3.4.0 release, under [HF-307](https://app.clickup.com/t/9015210959/HF-307) (feature packages and add-ons), decision D2.

`getAvailableFunctions` and `getFunctionDetails` shipped in both a static and an instance form. License-based entitlement splits the two apart: an instance knows its license key, so it can answer for the engine the caller actually holds; a static method has no key to read, so it can only ever answer for the package as a whole. Only the instance answer is the one integrators need — the static one invites a function picker to advertise functions that then fail on evaluation, and nothing errors at integration time.

Both methods are unreleased, which is the only free moment to remove them.

**This PR removes the static variants. The function metadata API is instance-scoped only.**

Code removed as dead after that change:

- `FunctionRegistry.getListableFunctionIds` (static) — its only caller was the static `getAvailableFunctions`
- the `getPlugin` callback threaded through `buildAvailableFunctions` — it existed only to abstract over the static and the instance registry. Both private builders now take the engine's own `FunctionRegistry`, which also removes the possibility of ids and plugins being resolved from two different sources.

**Docs generator.** The built-in functions guide page is generated from this API, so `script/generate-builtin-functions-doc.ts` now builds an engine and reads the instance methods. It builds that engine with the `gpl-v3` license key — the fully-entitled one — so the published reference keeps documenting every built-in instead of narrowing to the tier of whichever key the build environment happens to supply. The ADR calls this out as a real risk introduced by the decision, so the key is named in the script rather than left to the environment, with the reasoning recorded there and in `docs/README.md`.

**Changelog.** The 3.4.0 entry that introduced these methods is amended (`(both static and instance)` → `instance methods`) rather than accompanied by a "Removed" note: nothing was ever released to remove, and announcing the removal of a method 3.4.0 never shipped would only confuse readers. Same edit in `docs/guide/release-notes.md`.

Tests live in the private repo — companion PR: handsontable/hyperformula-tests#29. Both target `release/3.4.0` and must land together.

The two JSDoc nuances the removed static docs carried (a custom plugin registered *over* a built-in id inherits that id's catalogue entry, in both the list and the details) were folded into the instance methods' JSDoc, so the generated API reference does not lose behaviour that is still tested.

### How did you test your changes?

- `npm run test:ci` with the private suite attached: **502/502 suites, 6180 tests passing**
[...]

PR #1731, opened 2026-08-12
HF-307 PR 4/4: describe only the functions the license grants

marcin-kordas-hoc:
Last of four. Stacks on `hf-307-entitlement-gating-pr3` (#1730) — merge that first. Tests: handsontable/hyperformula-tests#33.

### Context

[HF-307](https://app.clickup.com/t/9015210959/HF-307), decision D2.

`getAvailableFunctions()` and `getFunctionDetails()` read straight from the function registry with no license filter, while the interpreter gates the same functions per call. A restricted key therefore **advertised functions that return `#LIC!` when called** — precisely the failure #1724 (HF-349) removed the static variants to prevent:

> the static one invites a function picker to advertise functions that then fail on evaluation, and nothing errors at integration time

That PR removed the static methods on the rationale that *"an instance knows its license key, so it can answer for the engine the caller actually holds"*. The instance methods never learned to read it. This finishes the job.

### The change

Both methods now filter through `licenseListsFunction`, which shares **one** `licenseAllowsFunction` rule with the interpreter rather than spelling the same condition out twice, and canonicalises aliases the same way. Extracting that rule is the point of the PR as much as the filter is: two copies would drift, and the drift is invisible until a customer's function picker offers something that fails.

### The one decision worth reviewing

**The filter reads gate B only — never the key's validity state.**

A missing, invalid or expired key resolves to an *unrestricted* entitlement (the invariant PR 3 documents), so it reaches the filter with `unrestricted` set and keeps the whole catalogue. That falls out of the invariant rather than being a second decision — but it is a deliberate one, and the alternative is defensible, so it is worth an explicit look.

I chose it because narrowing on gate A would return **two** functions (the protected built-ins) to anyone who calls the API before configuring a key — an empty function picker with no clue why, for the exact integrator still wiring things up. A key problem is already reported on the console and by `#LIC!` in cells. The list narrows only for a *valid* key that genuinely excludes a function, which is when the answer is useful.

Pinned by tests in both directions; folding gate A into the filter fails 3 of them.

### Also

- **`types-of-errors.md`** described `#LIC!` as only ever meaning a key problem. It now also means "not in your package".
- **CHANGELOG** — the feature's first user-visible entry. PRs 1–3 were internals by design and deliberately carry none.
- **The guide documents the mechanism, not the package contents.** HF-306 is still in review with six open questions, so publishing the function lists now would put moving targets in the public docs.

- full suite with the private tests attached: **512 suites, 6383 tests passing**, 3 skipped, 0 failures *(was 6323 when this PR was opened; the stack has since been rebased onto a corrected PR 2 and PR 3 gained four test-hardening fixes)*
- `npx tsc --noEmit`: clean. `npm run lint`: **0 errors**
- `npm run docs:generate-function-docs`: succeeds, and the generated reference still documents `BITAND` / `VLOOKUP` / `XIRR` — the generator builds with `gpl-v3`, so the published docs do not narrow to a tier
- **mutation-tested three ways**: removing the list filter fails 4 tests, dropping the custom-function exemption fails 2, folding gate A into the filter fails 3

Not run here: `npm run test:browser` (Karma needs Firefox, unavailable in this environment). No Jest-only matcher forms were used.

### Types of changes
- [ ] Breaking change
- [ ] Bug fix
- [ ] Additional language file

Not marked breaking: the narrowing only happens for a valid restricted key, and no such key can exist for a released version yet.

### Related issues
1. HF-307 — Implement feature packages and add-ons in HF (decision D2)
2. #1724 / HF-349 — removed the static metadata API on the rationale this PR delivers
3. Companion test PR: handsontable/hyperformula-tests#33

### Checklist
- [x] I have reviewed the guidelines about [Contributing to HyperFormula](https://hyperformula.handsontable.com/guide/contributing.html)
- [ ] I have signed the [Contributor License Agreement](https://goo.gl/forms/yuutGuN0RjsikVpM2)
- [ ] My change is compliant with the OpenDocument standard
- [ ] My change is compatible with Microsoft Excel
- [ ] My change is compatible with Google Sheets
- [x] I described my changes in the CHANGELOG.md file
- [x] My changes require a documentation update

The three compatibility boxes are left unticked as not applicable: this change touches no formula semantics.

### Found while working on this, NOT fixed here

`=OFFSET()` with no arguments **throws a `TypeError` out of the parser** instead of returning an error value — `handleOffsetHeuristic` (`src/parser/FormulaParser.ts:764`) reads `args[0].type` with no arity check. Pre-existing, unrelated to licensing, and out of scope; flagging it for a separate issue.

`getRegisteredFunctionNames()` still returns the whole catalogue under a restricted key — measured: 422 names on a `crm`-tier key, including functions that evaluate to `#LIC!`. Two independent reviewers in the 19.08 pass raised it and both times it was refuted **on scope, not on accuracy**: it is a registry/i18n surface ("what is registered, translated"), not the "what can this engine compute" surface this PR narrows, it sits outside this PR's diff, and it has a static counterpart that would have to move with it. Recording it because a function picker built on that method has exactly the problem #1724 and this PR exist to prevent, so somebody should decide deliberately rather than by omission. Not changed here.

## Update 19.08 — spec-to-ship review: 5 confirmed of 11

Five dimensions, every finding adversarially verified by an agent tasked with refuting it. **Two documentation defects fixed in `e9863f27` here; two test-level gaps fixed in the paired PR; one stale-numbers fix above.** Six findings were refuted, including two independent reports that `getFunctionPlugin()`/`getAllFunctionPlugins()` leak the catalogue — real behaviour, but no promise in this PR is broken by it.

**Both docs defects were measured, not read.** The guide promised custom functions are "always available, whatever your key grants" and the JSDoc added here repeated it; the real rule is "not covered by the capability table", so a plugin registered under a built-in id the key excludes is hidden and returns `#LIC!` (verified with an own plugin implementing `BITAND` on a `crm` key). Separately the guide said a bad key means "every function evaluates to `#LIC!`" and "stops formulas from calculating" — but `VERSION()`, `OFFSET()` and all arithmetic keep working, so a user whose key lapsed was told to expect a blank sheet.

**The capability table was pinned by cardinality only** — the size, nesting and registry-completeness checks cannot see a count-preserving move, and only 6 of ~340 entries were named anywhere in the suite. Swapping `INT` with `SUMIF` across the spreadsheet/calculated-fields boundary passed all 193 license tests and all 128 metadata-API tests; a second swap at a different boundary survived too. Each tier's sorted membership is now checked in and compared by name, so any re-derivation is a reviewable diff. This matters most for a table transcribed by hand from a doc.

**The one test named for the protected built-ins asserted evaluation, not listing**, justified by a comment claiming `OFFSET`/`VERSION` are "excluded from the listable ids entirely" — false, `getListableFunctionIds()` returns them, and `getAvailableFunctions()` does contain both. Dropping the `functionIsProtected` short-circuit and folding the two ids into the table removed them from every restricted key's picker while `=OFFSET(...)` still evaluated, with the suite green. Now asserted on the listing path, with a floor case for a Handsontable-only key.

*Checked and clean:* `baseRefName` correct for last-of-four, head SHAs match, all checks completed and passed on both PRs, no unresolved review threads, and no key material or internal URLs in the public diff. The one thread on the tests PR was marked resolved while the text it flagged was still wrong — that text is what finding 2 above corrects.

---

> [!NOTE]
> **Medium Risk**
> Changes public metadata API behavior and the authoritative function-to-package mapping for restricted keys; wrong drift between interpreter and listing would confuse integrators, though invalid keys intentionally keep the full catalogue.
> 
> **Overview**
> **`getAvailableFunctions()` and `getFunctionDetails()` now list only functions the instance’s valid license actually allows**, using the same gate-B rule as the interpreter via shared **`licenseAllowsFunction`** (and **`licenseListsFunction`** for listing). That closes the gap where a restricted key could advertise functions that evaluate to **`#LIC!`**.
> 
> A **missing, invalid, or expired key does not shrink the catalogue**—gate A is handled separately in evaluation (`#LIC!` in cells), so integrators can still build a picker before a key is wired. **`VERSION()` / `OFFSET()`** and plain arithmetic stay outside this filter.
> 
> **Docs and changelog** document feature packages, **`#LIC!`** for “not in your package,” and the custom-function rule: own ids stay available; registering under a built-in id your key excludes is treated like that built-in.
> 
> The **capability membership tables** in `capabilities.ts` are re-derived from the internal packaging doc (tier counts and per-package function lists change accordingly).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cd79b0678c1e6e67ab7e96d351b5664a7b8f38ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/HyperFormula.ts
+++ b/src/HyperFormula.ts
@@ -4638,9 +4638,12 @@ export class HyperFormula implements TypedEmitter {
    * `#LIC!` error. The list therefore answers "what can this engine compute", not "what does this package contain".
    * Two consequences worth knowing:
    * - A missing, invalid or expired license key does **not** shorten the list. Such a key restricts nothing by
-   *   entitlement — it is reported on the console, and every function evaluates to `#LIC!` — so the full catalogue
+   *   entitlement — it is reported on the console, and every function call evaluates to `#LIC!` — so the full catalogue
    *   is still described. Use it to build a function picker before a key is configured.
-   * - Custom (user-registered) functions are never omitted; the license covers built-ins only.
+   * - A custom (user-registered) function is omitted only if it took a built-in id the key excludes. The rule is
+   *   "not covered by the capability table", not "not user-registered", so a plugin registered under an id the
+   *   built-in catalogue already uses is treated as that built-in. Registered under an id of its own, a custom
+   *   function is never omitted. See {@link getFunctionDetails}, which states the same exception.
    *
    * @example
    * ```js
--- a/src/HyperFormula.ts
+++ b/src/HyperFormula.ts
@@ -4638,8 +4638,9 @@ export class HyperFormula implements TypedEmitter {
    * `#LIC!` error. The list therefore answers "what can this engine compute", not "what does this package contain".
    * Two consequences worth knowing:
    * - A missing, invalid or expired license key does **not** shorten the list. Such a key restricts nothing by
-   *   entitlement — it is reported on the console, and every function call evaluates to `#LIC!` — so the full catalogue
-   *   is still described. Use it to build a function picker before a key is configured.
+   *   entitlement — it is reported on the console, and every licence-gated function call evaluates to `#LIC!` — so
+   *   the full catalogue is still described. `VERSION()` and `OFFSET()` are protected built-ins outside the licence
+   *   system, so they keep evaluating. Use it to build a function picker before a key is configured.
    * - A custom (user-registered) function is omitted only if it took a built-in id the key excludes. The rule is
    *   "not covered by the capability table", not "not user-registered", so a plugin registered under an id the
    *   built-in catalogue already uses is treated as that built-in. Registered under an id of its own, a custom
--- a/src/HyperFormula.ts
+++ b/src/HyperFormula.ts
@@ -44,7 +44,7 @@ import {
 import {Evaluator} from './Evaluator'
 import {ExportedChange, Exporter} from './Exporter'
 import {LicenseKeyValidityState} from './helpers/licenseKeyValidator'
-import {allowsFeature} from './license/CapabilityRegistry'
+import {allowsFeature, licenseAllowsFunction} from './license/CapabilityRegistry'
 import {FeatureId} from './license/LicenseEntitlement'
 import {buildTranslationPackage, RawTranslationPackage, TranslationPackage} from './i18n'
 import {FunctionPluginDefinition} from './interpreter'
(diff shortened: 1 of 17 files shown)


qunabu:
Task linked: [HOT-15413 Pick function sets for each HF product package](https://app.clickup.com/t/9015210959/HOT-15413)

src/HyperFormula.ts
```diff
@@ -4598,6 +4634,17 @@ export class HyperFormula implements TypedEmitter {
    * plugin registered without translations for that language. A translation set to an empty string is not a missing
    * entry: it falls back to the canonical id, so the function stays listed under its canonical name.
    *
+   * A function the instance's license key does not include is omitted for the same reason: it would evaluate to a
+   * `#LIC!` error. The list therefore answers "what can this engine compute", not "what does this package contain".
+   * Two consequences worth knowing:
+   * - A missing, invalid or expired license key does **not** shorten the list. Such a key restricts nothing by
+   *   entitlement — it is reported on the console, and every function call evaluates to `#LIC!` — so the full catalogue
```
Tobiadefami: `every function call` is still too broad, and it also makes the preceding “what can this engine compute” contract false for gate A. On this head with `licenseKey: ''`, the list contains `SUM`, `OFFSET`, and `VERSION`; `SUM()` returns `#LIC!`, while `OFFSET(A1,0,1)` returns `2` and `VERSION()` returns the version because the protected built-ins bypass gate A. The guide repeats the absolute claim and then names those exceptions. It would help to describe the filter as gate-B/package alignment and qualify the bad-key wording consistently here and in the guide.

Tobiadefami (commented):
Reviewed at e9863f27 together with the paired tests at 903aabae. The shared gate-B predicate, alias canonicalization, list/details filtering, protected and custom-function behavior, capability-table membership, documentation, changelog, and current automated findings were checked. Six focused paired suites pass (244 tests), and the current engine checks are green. I found one material documentation inconsistency: the bad-key text says every function call returns #LIC! even though protected OFFSET and VERSION remain callable, as noted inline.

src/HyperFormula.ts
marcin-kordas-hoc: Agreed on both halves — the wording promised a gate wider than the code, and it made the gate-A contract false. 6bfe8f8e7 narrowed the claim to what the metadata surface actually does, and 693afdd19 removed the second promise the docs made and the code did not keep.

marcin-kordas-hoc:

PR #1733, opened 2026-08-13
fix: preserve zero results from AVERAGEIF

Tobiadefami:
### Context

  `AVERAGEIF` returned a `#DIV/0!` error when matching values produced a valid average of `0`.

  This happened because the result used a logical OR fallback, which treated `0` as if no average had been
  calculated. This change uses nullish coalescing so that only an absent result produces `#DIV/0!`, while a
  valid zero is returned normally.

  Added a regression test where the matching values are `-1` and `1`, producing an average of `0`.

  The test was verified to fail with `#DIV/0!` before the fix and pass with `0` after the fix.

  I also ran:

  - The focused `AVERAGEIF` test suite: all 14 tests passed.
  - `npm run bundle:cjs`: TypeScript and the CommonJS build completed successfully.
  - Targeted linting: no errors.

  ### Types of changes

  - [ ] Breaking change (a fix or a feature because of which an existing functionality doesn't work as
  expected anymore)

  ### Related issues:

  None.

  ### Checklist:

  - [x] I have reviewed the guidelines about [Contributing to HyperFormula](https://
  hyperformula.handsontable.com/guide/contributing.html) and I confirm that my code follows the code style of
  this project.
  - [x] My change is compliant with the [OpenDocument](https://docs.oasis-open.org/office/OpenDocument/v1.3/
  os/part4-formula/OpenDocument-v1.3-os-part4-formula.html) standard.
  - [x] I described my changes in the [CHANGELOG.md](https://github.com/handsontable/hyperformula/blob/master/
  CHANGELOG.md) file.

---

> [!NOTE]
> **Low Risk**
> Single-operator change in AVERAGEIF result handling with no impact on auth, data, or broader aggregation APIs.
> 
> **Overview**
> **`AVERAGEIF` now returns `0` when the conditional average is legitimately zero**, instead of incorrectly surfacing `#DIV/0!`.
> 
> The bug came from using logical OR (`||`) after `averageValue()`: a computed average of `0` was treated like a missing result. **`ConditionalAggregationPlugin`** switches that fallback to nullish coalescing (`??`), so only `undefined` (no matching numeric cells / zero count) still maps to `#DIV/0!`. **CHANGELOG** documents the fix under Unreleased.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2eb85e9775e3ea2970ba6fbdbd31c20d92ab3ef4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed the `AVERAGEIF` function returning a division-by-zero error when the calculated average was `0`.
+
 ## [3.4.0] - 2026-08-10
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Fixed
 
-- Fixed the `AVERAGEIF` function returning a division-by-zero error when the calculated average was `0`.
+- Fixed the `AVERAGEIF` function returning a division-by-zero error when the calculated average was `0`. [#1733](https://github.com/handsontable/hyperformula/pull/1733)
 
 ## [3.4.0] - 2026-08-10
 
--- a/src/interpreter/plugin/ConditionalAggregationPlugin.ts
+++ b/src/interpreter/plugin/ConditionalAggregationPlugin.ts
@@ -200,7 +200,7 @@ export class ConditionalAggregationPlugin extends FunctionPlugin implements Func
       if (averageResult instanceof CellError) {
         return averageResult
       } else {
-        return averageResult.averageValue() || new CellError(ErrorType.DIV_BY_ZERO)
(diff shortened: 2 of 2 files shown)


marcin-kordas-hoc (approved):
Approving. Verified rather than eyeballed:

- Differential check of the fix: the same formulas that return `0` on this head return `#DIV/0!` on clean `develop`, so `||` → `??` does exactly what the description claims and nothing more. A genuine no-match case (`=AVERAGEIF(A1:A2,">100")`) still correctly returns `#DIV/0!`, so the fallback is not weakened.
- Paired tests (`hyperformula-tests` `fix/averageif-zero-average`, b163c40) run green against this head: 14/14 in `function-averageif.spec.ts`. The regression test is non-vacuous — it fails on `develop` without this fix.
- CHANGELOG entry present and linked, scope limited to the one expression.

Known and deliberately out of scope: `AVERAGEIF`'s mapFunction still drops `CellError` values from matching rows (unlike `AVERAGEIFS`, which propagates them). That is the separate item we settled in the thread on #1732 — not a blocker here.

Note that merging still waits for the freeze to lift.

sequba (approved):

PR #1736, opened 2026-08-18
HF-307: warn once when a typed key enters its notice window (5/8)

marcin-kordas-hoc:
Consumes the license key's `notice` field: a VALID typed key whose `usage_until` lies within `notice` days of the current UTC instant prints a single console warning naming the expiry date (with a `(UTC)` marker). Stacks on #1731; **rebased onto its current head on 19.08** (the base moved during PR3/PR4's review passes, which had left this PR conflicting).

## Why now

Per Kuba's D5-A (ClickUp, 12.08): hard blocking at/after expiry **stays** in 3.5.0 and the full rev 5 §4.1 window model is a follow-up — but trials made the notice warning concrete for this release: a trial is technically just a key with `grace=0` and `notice>0` whose warnings must surface in the console (packages meeting 12.08), and the trial mechanism lands in August.

## What changed

- `licenseResolution.ts`: `expiryWithinNoticeWindow()` — usage-axis only (rev 5: notice/grace have no effect on the `release_until` axis), window ends exactly where the soft-stop phase would begin, blind to `graceDays` by design.
- `licenseKeyValidator.ts`: `notifyLicenseKeyNotice()` with **per-key** warn-once accounting (`_noticedKeys` keyed by the raw key string) — deliberately not the process-lifetime boolean the state messages use: two engines built with two different keys each get their own warning.
- Blocking behaviour at/after expiry is byte-identical to before; the key's silent flags suppress the warning; CHANGELOG entry included.

## Verification

Paired tests: `hyperformula-tests@spike/hf307-notice-window` (authored RED-first, 8 assertions: inside/outside window, release_until never warns, expired still hard-blocks, silent suppresses, per-key warn-once across two engines). Full license suite **212/212** under Jest (12 suites, `unit/license` + `unit/helpers/licenseKeyValidator`), re-measured after the 19.08 rebase onto the current #1731 (`e9863f27`) — the earlier 165/165 predated PR3's and PR4's review fixes, `tsc --noEmit` clean, `eslint --quiet` clean on changed files.

---

> [!NOTE]
> **Medium Risk**
> Touches license validation and console messaging on every engine build, but changes are additive warnings with explicit silencing for transient configs and no change to expiry blocking or entitlements.
> 
> **Overview**
> Adds a **one-time console warning** for valid typed license keys whose **`usage_until`** expiry falls inside the key’s configured **`notice`** window. The message names the **last covered day** (“valid until … (UTC)”), respects the key’s **silent** flags, and **never** runs for **`release_until`** keys or after hard expiry—blocking at/after expiry is unchanged.
> 
> **`licenseResolution`** introduces **`expiryWithinNoticeWindow()`** (usage axis only, window ends at the usage deadline, ignores grace) and only applies **`noticeDays`** for **rev-5** license shapes. **`resolveLicense`** accepts **`notifyConsole`** so transient config resolution can skip console output.
> 
> **`licenseKeyValidator`** adds **`notifyLicenseKeyNotice()`** with **per-key** warn-once tracking (checksum-based identity), separate from the existing once-per-page invalid/missing/expired messages.
> 
> **`Config`** / **`mergeConfig`** pass **`notifyLicenseMessages`** into resolution; **`rebuildWithConfig`** silences notices for the serialization-only config built from the **outgoing** key while the caller may be replacing it.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit db92bcdbea565bc51ca4d778ad052b61378d5bca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/license/licenseResolution.ts
+++ b/src/license/licenseResolution.ts
@@ -408,10 +408,13 @@ function expiryWithinNoticeWindow(terms: LicenseTerms): Date | null {
     return null
   }
 
-  // The first instant no longer on the usage_until day — the same boundary `validityOf` uses
-  // before adding its grace term.
+  // The window ends at the first instant no longer on the usage_until day — the same boundary
+  // `validityOf` uses before adding its grace term — and opens `notice` days before the licensed
+  // day ITSELF, not before that end. Counting back from the end would shorten the window by a day:
+  // the date-semantics fixtures pin 2027-06-13T00:00:00Z for usage_until 2027-08-12 with notice 60,
+  // and a trial whose notice equals its whole term must warn from the day it is issued.
   const usageAxisDeadline = terms.expiryTimestamp + MILLISECONDS_PER_DAY
-  const noticeWindowStart = usageAxisDeadline - (terms.expiry.noticeDays * MILLISECONDS_PER_DAY)
+  const noticeWindowStart = terms.expiryTimestamp - (terms.expiry.noticeDays * MILLISECONDS_PER_DAY)
   const now = Date.now()
 
   return now >= noticeWindowStart && now < usageAxisDeadline ? new Date(terms.expiryTimestamp) : null
--- a/src/license/licenseResolution.ts
+++ b/src/license/licenseResolution.ts
@@ -391,7 +391,7 @@ function validityOf(terms: LicenseTerms): {state: LicenseKeyValidityState, expir
  * Deliberately blind to `graceDays`: notice is about the usage_until axis itself, not about the
  * grace extension past it. Key spec rev 5 §4.1 sequences notice, then a soft-stop window, then the
  * hard-stop this build already enforces; only the hard stop and this notice are built for 3.5.0
- * (Kuba's decision D5-A), so the window checked here ends exactly where the soft-stop phase would
+ * (decision D5-A), so the window checked here ends exactly where the soft-stop phase would
  * begin, rather than reaching into grace and printing a notice for a key already past its expiry.
  *
  * `release_until`-axis keys never reach here with a non-`null` result — `kind` is `'release'` for
--- a/src/license/licenseResolution.ts
+++ b/src/license/licenseResolution.ts
@@ -6,6 +6,7 @@
 import {
   checkLicenseKeyValidity,
   LicenseKeyValidityState,
+  notifyLicenseKeyNotice,
   notifyLicenseKeyState,
 } from '../helpers/licenseKeyValidator'
 import {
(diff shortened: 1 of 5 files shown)


qunabu:

marcin-kordas-hoc:
Paired tests PR: handsontable/hyperformula-tests#37 — merge it BEFORE this one (fetch-tests pairs by branch name).

src/license/licenseResolution.ts
```diff
@@ -377,6 +382,41 @@ function validityOf(terms: LicenseTerms): {state: LicenseKeyValidityState, expir
...
+ */
+function expiryWithinNoticeWindow(terms: LicenseTerms): Date | null {
+  if (terms.expiry.kind !== 'usage' || terms.expiry.noticeDays <= 0 || terms.expiryTimestamp === null) {
+    return null
+  }
+
+  // The first instant no longer on the usage_until day — the same boundary `validityOf` uses
+  // before adding its grace term.
+  const usageAxisDeadline = terms.expiryTimestamp + MILLISECONDS_PER_DAY
+  const noticeWindowStart = usageAxisDeadline - (terms.expiry.noticeDays * MILLISECONDS_PER_DAY)
```
Tobiadefami: Using `usageAxisDeadline` to derive the start shifts the window one day late. For `usage_until: 2027-08-12` and `notice: 60`, 60 days before the governing date is `2027-06-13T00:00:00Z`, but this computes June 14; I reproduced zero warnings at the former instant. The same off-by-one means a 45-day trial with `notice: 45` misses its intended first warning day. It would help to subtract `noticeDays` from `terms.expiryTimestamp` while keeping `usageAxisDeadline` as the exclusive end, and add an exact-boundary regression.

Tobiadefami (commented):
Reviewed at aa835bf1 together with the paired tests at c832fff7. I checked the notice-window boundaries, usage/release axes, silence and warn-once behavior, config rebuild path, current automated findings, and later stacked branches. The 12 focused license suites pass (215 tests), and the current engine checks are green. I found one material boundary issue: the notice period opens one day late, as noted inline.

src/license/licenseResolution.ts
marcin-kordas-hoc: Correct, and the cause is the one you named: the start was derived from the usage-axis deadline rather than the governing expiry. 78849f1a7 computes it as `terms.expiryTimestamp - noticeDays * MILLISECONDS_PER_DAY`, so `usage_until: 2027-08-12` with `notice: 60` opens the window on 2027-06-13, not a day later.

marcin-kordas-hoc:

Issue #1641, opened 2026-03-20
[Bug]: ADDRESS returns "!R2C3" when sheetName is empty (should return "R2C3")


marcin-kordas-hoc:
### Description

`ADDRESS(2,3,1,FALSE(),)` returns `!R2C3` instead of `R2C3`.

When the `sheetName` argument is an empty string (either from an empty cell reference or trailing comma), HyperFormula still prepends the `!` separator. Excel ignores empty sheetName and returns the address without prefix.

### Root cause

In `src/interpreter/plugin/AddressPlugin.ts:70`, the condition:

```ts
if (sheetName !== undefined && sheetName !== null)
```

does not exclude empty strings. Fix: add `&& sheetName !== ''`.

### Reproduction steps

```typescript
const hf = HyperFormula.buildFromArray([
  ['=ADDRESS(2,3,1,FALSE(),)'], // empty sheetName — should be "R2C3", bug: "!R2C3"
], { licenseKey: 'gpl-v3' })

hf.getCellValue({ sheet: 0, row: 0, col: 0 })
// Actual:   "!R2C3"
// Expected: "R2C3" (Excel behavior, Microsoft 365 desktop)
```

### Excel reference

`=ADDRESS(2,3,1,FALSE(),)` → `R2C3` (Microsoft 365 desktop)

### HyperFormula version

3.2.0

### Your environment

Node.js

PR #1739, opened 2026-08-18
HF-160: ADDRESS drops the sheet separator when sheetName is empty

marcin-kordas-hoc:
Fixes HF-160 (GH #1641): `=ADDRESS(2,3,1,FALSE(),)` returned `!R2C3` — the sheet separator was emitted for an empty-string sheet name. Excel returns `R2C3`.

## What changed

One condition in `AddressPlugin`: an empty string no longer produces a sheet prefix, for both A1 and R1C1 styles. Non-empty sheet names keep producing `Sheet1!R2C3` exactly as before.

## Verification

Paired tests: `hyperformula-tests@spike/hf160-address-sheetname` (37 assertions: trailing-comma / explicit `""` / omitted forms in both styles, non-empty names pinned unchanged). Full `unit/interpreter` 390 suites / 3531 tests green, `tsc --noEmit` clean, `eslint --quiet` clean.

---

> [!NOTE]
> **Low Risk**
> Behavior change is scoped behind the new `emptyAsAbsent` flag; only `ADDRESS` adopts it, with limited impact on the shared argument-coercion path.
> 
> **Overview**
> **`ADDRESS`** now matches Excel for the optional sheet name: omitted or trailing-comma empty slots no longer add the `!` prefix (e.g. `R2C3` instead of `!R2C3`), while an explicit empty value (`""` or an empty cell) still keeps `!`.
> 
> This is enabled by a new **`emptyAsAbsent`** parameter flag on custom/built-in functions. In **`FunctionPlugin`**, syntactically empty argument slots can be passed through as **`undefined`** instead of type zero-values (`""`, `0`, `FALSE`), only when **`emptyAsDefault`** is not in play. **`ADDRESS`**’s fifth argument uses **`emptyAsAbsent: true`**.
> 
> Docs and changelog describe the new option and the **`ADDRESS`** fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dbab806399cf4dff4641d4b778cd2710d594e218. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [HF-160 Fix ADDRESS returning '!R2C3' when sheetName is empty](https://app.clickup.com/t/9015210959/HF-160)

marcin-kordas-hoc:
Paired tests PR: handsontable/hyperformula-tests#40 — merge it BEFORE this one.

marcin-kordas-hoc:
Excel-measured finding on this PR's scope (28.08, MS Graph / Excel Online, evidence workbook + per-case JSONs archived): the core trailing-comma case **matches** — `=ADDRESS(2,3,1,FALSE(),)` → `R2C3` — but an explicit `""` and a sheetName taken from an empty cell reference both **KEEP the bare separator** in Excel (`!R2C3` / `!$C$2`; `LEN`=5 confirmed twice in fresh sessions). Measured rule: only *syntactic omission* drops the `!`; an empty-string *value* keeps it.

That contradicts three of the paired tests (`function-address.spec.ts` l.352 / l.364 / l.448) and the task AC's premise ("consistent with Excel") for those two of its three cases.

Caveat: measured on Excel Online. @Tobiadefami — could you spot-check both formulas in desktop M365 (two minutes: `=ADDRESS(2,3,1,FALSE(),"")` and `=ADDRESS(2,3,1,FALSE(),Z99)` with Z99 empty)? Then the call is: narrow the fix to the omission case (strict Excel), or keep all three per the AC as a documented deviation and amend the AC wording.

marcin-kordas-hoc:
@Tobiadefami — requesting your review. Write-up with the full context (requirement, decision, technical pass, verification) is here: https://claude.ai/code/artifact/23606cf3-6e22-40e8-ab00-997780a674b9 (item 1). CI 34/34 green, 0 open threads.

Tobiadefami (approved):
The implementation looks good, and the absent-versus-empty behavior matches the cases verified in Excel Online.

Could we update “What changed”? Explicit `""` still retains the sheet prefix; omitted arguments and syntactically empty slots omit it.

PR #1737, opened 2026-08-18
HF-307: wire the two commercial add-on tokens, strictly additively (6/8)

marcin-kordas-hoc:
Wires the two commercial add-on tokens decided in the 12.08 packages meeting, **strictly additively**. Stacks on #1731; **rebased onto its current head on 19.08** (the base moved during PR3/PR4's review passes, which had left this PR conflicting).

## What changed

- `spreadsheet` (Spreadsheet Bundle add-on) now grants `FeatureId.Crud`, `UndoRedo`, `Clipboard` and `Batching` ("chyba też" batching — Kuba, 12.08).
- `import_export` grants `FeatureId.ImportExport` — a **reserved** grant: no public method is gated on it until HF-107 ships the feature it would gate.
- Additive only: a key naming neither add-on keeps every feature area it has today (the opt-in rule for keys carrying no `feat:*` tokens is untouched), so no real key can start throwing as a side effect. The open product question — does a package key without the bundle keep CRUD once packages go live — stays open and is documented as such in `docs/guide/license-key.md`.

## Verification

Paired tests: `hyperformula-tests@spike/hf307-addon-grants` (6 assertions incl. the pin that a key without either add-on keeps all five pre-existing feature areas). Full license suite **212/212** under Jest (12 suites, `unit/license` + `unit/helpers/licenseKeyValidator`), re-measured after the 19.08 rebase onto the current #1731 (`e9863f27`) — the earlier 165/165 predated PR3's and PR4's review fixes, `tsc --noEmit` clean, `eslint --quiet` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> License capability metadata and documentation only; runtime gating for current keys is unchanged because grants are additive and import/export has no gates yet.
> 
> **Overview**
> Wires the **`spreadsheet`** and **`import_export`** commercial add-on tokens into `CAPABILITY_TABLE`, matching the 2026-08-12 packages meeting.
> 
> The **`spreadsheet`** (Spreadsheet Bundle) token now grants **CRUD**, **undo/redo**, **clipboard**, and **batching** — **named expressions** stay outside the bundle by design. **`import_export`** grants **`FeatureId.ImportExport`** as a reserved entitlement; nothing in the public API gates on it until HF-107 ships import/export.
> 
[...]

PR #1740, opened 2026-08-20
HF-329: re-port the license key reader to the entitlement key format (7/8)

marcin-kordas-hoc:
7/8 of the HF-307 stack (#1728 → #1729 → #1730 → #1731 → #1736 → #1737 → this). Implements **HF-329** ("Import license key validator shared code to HF"). Pairs with `hyperformula-tests#41` — **merge the tests PR first** (CI clones it by branch name).

## Why now

Upstream `handsontable/license-key` **4.0.0** (18.08, DEV-2512) **deleted `src/typed-key/`** — the exact directory our vendored reader was ported from — and replaced the tagged key format with the **entitlement key format**:

```
<prose>\n\n[<base64url-payload><sha512-checksum>]
```

my.handsontable.com already pins 4.0.0 and mints entitlement keys for trials, so the vendored reader was reading a format that no longer exists anywhere. The byte-level rules are specified in the key spec's "Technical implementation" addendum (T1–T14).

## What changed

**Only the envelope.** The entitlement layer — token resolution, the capability table, gate A/B, `ensureCapability` — is untouched by design.

- Re-vendored from `src/entitlement-key/` at tag 4.0.0: `detectFormat.ts` and `extractKeyData.ts` are new ports; `sha512.ts`/`utils.ts` are **byte-identical upstream** (same hashes) and carry over. `PROVENANCE.md` records the new tag, commit, and sha256 table.
- The reader is **schema-free** by upstream design (strict on shape, lenient on vocabulary — T7), so `defaultSchema.ts` is no longer vendored; the only name the engine reads is its own `hyperformula` product entry.
- `resolveLicense` routes on the vendored `detectLicenseKeyFormat` (normative T12 order: literals → bracketed block → legacy shape). Literal keys and legacy 25-character keys reach `checkLicenseKeyValidity` **completely unchanged**.
- **The invariant stands**: only a VALID entitlement key may restrict the entitlement; MISSING/INVALID/EXPIRED resolve to `unrestrictedEntitlement()` (mutation-tested in the suite).

## Decisions to confirm in review

1. **The tagged format is removed outright, not kept alongside.** It was never issued to anyone (its 3.5.0 carrier is unreleased, and my.handsontable.com's own PR says "the typed keys never reached production"), upstream deleted it, and PROVENANCE mirrors upstream. With it goes `TIER_TO_CAPABILITY_TOKEN` — the entitlement format has no `tier` field.
2. **Deliberate behaviour deltas**, each spec-mandated and pinned in the paired tests:
   - a **prose edit keeps a key VALID** (T8, stated by the addendum as a reversal of the typed-key behaviour);
   - a key granting **another product only never expires for HyperFormula** (each entry carries its own terms; the licensed-product rule is gone);
   - **both-dates, numeric-date, and null-entry payloads are INVALID** (T7 strict shape; the old adapter tolerated each).
3. Dates are carried as the payload's own `YYYY-MM-DD` strings, never routed through `Date` formatting (fixture J11).

## Testing

Fixtures re-minted with upstream's **real generator** at 4.0.0 and round-tripped through upstream's extractor; the addendum's **T14 conformance vectors pinned 1:1**, baselined on its T13 worked example verbatim. Full private suite: **516 suites / 6430 passed, 3 pre-existing skips**; `tsc --noEmit` and ESLint clean.

## Found while working on this, NOT fixed here

`CHANGELOG.md`'s `[Unreleased]` has no `Removed` entry for the static function metadata API removed by HF-349 (#1724) — a breaking change currently shipping into 3.5.0 unannounced. Raised separately.

---

> [!NOTE]
> **Medium Risk**
> Rewrites proprietary license parsing and routing, which is business-critical, but legacy/gpl paths and the “bad key → unrestricted entitlement” invariant are preserved; risk is mainly wrong verdicts or grants for new entitlement keys and hostile payloads.
> 
> **Overview**
> **Replaces the internal tagged (“typed”) proprietary key reader with the upstream entitlement key format** — human-readable text plus a trailing `[base64url-payload + sha512]` block — vendored from `handsontable/license-key` **4.0.0**. The old tagged format and `defaultSchema.ts` / tier-to-package mapping are removed; classic 25-character keys and `gpl-v3` still route through the unchanged legacy validator.
> 
> `resolveLicense` now uses `detectLicenseKeyFormat` (literals → bracketed block → legacy shape) and `extractEntitlementKeyData`. Terms come only from the payload’s **`hyperformula` product entry** (per-entry dates, notice, grace, flags); keys for other products or with no HyperFormula entry are valid but grant nothing for this library.
> 
> **Behavior fixes on the entitlement path:** capability tokens are appended one-by-one so huge payloads cannot crash construction via `push(...array)`; the “grant all five `feat:*` areas” fallback runs only when the key names a **recognized** `feat:*` token, so unknown `feat:*` tokens stay inert per D3. Console notice identity uses the checksum plus closing bracket (129 trailing chars).
> 
> Docs and **CHANGELOG** describe both key formats, validation axes (`usage_until` vs `release_until`), and the valid-key expiry notice.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a425d37be6de91f684d3ba3dd5d3e6c9a8fabbd0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Spec-to-ship review (2026-08-20): 3 findings fixed here

A six-dimension adversarial review ran over this PR, #1741 and #1743 (spec conformance, decision
conformance, port fidelity, hostile payloads, dialect table, API surface). What it found against
*this* commit, all fixed above and each one verified by running it:

1. **A hostile key could crash engine construction.** `capabilityTokens.push(...grant.capabilities)`
   spread an attacker-sized array into a call: measured, a checksum-valid key carrying 125 000
   tokens threw `RangeError: Maximum call stack size exceeded` out of `buildFromArray` instead of
   resolving to a verdict — the one thing this module's own doc comment promises never happens. Now
   appended element by element; 200 000 tokens (a 3 MB key) resolves in 141 ms. Pinned by a test.
2. **The T14 payload-tamper vector was vacuous.** `rejects one character changed in the payload
   segment [catches: the checksum doing nothing]` passed with the checksum comparison deleted
   outright, because the flipped character corrupted the payload's `products` key and the T7 shape
   check rejected it first. Mutation-verified in both directions: with the comparison removed only
   the *checksum-segment* vector failed. Replaced with a well-formed payload (the T13 record with
   `notice: 61`) carrying T13's original checksum, plus a positive control asserting the same
   payload with its own checksum is accepted — so the rejection can only come from the checksum.
3. **Two inaccuracies in what this PR wrote.** `PROVENANCE.md`'s divergence #3 claimed the typed
   entry's "fields carry their verified types", but the date field is verified only by matching
   `String(value)` against `YYYY-MM-DD`, so an array holding one date string passes and the declared
   `string` is wider than the value (faithful to upstream — the manifest was the thing overstating).
   And `isNonNegativeInteger` had drifted to the global `isFinite` where upstream uses
   `Number.isFinite`: behaviour-identical behind the `typeof` guard, but an undocumented edit inside
   a hash-pinned mirror. Both corrected.

Also corrected in the guide: the classic-key example showed 20 characters next to the sentence
"25 characters", and the entitlement-key paragraph said the text "may be re-wrapped or re-formatted
… without affecting the key" — true of the prose, false of the block, which has to arrive character
for character.


qunabu:
Task linked: [HF-329 Import license key validator shared code to HF](https://app.clickup.com/t/9015210959/HF-329)

Tobiadefami (commented):
Reviewed at 523316e5 together with the paired tests at 81f180ce. I checked format routing, checksum-before-parse ordering, the strict-shape/lenient-vocabulary contract, per-product expiry, the invalid-key invariant, hostile-payload handling, provenance, docs and changelog, current automated findings, and later stacked branches. The 15 focused license suites pass (256 tests), and all current engine checks are green. I found one material shape-validation issue: a single-element array is accepted where an entitlement date must be a YYYY-MM-DD string, as noted inline.

src/license/vendor/extractKeyData.ts
```diff
@@ -4,212 +4,275 @@
...
+ * @param {unknown} value - the value to check
  */
-function resolveLicensedProduct(payload: TypedKeyPayload): LicensedProduct | null {
-  const {products} = payload
+function isIsoDate(value: unknown): boolean {
+  try {
+    // The cast mirrors upstream, where the call is untyped: `parseIsoDate` stringifies its
+    // argument before matching it against the `YYYY-MM-DD` shape, so a non-string value is
+    // rejected by the shape check rather than by a type guard here.
+    parseIsoDate(value as string, 'license')
```
Tobiadefami: `value as string` is only a compile-time assertion, while `parseIsoDate` stringifies at runtime. A checksum-valid payload with `usage_until: ["2099-12-31"]` therefore passes extraction unchanged and `resolveLicense` reports it `VALID`; I reproduced both on this head. That contradicts T7/J6's strict `YYYY-MM-DD` string shape and makes the declared `usage_until?: string` type untrue, so malformed input can activate a restricted entitlement instead of taking the invalid-key path. It would help to require `typeof value === 'string'` before parsing, fix the mirror upstream first, and pin the array case.

marcin-kordas-hoc: Yes — `as string` is erased at runtime and `parseIsoDate` would have stringified an array into a date-shaped string. a425d37be adds a real `typeof value === 'string'` check before the parse, so a checksum-valid payload with `usage_until: ["2099-12-31"]` is rejected as malformed instead of silently becoming a date.

marcin-kordas-hoc:

PR #1741, opened 2026-08-20
HF-306/HF-307: resolve both capability-token dialects from one group registry (8/8)

marcin-kordas-hoc:
8/8 of the HF-307 stack, stacked on #1740. Pairs with `hyperformula-tests#42` — **merge the tests PR first**.

## Why

Two capability-token vocabularies exist for HyperFormula and the business choice between them is still open (D8, asked on HF-307 on 13.08, unanswered):

- the **key spec / generator vocabulary** — `functions_1..4` package slices plus the `spreadsheet`/`import_export` add-ons. This is what upstream `license-key@4.0.0`'s own schema mints today;
- the **packaging doc vocabulary** (§6, 12.08, linked from HF-307 by its author as "the list of capability tokens") — `fun:all`, 21 group tokens `fun:<family>.<A|B|C>`, and a single-function token `fun:<CANONICAL_NAME>` for every catalog entry.

Rather than bet on one, the engine now resolves **both** (owner's decision, 20.08). This is spec-clean: T7 defines an unrecognized token as "a grant this version does not implement", so implementing a superset breaks nothing — and whichever dialect keys end up minted in, they work without another engine release.

## Design: one source of truth, two dialects reading from it

- The **21 groups are transcribed 1:1 from the doc's §6** into `FUNCTION_GROUPS` (members verbatim, operators and protected built-ins included, so a drift check can compare against the doc's published counts).
- The `functions_1..3` package slices are now **DERIVED** from the groups as the doc's own cumulative unions (`.A` / `.A+.B` / `.A+.B+.C`), minus the operators (granted by `core`) and the protected built-ins. The derived memberships are pinned **by name** by the pre-existing `capability-table.spec.ts` lists, which is the consistency proof: derivation reproduces the previously authored lists exactly.
- `fun:all` grants byte-for-byte what `functions_4` grants; the excel-simulator tail stays an authored list (the doc does not itemize it, and deriving it from the registry would gate user-registered functions — D1).
- `fun:info.A` / `fun:lookup.A` / `fun:OFFSET` / `fun:VERSION` resolve to **empty grants**: their members are the protected built-ins, which are always available and must never become table-covered. Recognized bookkeeping, exactly as the doc frames them ("technical limitation").
- **Token matching is case-insensitive** (the doc states it outright for `fun:*`; the other tokens tolerate it for free — nothing collides under lowercasing). Normalization happens at lookup; entitlements keep the key's own spellings as diagnostics.
- Alias-**named** tokens stay unrecognized: tokens reference canonical names, and an alias travels with its canonical function because the gates canonicalize before consulting the table (`fun:STDEV.S` accepts `=STDEV(...)`).
- Table entry order is load-bearing for `capabilityOf`'s first-wins reverse index: the package slices stay ahead of the new tokens, so its answers are unchanged.

## ⚠️ A pinned behaviour is deliberately inverted

The 18.08 test "a key in the published `fun:*` vocabulary grants zero functions, silently" pinned a measured outcome *pending D9*. This PR resolves it the way the pin's own comment anticipated ("translate the vocabulary, warn, or reject — it has to change this test"): the vocabulary is now implemented. The inversion is by the owner's 20.08 decision, not a regression, and the replacement tests say so in place.

## Risks, stated

- The group registry is one more artifact to keep in sync with the packaging doc — mitigated by the drift check (per-group counts from §6) and the by-name membership pins.
- This does **not** resolve the commercial 3-vs-4 package split (upstream's default schema sells Essential/Pro/Complete; the packaging doc describes four packages). Packages are generator-side data — the engine only ever sees tokens — so that's a myHOT/business call, not an engine one.
- If the business ultimately picks a single dialect, the other one becomes dead code — cheap to keep (it's derived data, one source of truth) or to remove.

## Testing

`unit/license` + `unit/helpers/licenseKeyValidator`: **16 suites / 266 tests** green; dialect-equivalence pinned per package (`functions_N` ≡ `fun:*` union, `fun:all` ≡ `functions_1..4`); `tsc --noEmit` and ESLint clean.

---

> [!NOTE]
> **High Risk**
> Changes commercial license entitlement resolution and feature gating predicates; incorrect grants or fallback logic would block or over-grant API areas and functions for paying customers.
> 
> **Overview**
> This PR unifies **two license capability vocabularies** so keys minted with either `functions_1`–`functions_4` (and add-ons) or the packaging doc’s `fun:all`, `fun:<family>.<A|B|C>`, and per-function `fun:<NAME>` tokens grant the **same function sets**. The packaging §6 groups live in **`FUNCTION_GROUPS`**; `functions_1`–`functions_3` memberships are **derived** from those cumulative group unions instead of hand-maintained lists.
> 
> **Lookup behavior** now uses **`normalizeCapabilityToken`** (trim + lowercase) everywhere the table is consulted, including **`CapabilityRegistry.resolve`**, which also drops **`queue.shift()`** for a read cursor to avoid **O(n²)** expansion on huge token lists.
> 
> **License resolution** fixes two edge cases: an unrecognized **`feat:*`** token no longer suppresses the “grant all five feature areas” fallback (only **recognized** feature tokens opt into explicit feature gating), and padded tokens like `' feat:crud'` vs `'feat:crud '` no longer disagree between the feature-prefix rule and table lookup.
> 
> `fun:all` matches **`functions_4`**; protected built-ins (`OFFSET`, `VERSION`) and their group tokens stay **recognized but grant nothing**; table entry order preserves prior **`capabilityOf`** answers for package slices.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6b23a4038a42575ce15d96e0825a676345d6ba05. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Spec-to-ship review (2026-08-20): 3 findings fixed here

1. **An unrecognized `feat:*` token revoked the whole gated API instead of being inert.** The
   feature opt-in trigger was a purely syntactic prefix test, so a token this version does not
   implement suppressed the "this key does not talk about features" fallback and left the key with
   **zero** of the five areas. Measured:

   | key's capabilities | feature areas granted |
   |---|---|
   | `functions_1` | 5 / 5 |
   | `functions_1`, `feat:crud` | 1 / 5 — correct |
   | `functions_1`, `feat:teleport` (unknown) | **0 / 5** |
   | `functions_1`, `feat:cruds` (typo at issuing time) | **0 / 5** |

   That inverts the additive-safety rule for the two cases that matter — an older build meeting a
   key from a newer generator, and a one-character typo when a key is minted — and it contradicts
   D3, which makes an unrecognized token *silently ignored*. The trigger now requires a feature
   token this version **recognizes**. Four tests pin both directions.

   Scope note for the reviewer: the defect predates this PR (it came in with the opt-in rule in
   PR 3/9) — the adversarial verifier correctly flagged that. It is fixed here because this PR
   rewrites that exact predicate to add case-insensitivity, and because leaving a known revocation
   bug in place while touching the line would be the wrong call.

2. **Group membership was pinned only by COUNT.** `FUNCTION_GROUPS` is the single source of truth
   for both dialects, but only the doc's per-group *sizes* were checked — so moving a function
   between two groups of the same tier letter changed what a `fun:<family>.<letter>` key grants
   while leaving every count, every derived package union and every other assertion identical.
   Demonstrated live: with `EXP` and `MEDIAN` swapped between `fun:math.C` and `fun:stat.C`, the
   whole license suite stayed green at 272/272. The 21 memberships are now checked **by name**
   against §6 of the packaging doc (the expected literal was generated by parsing the doc itself),
   and that check catches the swap.

3. **Token expansion was quadratic, and whitespace split a token's meaning in two.**
   `CapabilityRegistry.resolve` drained its queue with `Array.prototype.shift` — O(n) per call, so
   O(n²) overall: 1.4 s inside the constructor for a key with 100 000 tokens. Now a read cursor.
   Separately, `normalizeCapabilityToken` lowercased but did not trim, and the two rules that read a
   token disagreed for a padded one: `' feat:crud'` failed the prefix test and was granted all five
   areas, `'feat:crud '` passed it and was granted none — neither was "crud". Normalization now
   trims, pinned by a test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Changed `getAvailableFunctions()` and `getFunctionDetails()` to describe only the functions the instance's license key includes, so they no longer advertise a function that would evaluate to a `#LIC!` error. A missing, invalid, or expired key does not shorten the list. [#1731](https://github.com/handsontable/hyperformula/pull/1731)
 - Changed the parser for the new proprietary license keys to the entitlement key format (a human-readable text ending with a machine-readable block in square brackets), following its upstream specification. This replaces the tagged key format, which was never issued to anyone. Classic 25-character license keys and `gpl-v3` are unaffected. [#1740](https://github.com/handsontable/hyperformula/pull/1740)
+- Changed the license capability tokens to be matched case-insensitively, and added support for the packaging group-token vocabulary (`fun:all`, `fun:<family>.<A|B|C>`, and per-function `fun:<FUNCTION_NAME>` tokens) alongside the existing package tokens (`functions_1`–`functions_4` and the add-ons). A key worded in either vocabulary grants the same functions. [#1741](https://github.com/handsontable/hyperformula/pull/1741)
 
 ## [3.4.0] - 2026-08-10
 
--- a/src/license/CapabilityRegistry.ts
+++ b/src/license/CapabilityRegistry.ts
@@ -4,7 +4,7 @@
  */
 
 import {FeatureId, LicenseEntitlement} from './LicenseEntitlement'
-import {CAPABILITY_TABLE, CapabilityGrant} from './capabilities'
+import {CAPABILITY_TABLE, CapabilityGrant, normalizeCapabilityToken} from './capabilities'
 
 /**
  * The capabilities a resolved {@link LicenseEntitlement} grants, ready for gate B (the
@@ -59,12 +59,13 @@ export class CapabilityRegistry {
   /**
    * Expands an entitlement's capability tokens into the concrete functions and features they
    * grant. An `unrestricted` entitlement short-circuits to an unrestricted result without
-   * consulting the table at all. Expansion through `implies` is transitive and cycle-safe (a
-   * visited set guards against a token implying itself, directly or through others); an
-   * unrecognized token is skipped without an error.
+   * consulting the table at all. Tokens are matched case-insensitively (the table is keyed by
+   * the normalized spelling — see {@link normalizeCapabilityToken}). Expansion through `implies`
+   * is transitive and cycle-safe (a visited set guards against a token implying itself, directly
+   * or through others); an unrecognized token is skipped without an error.
    *
    * @param {LicenseEntitlement} entitlement - the entitlement to resolve, e.g. one built by
-   * hand in a test or produced by PR 3's license-key payload adapter
+   * hand in a test or produced by the license-key payload adapter
    */
   public resolve(entitlement: LicenseEntitlement): ResolvedCapabilities {
     if (entitlement.unrestricted) {
@@ -74,10 +75,17 @@ export class CapabilityRegistry {
     const functions = new Set<string>()
     const features = new Set<FeatureId>()
     const visited = new Set<string>()
+    // Walked with a read cursor rather than `queue.shift()`: `shift` is O(n) in most engines, which
+    // made expansion O(n^2) in the number of tokens a key carries - measured at 1.4 s inside the
+    // constructor for a key with 100 000 tokens, which the format's missing size limit allows.
+    // `implies` still appends, so the queue has to stay a growable array.
     const queue = [...entitlement.capabilities]
+    let cursor = 0
 
-    while (queue.length > 0) {
-      const token = queue.shift() as string
+    while (cursor < queue.length) {
+      const token = normalizeCapabilityToken(queue[cursor])
+
+      cursor += 1
       if (visited.has(token)) {
         continue
       }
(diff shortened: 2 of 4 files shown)


qunabu:
Task linked: [HOT-15413 Pick function sets for each HF product package](https://app.clickup.com/t/9015210959/HOT-15413)

Tobiadefami (commented):
Reviewed at 4c8258e8 together with the paired tests at e2a35206. I checked the 21 group memberships against the captured packaging source, package and single-function equivalence across both token dialects, case and whitespace normalization, unknown feature-token inertness, protected built-ins, reverse-index ordering, hostile-token-list performance, documentation, changelog, and current automated findings. The 16 focused license suites pass (278 tests), TypeScript passes, all current engine checks are green, and I found no material issue in this PR.

marcin-kordas-hoc:

PR #1743, opened 2026-08-20
HF-307: align getRegisteredFunctionNames with the license gate, deprecate its static form

marcin-kordas-hoc:

## Update 2026-09-18 — rebased onto the collapsed engine/tests PRs

`hyperformula#1730` and `hyperformula-tests#32` were folded from five PRs each into one, so this
PR's base moved to `#1730`. Nothing in this PR's own diff changed; the "9/9" numbering and the
`#1741` base reference below are stale — the stack is now `#1728 → #1729 → #1730 → this PR`.

---

9/9 of the HF-307 stack, stacked on #1741. Pairs with `hyperformula-tests#43` — **merge the tests PR first**. Finishes decision **D2** on the one API it missed.

## The problem

`getRegisteredFunctionNames()` returned the whole catalogue under a restricted key (measured: 422 names on a calculated-fields key, including functions that evaluate to `#LIC!`). A function picker built on it has exactly the problem #1724 (HF-349) and #1731's metadata filter exist to prevent. Flagged in #1731's "found, not fixed here" section; this closes it.

## What changed

**Instance method** — now lists exactly what the instance can evaluate, sharing the one `licenseListsFunction` rule with the interpreter and the metadata API so the three surfaces cannot drift:

- reads `getListableFunctionIds()` instead of `getRegisteredFunctionIds()` — the protected built-ins are included uniformly (`OFFSET` was missing before; `getAvailableFunctions` already listed it);
- reads `config.translationPackage` — the instance's own snapshot — instead of a fresh global `getLanguage` lookup, which can report a localized name the instance refuses to evaluate, and which throws once the host unregisters that language code;
- filters by the license, with the invariant intact: a missing, invalid, or expired key does **not** shorten the list.

**Static method — deprecated, not removed.** See below; this is a change from the first version of this PR.

## The static: deprecated rather than removed (changed after review)

The first version of this PR deleted `HyperFormula.getRegisteredFunctionNames(code)`, citing HF-349 as precedent. That was wrong, and the review caught it:

```
$ git show 3.4.0:src/HyperFormula.ts | grep -c "public static getRegisteredFunctionNames"   # 1
```

The method **is in the released 3.4.0 tag**, and HF-349's own commit message says its removal was free precisely because *"Both methods are unreleased, which is the only free moment to remove them."* So the precedent does not extend here: deleting it would be a breaking change in a minor release, against the Semantic Versioning this project states it follows, and DEV_DOCS's Definition of Done would require a migration-guide section that `docs/guide/` has no 3.x home for.

So it is now `@deprecated` with the wording this repo already uses for that situation (`arraySizeMethod` / `arrayFunction` in 3.1.0: "deprecated and will be removed in one of the next major releases"), plus a `Deprecated` changelog entry. **Overturnable in one comment** if you would rather take the break now — the direction is Kuba's D2 either way; only the timing changed.

The deprecation notice is explicit that the two are not interchangeable: the static translates into any registered language without an engine, so migrating means building one (`HyperFormula.buildEmpty({ language: 'plPL' }).getRegisteredFunctionNames()`).

## D2's precondition, discharged

Kuba's D2 answer made the removal conditional: *"perhaps we can remove the static methods, I'll check which methods does formula-builder use"*. That check is done — searched the org, and **both consumers call the instance form, not the static one**:

| Consumer | Call site | Form |
|---|---|---|
| `formula-builder` | `packages/core/src/engine/functionCatalog.ts:81` — `this.engine?.getRegisteredFunctionNames?.()`, in try/catch, declared optional in `engine/types.ts:66` | instance |
| `aurasheet` | `src/core/FormulaEngine.ts:97` — `this.hf.getRegisteredFunctionNames()` | instance |

Both use it to build a **function picker** (`functionCatalog.ts`; `FormulaAutocompletePlugin.ts`) — the surface #1724's rationale was about — so the license filter added here improves both rather than disturbing them. Under `gpl-v3` or any unrestricted key their lists are unchanged.

## Spec-to-ship review (2026-08-20): also fixed here

- **The translation-snapshot change was unpinned.** Mutation-verified: reverting it to the global lookup left 309 tests green, even though the commit message and the new JSDoc both name it. Now pinned by a test that unregisters the language after the engine is built — the scenario that distinguishes the two implementations, since `getLanguage` throws for an unregistered code while the snapshot keeps working.
- **The licence guide listed only two of the three narrowing methods.** It now names this one too, in both places.
- **The new JSDoc over-claimed parity** with `getAvailableFunctions`: for a function whose translation is the empty string this method returns `''` while that one falls back to the canonical id. The claim is now scoped to the ids and the licence rule, with the naming difference stated.
- Reverting the static removal also removed the docs-build change it required, so `docs/.vuepress/config.js` is untouched by this PR again (it no longer builds one engine per documentation page).

## Testing

New suite pins the alignment in both directions (agrees with `getAvailableFunctions` name for name; narrows on a restricted key; never narrows on a bad key; aliases gate with their canonical; answers from the instance's own snapshot). Full private suite: **517 suites / 6462 passing, 3 pre-existing skips**; `tsc --noEmit` and ESLint clean.

## Note for review

`functions-metadata.spec.ts` now skips listed ids with no plugin: `OFFSET` is listed (it is callable) but parse-time resolved, so it legitimately has no registry metadata.

## The codecov/project dip, traced

`codecov/project` was red at −0.02% while `codecov/patch` reported 100% of the diff hit. Rather than
write that off as a threshold artifact, I measured coverage on this commit and on its base and
diffed the per-file numbers:

| file | base (8/9) | this PR, before the fix |
|---|---|---|
| `src/HyperFormula.ts` | 674 / 675 | 675 / 676 |
| `src/interpreter/FunctionRegistry.ts` | 130 / 130 | **129 / 130** |
| total | 12 933 / 13 268 | 12 933 / 13 269 |

So the covered count did not move and one previously covered line stopped executing — a real
consequence of this change, not a rounding artifact. The line was the **instance**
`FunctionRegistry.prototype.getRegisteredFunctionIds()`, whose only caller in the whole repository
was the method this PR rewrites. Nothing in `src/`, nothing in the private suite, and nothing
outside (the class is not exported from `src/index.ts`) calls it any more.

Removed, since this change is what orphaned it. The static `FunctionRegistry.getRegisteredFunctionIds()`
is untouched and still used — by the deprecated static method above and by three specs.

For the record, the single uncovered line left in `HyperFormula.ts` is pre-existing and not mine:
`removeNamedExpression`'s unreachable `return []`, which already carries a `codecov note` comment
explaining why it cannot be hit.

---

> [!NOTE]
> **Low Risk**
> Documentation and deprecation only; instance listing behavior is unchanged and no license or evaluation logic is modified in this diff.
> 
> **Overview**
> Clarifies how **registered** function names relate to **license-gated** metadata: the instance `getRegisteredFunctionNames()` still lists every function in the instance registry (unchanged behavior), while `getAvailableFunctions()` / `getFunctionDetails()` stay the APIs for license-aware pickers.
> 
> **Deprecates** the static `HyperFormula.getRegisteredFunctionNames(code)` with JSDoc explaining the global registry vs instance scope and a migration path via `HyperFormula.buildEmpty({ language: '…' }).getRegisteredFunctionNames()`. Changelog and license/release docs are updated to match, including a retroactive 3.4.0 note that static `getAvailableFunctions()` / `getFunctionDetails()` were removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d8b9285c64cba26e2d0230df49895f29a9db11b7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---

### Open in the 2026-09-22 review round

- @sequba: "`getRegisteredFunctionNames` should still list ALL functions". Agreed on the principle — this method answers for the REGISTRY, and registration is not entitlement, while `getAvailableFunctions()` and `getFunctionDetails()` answer about availability and stay filtered. That removes most of this PR's substance, leaving the deprecation note and a docs correction. **Awaiting a decision on whether to reduce this PR to that or close it with the deprecation folded into #1730.**
- Fixed here meanwhile: `docs/guide/release-notes.md` now carries the same retroactive `### Removed` section for 3.4.0 that `CHANGELOG.md` gained in this PR's history. The two are mirrors and only one had it.


qunabu:

Tobiadefami (commented):
Reviewed at dcf93543 together with the paired tests at a406fc76. I checked the instance registry and translation snapshots, license filtering, aliases, custom and protected functions, raw translation behavior, static API compatibility and deprecation, documentation, changelog, and current automated findings. The focused paired run passes 17 suites (317 tests), TypeScript passes, and all current engine checks are green. I found no additional material issue in this PR.

docs/guide/license-key.md
```diff
@@ -80,10 +80,12 @@ If your key grants only part of the library, then:
   [error value](types-of-errors.md). Everything else in the sheet keeps calculating.
 * An API method your key doesn't include throws a `LicenseCapabilityMissingError` when you call
   it. Getters never throw; `copy()` and `cut()` do, because they belong to the clipboard feature.
-* [`getAvailableFunctions()`](../api/classes/hyperformula.md#getavailablefunctions) and
-  [`getFunctionDetails()`](../api/classes/hyperformula.md#getfunctiondetails) describe only the
-  functions your key includes, so a function picker built from them never offers a function that
-  then fails.
+* [`getAvailableFunctions()`](../api/classes/hyperformula.md#getavailablefunctions),
+  [`getFunctionDetails()`](../api/classes/hyperformula.md#getfunctiondetails) and the instance
+  [`getRegisteredFunctionNames()`](../api/classes/hyperformula.md#getregisteredfunctionnames)
```
sequba, line 67: `getRegisteredFunctionNames` should still list ALL functions

sequba (changes requested):

docs/guide/license-key.md
marcin-kordas-hoc: @sequba Agreed on the principle, and the distinction is a sound one: this method answers for the *registry*, and registration is not entitlement. `getAvailableFunctions()` and `getFunctionDetails()` answer about availability and stay filtered — you did not touch those, and I read that as deliberate.

The consequence is that this PR loses most of its substance. It does exactly two things: deprecate the static overload, and make the instance method filter by licence. Dropping the filter leaves the deprecation note and a docs correction, roughly 20 lines of the 45.

Do you want it reduced to that, or closed with the deprecation folded into #1730?

CHANGELOG.md
```diff
@@ -38,6 +43,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Added an Indonesian (Bahasa Indonesia) language pack. [#1674](https://github.com/handsontable/hyperformula/pull/1674)
 - Added a `stringifyCurrency` config option that lets you plug in a custom currency formatter for the `TEXT` function. [#1145](https://github.com/handsontable/hyperformula/issues/1145)
 
+### Removed
+
+- **Breaking change**: Removed the static `HyperFormula.getAvailableFunctions()` and `HyperFormula.getFunctionDetails()` methods. Use the instance methods of the same names, added in this release: a static method has no engine, and therefore no configuration, in scope, so it can only answer for the package as a whole. This entry was missing from the 3.4.0 release notes and is added retroactively. [#1724](https://github.com/handsontable/hyperformula/pull/1724)
```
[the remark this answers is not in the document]
marcin-kordas-hoc: Fixed. `docs/guide/release-notes.md` now carries the same `### Removed` section for 3.4.0 that `CHANGELOG.md` has, so the retroactive note about the static function-metadata methods is visible on the docs site too.

For the record, this was ours rather than pre-existing: neither file has that section on `develop`, and the retroactive entry was added to `CHANGELOG.md` only, in this PR's own history.

CHANGELOG.md
```diff
@@ -10,14 +10,26 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+
+### Fixed
+
+- Fixed the `AVERAGEIF` function returning a division-by-zero error when the calculated average was `0`. [#1733](https://github.com/handsontable/hyperformula/pull/1733)
+- Fixed the localized names of `VSTACK` and `HSTACK` in 14 language packs to match Microsoft Excel. [#1748](https://github.com/handsontable/hyperformula/pull/1748)
+- Fixed the MAXPOOL and MEDIANPOOL functions throwing an uncaught `TypeError` instead of returning the `#VALUE!` error when the range dimensions are not a whole multiple of the window size and the stride. [#1718](https://github.com/handsontable/hyperformula/pull/1718)
+- Fixed the `MOD` function returning a remainder with the sign of the dividend instead of the sign of the divisor, which made the results differ from Excel and Google Sheets for arguments with opposite signs (e.g. `=MOD(-3, 12)` now returns `9` instead of `-3`). [#1747](https://github.com/handsontable/hyperformula/issues/1747)
```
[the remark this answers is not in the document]
marcin-kordas-hoc: Not applicable to this PR. That `MOD` line is already on `develop` — CHANGELOG.md line 15, introduced by #1748 (`50e7170cd`). It appears in this diff as context, not as a change this PR makes, so the PR-link convention has nothing to correct here.

docs/guide/license-key.md
sequba: I want it reduced to that.

marcin-kordas-hoc: @sequba Done. Only the static method's deprecation remains.

CHANGELOG.md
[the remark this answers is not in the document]
marcin-kordas-hoc: Correction: the fix was to remove the entry, not mirror it — those static methods never shipped (added and removed within 3.4.0). Entry withdrawn, `release-notes.md` reverted.

PR #1754, opened 2026-08-28
[BREAKING CHANGE] Make INDEX Excel-compatible for zero and omitted indices (HF-224)

sequba:
### Context

[HF-224](https://app.clickup.com/t/9015210959/HF-224). A customer reported that `INDEX(range, row_num, 0)` does not return the entire row as it does in Excel. The task asks for INDEX to be made Excel-compatible as far as is possible without major architectural changes.

`known-limitations.md` recorded the gap: *"The INDEX function doesn't support returning whole rows or columns of the source range – it always returns the contents of a single cell."* This pull request removes that limitation and fixes the other Excel divergences found in INDEX along the way.

**This is a breaking change and ships with a migration guide** — [`docs/guide/migration-from-3.x-to-4.0.md`](https://github.com/handsontable/hyperformula/blob/feature/HF-224-index-excel-compatible/docs/guide/migration-from-3.x-to-4.0.md). Formulas that work today can change result: most importantly `=INDEX(A1:C3, 2)`, which returned the value of `A2` and now returns `#REF!`, because Excel requires a one-dimensional range when the column argument is left out.

### Verified against real Excel

The behaviour was checked formula by formula in Excel by the maintainer, using `ERROR.TYPE` so that a localized Excel still gives an unambiguous answer. **All fifteen probed formulas now agree.** Two of them disagreed at first and the implementation was corrected:

| Formula | Excel | this branch |
|---|---|---|
| `=INDEX(A1:C3, 9, 1)` | `#REF!` | `#REF!` |
| `=INDEX(A1:C3, 1, 9)` | `#REF!` | `#REF!` |
| `=INDEX({1,2,3;4,5,6;7,8,9}, 9, 1)` | `#REF!` | `#REF!` (array form does not differ) |
| `=INDEX(A1:C3, -1, 1)` | `#VALUE!` | `#VALUE!` |
| `=INDEX(A1:C3, -0.5, 1)` | `#VALUE!` | `#VALUE!` (the sign is judged before truncation) |
| `=INDEX(A1:C3, 2.9, 1)` | `4` | `4` |
| `=INDEX(A1:C1, 3)` | `3` | `3` |
| `=INDEX(A1:C1, 3, )` | `#REF!` | `#REF!` ← **corrected**, previously `3` |
| `=INDEX(A1:C1, 4)` | `#REF!` | `#REF!` |
| `=INDEX(A1:A3, 2)` | `4` | `4` |
| `=INDEX(A1:C3, 2, 2)` | `5` | `5` |
| `=INDEX(A1:C3, 2, 0)` | `4, 5, 6` | `4, 5, 6` |
| `=INDEX(A1:C3, 0, 2)` | `2; 5; 8` | `2; 5; 8` |
| `=INDEX(A1:C3, 0, 0)` | the whole range | the whole range |
| `=INDEX(A1:C3, 2)` | `#REF!` | `#REF!` ← **corrected**, previously the whole row |
| `=SUM(INDEX(A1:C3, 0, 1))` | `12` | `12` |

The rule the data supports, and what is now implemented: **leaving `column_num` out is not the same as passing `0`.** Left out, Excel requires the range to be a single row or a single column and reads the only index as the position along it; given several rows and several columns there is nothing for that index to mean and the answer is `#REF!`. Left *empty* (`=INDEX(A1:C1, 3, )`) it is a column number of zero, which asks for a whole row.

One measured divergence remains: with array arithmetic enabled Excel returns `1, 4` for `=INDEX(A1:C3, {1,2}, 1)` and HyperFormula returns `1`. INDEX declares `vectorizationForbidden`, as every other array-output function does, because a vectorized call evaluates the function once per element and rejects an array result — without it `=INDEX(A1:C3, {1,0}, 1)` crashed the engine with `Error: Function returning array cannot be vectorized.` This is recorded in [list-of-differences.md](https://github.com/handsontable/hyperformula/blob/feature/HF-224-index-excel-compatible/docs/guide/list-of-differences.md) and in the migration guide.

**How array results work.** A result spanning more than one cell is an array, so the sheet has to reserve space for it before the formula is evaluated. `INDEX` declares a `sizeOfResultArrayMethod`, and the shape is derived from the formula alone: `indexArraySize` and `index` share one `resolveIndexArguments` helper, so the predicted shape cannot disagree with the value returned later. A shape that does not follow from the formula — a computed index, a named expression, an unbounded range — is predicted as a single cell: `=SUM(INDEX(A1:C3, B1, 0))` still sums the whole row, but `=INDEX(A1:C3, B1, 0)` on its own returns `#VALUE!`.

**Not implemented**, because both are blocked by the parser rather than by INDEX (see [known limitations](https://github.com/handsontable/hyperformula/blob/feature/HF-224-index-excel-compatible/docs/guide/known-limitations.md#index-function)):

- Excel's reference form `INDEX(reference, row_num, column_num, area_num)` — HyperFormula has no multi-area references (`(A1:B1,A2:B2)` is a parsing error).
- INDEX as a range endpoint — `=SUM(A1:INDEX(A1:A10, 5))` is a parsing error; only a cell reference or `OFFSET` may sit there.

- Paired test pull request: handsontable/hyperformula-tests#47 — 59 cases, including one per row of the Excel table above.
- Full private suite: `npx jest` → 502 suites, 6264 tests, 0 failures. No test outside the INDEX spec needed a change.
- `npx tsc --noEmit` clean; `npx eslint` reports 0 errors. Codecov: all modified and coverable lines covered.
- `npm run docs:generate-function-docs` regenerates cleanly; the generated syntax line is `INDEX(range, row, [column])`.
- Engine behaviour was established by experiment, not by reading: a runtime array larger than the predicted `ArraySize` throws `Error('Resizing to smaller array')` (which the shared helper prevents); a vectorized array result throws `Error('Function returning array cannot be vectorized.')` (which `vectorizationForbidden` prevents); an unbounded range's predicted dimension is `Number.POSITIVE_INFINITY` and `ArrayValue.resize` will not grow toward it, which silently truncated results until commit 7; `addRows`/`addColumns`/`removeColumns` resize the INDEX array vertex as they do for `TRANSPOSE`.
- Worth knowing for review: **no benchmark in `performance/` exercises INDEX**, so the perf job cannot see this change; its deltas on this branch are noise.

### Types of changes

- [x] Breaking change (a fix or a feature because of which an existing functionality doesn't work as expected anymore)

### Related issues:
1. [HF-224](https://app.clickup.com/t/9015210959/HF-224)
2. Paired tests: handsontable/hyperformula-tests#47

### Checklist:
- [x] I described my changes in the [CHANGELOG.md](https://github.com/handsontable/hyperformula/blob/master/CHANGELOG.md) file.
- [x] My changes require a migration guide.

**Note on versioning.** The migration guide is written as 3.x → 4.0 and registered in the docs sidebar next to the existing ones. If these changes are meant to ship in a minor release instead, the file needs renaming and the four changelog links updating — say so and I will.

The decisions and the engine constraints behind them are recorded in the ADR in the paired pull request (`dev_docs/2026-08-28-index-excel-compatibility.md`).


qunabu:
Task linked: [HF-224 Make INDEX function Excel-compatible](https://app.clickup.com/t/9015210959/HF-224)

PR #1757, opened 2026-08-31
HF-359: Turn the repository into a monorepo, and rebuild the developer documentation

sequba:
### Context

HF-359. Turns the repository into a monorepo, and replaces the single always-loaded `DEV_DOCS.md` with a documentation system that survives having more than one package in the tree.

```
hyperformula/      the engine (published)   — src/, test/, its toolchain, dev-docs/, script/
hyperformula-ui/   placeholder              — imported separately, preserving history
docs/              the portal               — its own package, installs on its own
script/            release/ and the licence gate — the only scripts that span the repo
dev-docs/          repository-wide standards and process
```

**No changes under `hyperformula/src/`.** Every file there is byte-identical to its old location at `src/`; git records the moves as renames throughout, so history follows. The published package's name, version, entry points and `exports` map are unchanged, and the publish-package check confirms the tarball contents.

## The monorepo

- **npm workspaces**, with a private root carrying fan-out scripts. **Linting is the deliberate exception** — it runs once from the root over the whole repository, so nothing between packages falls through the gap.
- **`docs/` left the workspace.** VuePress 1.x and its `--openssl-legacy-provider` tree must never reach an engine install. `npm run docs:install` installs it; CI installs it before building the portal.
- **The portal owns its deployment**: `wrangler.jsonc` and `worker/` moved under `docs/`, and the asset directory is now `./.vuepress/dist`.
- **The packages release together, on one version, from one changelog.** A release cuts every published package at the same version whether or not its code moved, in exchange for one number that identifies the state of the whole repository. `CHANGELOG.md` stays at the root.
- **The private test suite stays branch-matched**; only its checkout path moved. Its 501 spec files needed no edit — they import the engine relatively, and because `src/` and `test/` moved *together*, the depth from a spec to the package root is unchanged.

### Everything lives with what owns it

| | Repository level | Package level |
|---|---|---|
| `dev-docs/` | definition of done, code style, testing standards, documentation rules, build and release, pull requests, worktrees, agent setup | `hyperformula/dev-docs/` — architecture, parser, interpreter, dependency graph, function catalogue, i18n, performance, its suites, its build |
| `script/` | `release/`, `check-licenses.mjs` | `hyperformula/script/` — build checks; `docs/script/` — the generator and the Worker assets |

Three `dev-docs` pages were genuinely mixed and were **split, not moved**: `TESTING` (what a change must prove stays; the suites go down), `BUILD` (workspace, deploy and release stay; the intermediate build and packaging go down), and `CODE-STYLE` (style stays; the hot-path table becomes `hyperformula/dev-docs/PERFORMANCE.md`).

The `script/` split was decided by the callers rather than by judgement: every script had exactly one caller and the caller named its owner. Twelve call sites stopped reaching across a package boundary.

Root `examples/` is **deleted** — three PNGs were byte-identical duplicates of `docs/.vuepress/public/`, which is what the site actually serves, and the other five files were referenced nowhere.

## The documentation

`DEV_DOCS.md` said "start by reading entire DEV_DOCS.md". Every agent working anywhere loaded the docs-portal conventions, the i18n sources and the function-catalogue rules before it could look at a parser bug. Three layers replace it:

| Layer | Answers | Loaded |
|---|---|---|
| `AGENTS.md` | *What is this directory, and where do I look next?* | Always, within its subtree |
| `dev-docs/` | *How does this work and why?* | On demand |
| `.claude/skills/` | *How do I do task X?* | On skill trigger |

**`dev-docs/` is the single source of truth**, and nothing outside it restates what is in it: every `AGENTS.md`, `README.md` and `SKILL.md` carries only what is so specific to its own directory that it would be useless elsewhere, and links for the rest.

| | Before | After |
|---|---|---|
| Always-loaded `AGENTS.md` | ~610 lines | **147 lines** across 12 files |
| Skills | ~1100 lines | **378 lines** across 9 |

Twelve `AGENTS.md` files, each with a `CLAUDE.md` symlink so Claude Code and Cursor read one source. Nine skills, five scoped by a `paths` glob. `.claude/settings.json` enables `typescript-lsp` repository-wide and denies agent reads of build artifacts — but not of generated *documentation*, which is meant to be read. `.worktreeinclude` records what must never be copied into a worktree and why. `.ai/` holds one sentence pointing at `dev-docs/`, for agents that look there by habit.

`DOCS_CONTENT_GUIDE.md` moved in as `dev-docs/DOCS-CONTENT-GUIDE.md`.

## How did you test your changes?

| Command | Result |
|---|---|
| `npm ci` | Clean; the workspace resolves |
| `npm run lint` | **0 errors**, 54011 warnings — the exact pre-move baseline |
| `npm run test:jest` | **502 suites, 6232 passed**, 3 skipped, 0 failed |
| `npm run test:ci -- --coverage` | **97.42%**, written to `hyperformula/coverage` |
| `npm run test:browser` | **12464 assertions** across headless Chrome and Firefox, exit 0 |
| `npm run verify:typings` | `tsc --noEmit` clean |
| `npm run bundle-all` | Every bundle, and **`Publish package check: OK`** |
| `npm run check:licenses` | 3 production dependencies, and it **fails** when the allowlist is tightened |
| `npm run release code-freeze …` (dry run) | Correct version, dates and changelog section |
| `npm run docs:generate-function-docs` | 564 lines, 13 category sections |
| `npm --prefix docs run build:cf` | **217 pages**, sitemap, `_headers` and `_redirects` copied |

Every relative Markdown link in `AGENTS.md`, both `dev-docs/` levels, the skills and the READMEs was resolved against the filesystem. All resolve.

## What the reviews found

Four rounds, 58 findings. The ones worth knowing about:

**Flag arguments were being swallowed.** Every delegating root script dropped them, so `npm run test:ci -- --coverage` reached the workspace as bare `jest --forceExit` and CI had been uploading nothing to Codecov while staying green. Demonstrated with `npm run test:jest -- --listTests`, which ran all 6232 tests instead of listing them.

**The release script was never adapted.** It read and wrote `./package.json` — now the private root manifest with no `version` — and published from the root, where npm refuses a private package. Two dry runs surfaced two further paths no review had caught.

**`check:licenses` was checking nothing.** `license-checker` reads `<start>/node_modules`, and npm hoists, so from `hyperformula/` it saw an empty tree and from the root a manifest with no dependencies — zero packages either way, passing vacuously. Replaced with `script/check-licenses.mjs`, driven by `npm ls --omit=dev`.

**The portal was publishing internal agent instructions.** VuePress sweeps every markdown file under `docs/`, so `docs/AGENTS.md` and its symlink were being built into `/AGENTS.html` and an identical `/CLAUDE.html`, sitemap and search index included.

**`npm run test:browser` exited non-zero after passing.** The cause was `client.clearContext: false` in the Karma base config, and it reproduced on the pre-move tree. `concurrency: 1` was the obvious first guess and is wrong — it doubles wall-clock time and makes the failure deterministic rather than removing it.

**An audit traced the deleted `DEV_DOCS.md` into the new tree.** Almost nothing was lost; what it found was new prose that had invented things. `dev-docs/README.md` stated a link convention that every one of the 33 links in the directory violated, and `I18N.md` claimed argument separators are part of the language package when they are `ConfigParams`.

One finding was **rejected on measurement** — a claimed ESLint timeout risk, measured at 0.91s and 1.06s.

## Two pre-existing problems this uncovered

**The source language packs have never been linted.** The old ignore list carried a bare `languages` entry meant for the build output. An unanchored pattern matches a directory of that name at *any* depth, so it also excluded `src/i18n/languages/` — and the `sort-keys` override targeting exactly those files has never run. Anchoring the build-output entry exposed **881 violations** across 19 translation files. Confirmed against the pre-move tree, where ESLint reports `File ignored because of a matching ignore pattern`.

They are excluded again in `.eslintignore`, deliberately and with a comment, so this branch keeps the exact pre-move lint baseline. Sorting 19 translation files is a change of its own — and reordering translation entries is what the i18n guidance tells contributors not to do casually.

**`npm run test:compatibility` was broken for everyone**, pointing at a path in neither repository. Fixed here, since the branch had to touch it anyway.

## Not done, on purpose

**CI path filtering.** A naive `paths:` filter on a workflow that branch protection lists as a required check leaves it permanently pending and makes pull requests unmergeable. Doing it safely needs the required-checks list, which lives in repository settings, plus the `dorny/paths-filter`-and-single-gate shape. The CI *paths* are all updated; only the filtering is outstanding.

**Importing `hyperformula-ui`** — left for the teammate who owns the formula-builder move. It is deliberately *not* in the `workspaces` array until then: npm silently ignores an entry with no `package.json`.

Both are recorded in `dev-docs/STRUCTURE.md` under "What the move still owes", so they survive the pull request.

### Types of changes

Repository structure and tooling. The published package is unchanged.

### Related issues:
1. HF-359

### Checklist:
- [x] I have signed the [Contributor License Agreement](https://goo.gl/forms/yuutGuN0RjsikVpM2).
- [x] My change is compliant with the [OpenDocument](https://docs.oasis-open.org/office/OpenDocument/v1.3/os/part4-formula/OpenDocument-v1.3-os-part4-formula.html) standard. — not applicable; no engine behaviour changes.
- [x] My change is compatible with Microsoft Excel. — not applicable; no engine behaviour changes.
- [x] My change is compatible with Google Sheets. — not applicable; no engine behaviour changes.
- [ ] I described my changes in the [CHANGELOG.md](https://github.com/handsontable/hyperformula/blob/master/CHANGELOG.md) file. — **intentionally not done.** Nothing user-facing changed; the published package is identical.
- [x] My changes require a documentation update. — `dev-docs/` at both levels, `docs/README.md` and the public building guide are updated here.

### Companion pull request

handsontable/hyperformula-tests#51 gives the private test suite an `AGENTS.md` pointing here, and corrects three claims in its `README.md`.


qunabu:
Task linked: [HF-359 Add formula builder core to hyperformula repo](https://app.clickup.com/t/9015210959/HF-359)

dev-docs/BUILD.md
```diff
@@ -0,0 +1,62 @@
...
+
+`docs/guide/built-in-functions.md` is generated and git-ignored — never commit it, and never edit it by hand. Its source is the metadata catalogue; see [`FUNCTION-CATALOGUE.md`](FUNCTION-CATALOGUE.md).
+
+The portal is deployed as a Cloudflare Worker: `docs:build:cf` composes the assets through `script/prepare-cf-assets.js`, `docs:deploy:cf` and `docs:preview:cf` call `wrangler` against [`wrangler.jsonc`](../wrangler.jsonc). The Worker entry point is [`worker/index.js`](../worker/index.js).
+
+## Release
+
+`npm run release` runs `script/release/release.sh`. Releasing is maintainer-owned; do not invent steps around it.
+
+`npm run check:licenses` asserts that every production dependency carries a permissive licence.
```
sequba: This page should not duplicate information from `/docs/guide/building.md`. Keep here only the things that are not included there

dev-docs/MONOREPO.md
```diff
@@ -0,0 +1,109 @@
...
+
+| Package | Directory | Purpose | Published |
+|---|---|---|---|
+| `hyperformula` | `hyperformula/` | The calculation engine. Everything in `src/` and `test/` today. | yes |
+| `@hfe/core` | `hyperformula-ui/packages/core/` | Formula editor UI: reference highlighting, inline editor, function help. | yes |
+| `@hfe/handsontable-adapter` | `hyperformula-ui/packages/handsontable-adapter/` | Binds the editor to Handsontable. | yes |
+| `@hfe/plain-table-adapter` | `hyperformula-ui/packages/plain-table-adapter/` | Binds the editor to a plain HTML table. | yes |
+| `@hfe/tanstack-table-adapter` | `hyperformula-ui/packages/tanstack-table-adapter/` | Binds the editor to TanStack Table. | yes |
+| `hyperformula-docs` | `docs/` | The VuePress documentation portal. | no |
+| — | `hyperformula-ui/demo/`, `hyperformula-ui/e2e/` | Demo app and end-to-end tests for the UI packages. | no |
```
sequba: for now, there will be just one package named "hyperformula-ui" - a collection of UI components for better UI experience working with hyperformula. Remove other packages from this table

dev-docs/MONOREPO.md
```diff
@@ -0,0 +1,109 @@
...
+
+The HyperFormula agent skill stays in the shared `handsontable-skills` repository for now. Bringing it here is a separate decision, taken later and on its own terms; nothing in this layout depends on it.
```
sequba: Remove this paragraph

dev-docs/MONOREPO.md
```diff
@@ -0,0 +1,109 @@
...
+7. **Split `CHANGELOG.md` per package**, each keeping the current Keep a Changelog form. No fragment mechanism.
+8. **Give every package an `.nvmrc` saying `22`**, `docs/` included.
+9. **Keep the private test suite branch-matched.** `hyperformula-tests` stays keyed to this repository's branch name; only its checkout path moves, from `test/hyperformula-tests/` to `hyperformula/test/hyperformula-tests/`. Update `fetch-tests.sh`, the `test:setup-private` script, and `.gitignore` together, and re-check `.worktreeinclude`, which names the old path.
+10. **Path-filter CI.** Each package's jobs run only when its paths change; full runs on `develop`, `master`, and release branches.
+
+## Open questions
+
+- Does `examples/` move under `docs/` as well? Every reference to it comes from a guide in `docs/guide/`, so it has no other consumer, but it is not part of the deployment path the way the Worker is.
```
sequba: Remove these sections. We don't need to keep track of the decisions.

DEV_DOCS.md
sequba, line 1: Remove this file. The frontpage of the dev-docs is now dev-docs/README.md

docs/AGENTS.md
sequba, line 1: This file should be really thin. Just point to `dev-docs` dir as the most important source and to docs/README.md as the description of this directory contents. Move all the info from here to one of these 2 places.

script/AGENTS.md
sequba, line 1: This file should be really thin. Just point to `dev-docs` dir as the most important source and to script/README.md as the description of this directory contents. Move all the info from here to one of these 2 places.

hyperformula/src/DependencyGraph/AGENTS.md
sequba, line 1: This file should be really thin. Just point to `dev-docs` dir as the most important source. Move all the info from here to that directory.

hyperformula/src/i18n/AGENTS.md
sequba, line 1: This file should be really thin. Just point to `dev-docs` dir as the most important source. Move all the info from here to that directory.

hyperformula/src/interpreter/functionMetadata/AGENTS.md
sequba, line 1: This file should be really thin. Just point to `dev-docs` dir as the most important source. Move all the info from here to that directory.

hyperformula/src/interpreter/plugin/AGENTS.md
sequba, line 1: This file should be really thin. Just point to `dev-docs` dir as the most important source. Move all the info from here to that directory.

hyperformula/src/interpreter/AGENTS.md
sequba, line 1: This file should be really thin. Just point to `dev-docs` dir as the most important source. Move all the info from here to that directory.

hyperformula/src/parser/AGENTS.md
sequba, line 1: This file should be really thin. Just point to `dev-docs` dir as the most important source. Move all the info from here to that directory.

hyperformula/src/AGENTS.md
sequba, line 1: This file should be really thin. Just point to `dev-docs` dir as the most important source. Move all the info from here to that directory.

hyperformula/test/AGENTS.md
sequba, line 1: This file should be really thin. Just point to `dev-docs` dir as the most important source and to test/README.md. Move all the info from here to one of these 2 places.

dev-docs/BUILD.md
sequba: Done. `BUILD.md` now opens by pointing at `docs/guide/building.md` for the command reference, and keeps only what that guide omits: `npm ci` and the pinned Node version, the intermediate `lib/` build and `HF_COMPILE`, `clean`, the `languages/` output the guide leaves out of its format list, the generated documentation and the rule against committing it, the Cloudflare deployment path, release, and `verify:publish-package`. 74 lines down to 50, none of them a restatement.

dev-docs/MONOREPO.md
sequba: Done. The table is now three rows: `hyperformula`, `hyperformula-ui`, and `hyperformula-docs`. `hyperformula-ui` is a single package described as UI components for working with HyperFormula, and the tree no longer nests a `packages/` directory under it.

dev-docs/MONOREPO.md
sequba: Removed.

dev-docs/MONOREPO.md
sequba: Done — both "Why this shape" and "Open questions" are gone. `MONOREPO.md` is now the package table, the tree, and the migration steps, and the steps are actions rather than justifications. 75 lines.

DEV_DOCS.md
sequba: Deleted. `dev-docs/README.md` is the front page and carries what `DEV_DOCS.md` used to: the quick links, the index, and the pointers to `CONTRIBUTING.md` and the portal. The two remaining references to the old file, in `STRUCTURE.md` and `MONOREPO.md`, are gone too.

docs/AGENTS.md
sequba: Done — 10 lines, pointing at `docs/README.md`, `dev-docs/DOC-STANDARDS.md`, and `DOCS_CONTENT_GUIDE.md`. The portal mechanics (what is generated, how to run it, how to add a guide and register it in the sidebar) moved into `docs/README.md`; the documentation rules moved into `dev-docs/DOC-STANDARDS.md`.

script/AGENTS.md
sequba: Done — 7 lines. `script/README.md` did not exist, so it is new: the per-script table and the two things true only there (these run in CI as well as locally, and nothing here is covered by tests). The `claude/` hook’s design constraints moved to the new `dev-docs/AGENT-TOOLING.md`, which is also where `.claude/settings.json` and the skills conventions are now documented.

hyperformula/src/DependencyGraph/AGENTS.md
sequba: Done — 6 lines, pointing at the new `dev-docs/DEPENDENCY-GRAPH.md`, which holds the file table, the rules (ranges stay collapsed, mappings stay consistent, cycles are values, `EmptyCellVertex` is not "no vertex") and the testing note.

hyperformula/src/i18n/AGENTS.md
sequba: Done — 6 lines, pointing at `dev-docs/I18N.md`, which absorbed the rules, the language-pack procedure, and the testing note. The translation-source table was already there.

hyperformula/src/interpreter/functionMetadata/AGENTS.md
sequba: Done — 6 lines, pointing at `dev-docs/FUNCTION-CATALOGUE.md`, which already held the long-form explanation and now also carries the short rule list this file used to duplicate.

hyperformula/src/interpreter/plugin/AGENTS.md
sequba: Done — 8 lines, pointing at `dev-docs/INTERPRETER.md#built-in-functions`. The plugin contract, `runFunction`, both metadata tables, the "never" list, and the five places a function change must touch all moved there.

hyperformula/src/interpreter/AGENTS.md
sequba: Done — 6 lines, pointing at the new `dev-docs/INTERPRETER.md`, which covers the interpreter and the built-in function contract in one page.

hyperformula/src/parser/AGENTS.md
sequba: Done — 6 lines, pointing at the new `dev-docs/PARSER.md`, which holds the file table and the rules (the lexer is language-dependent, parsing and unparsing move together, the cache is load-bearing).

hyperformula/src/AGENTS.md
sequba: Done — 10 lines, pointing at `dev-docs/ARCHITECTURE.md`, `CODE-STYLE.md`, and `DEFINITION-OF-DONE.md`. The engine-wide invariants and the notes on the files without their own `AGENTS.md` moved into `ARCHITECTURE.md`.

hyperformula/test/AGENTS.md
sequba: Done — 9 lines, pointing at `test/README.md` and `dev-docs/TESTING.md`. How to write a case and what each kind of change must cover moved into `TESTING.md`, which also absorbed the `test-writing-discipline` skill’s rules; `TESTING.md` now points at `test/README.md` for the fetch mechanics instead of restating them.

AGENTS.md
```diff
@@ -1,10 +1,34 @@
...
+| Function-name translations | [`dev-docs/I18N.md`](dev-docs/I18N.md) |
+| Tests | [`dev-docs/TESTING.md`](dev-docs/TESTING.md) |
+| The documentation portal | [`dev-docs/DOC-STANDARDS.md`](dev-docs/DOC-STANDARDS.md) |
+| Building, bundling, releasing | [`dev-docs/BUILD.md`](dev-docs/BUILD.md) |
+| What a change must include before review | [`dev-docs/DEFINITION-OF-DONE.md`](dev-docs/DEFINITION-OF-DONE.md) |
+| Style and performance | [`dev-docs/CODE-STYLE.md`](dev-docs/CODE-STYLE.md) |
+| A linked git worktree | [`dev-docs/WORKTREES.md`](dev-docs/WORKTREES.md) |
+| How this repository is set up for agents | [`dev-docs/AGENT-TOOLING.md`](dev-docs/AGENT-TOOLING.md) |
+| The monorepo migration | [`dev-docs/MONOREPO.md`](dev-docs/MONOREPO.md) |
+| Step-by-step task workflows | [`.claude/skills/`](.claude/skills/) |
```
sequba: This table of content belongs in dev-docs/README.md

AGENTS.md
```diff
@@ -1,10 +1,34 @@
...
+
+Each directory's own `AGENTS.md` points at the page that covers it, and loads automatically when you work there.
```
sequba: Remove this line

.claude/skills/changelog-creation/SKILL.md
```diff
@@ -0,0 +1,31 @@
...
+| [`DOC-STANDARDS.md`](../../../dev-docs/DOC-STANDARDS.md#the-changelog) | Which section to use, the bullet format, and how to write the text |
+| [`PULL-REQUESTS.md`](../../../dev-docs/PULL-REQUESTS.md#order-of-operations) | Why the entry comes after the pull request, not before |
+
+## 2. Decide whether an entry is required at all
+
+Documentation-only, test-only, and CI or tooling changes take none, and neither does a bug that was introduced and never released.
+
+## 3. Open the pull request first
+
+The entry carries a link, so it needs the number. Do not guess it — read it from the URL. Skill `pr-creation`.
```
sequba: Each entry must have a GitHub link. GH issue if exist, otherwise a GH pull request link

dev-docs/STRUCTURE.md
```diff
@@ -0,0 +1,68 @@
+# Repository structure
```
sequba, line 1: This file duplicates info from MONOREPO.md. Can we merge them?

AGENTS.md
sequba: Moved. The table now lives in `dev-docs/README.md`, where it **replaced** the categorised file listing rather than sitting next to it — the two were the same information keyed differently, which is the duplication the single-source rule exists to prevent. Every page is still reachable from it, plus the pointers to `docs/README.md`, `test/README.md`, `script/README.md` and `CONTRIBUTING.md`. `AGENTS.md` is down from 68 to 48 lines and no longer says which page covers what.

AGENTS.md
sequba: Removed. The part of it worth keeping — that a directory's own `AGENTS.md` loads automatically when you work there — is folded into the sentence pointing at `dev-docs/README.md`, so it sits with the routing advice instead of trailing a table that has moved.

.claude/skills/changelog-creation/SKILL.md
sequba: Stated outright now, at the step where the number is obtained: "**Every entry ends with a GitHub link**: the public issue it fixes when one exists, otherwise the pull request. So the entry needs a number that only exists once the pull request is open." Step 4 then just says to end the bullet with the link from step 3, rather than re-deriving the rule two steps after it mattered. The long form stays in `dev-docs/DOC-STANDARDS.md#the-changelog`.

dev-docs/STRUCTURE.md
sequba: Merged. `STRUCTURE.md` now covers where everything lives today, and a "Where it is going" section carries the packages, the target tree and the migration steps. `MONOREPO.md` is deleted and its two inbound links (from `STRUCTURE.md` itself and from `WORKTREES.md`) repointed. Both files having a tree was the real problem — a reader had to already know which of the two was current.

qunabu:
**Overall:** Solid as a one-package monorepo, and it fixes real pre-existing bugs along the way. But the structure isn't yet proven for a *second* package, which is its entire purpose. I compared the branch against `formula-builder@master`, and the gap is bigger than one entry in the `workspaces` array.

**TL;DR:** `formula-builder` is pnpm + `catalog:` + Biome + vitest + rspack + ESM + TS 5.6, in **5 publishable packages** plus demos and a Playwright e2e. This PR standardizes on npm workspaces, root ESLint 8, jest 26/karma, webpack 4, TS 4.0.8, commonjs, and reserves **one** directory. Before merging, I'd want a second workspace actually landed on this branch to prove `npm ci` / lint / typecheck / test / build hold — plus written decisions on versioning and licensing. Not a block; don't merge on faith either.

1. **`hyperformula-ui/` is one directory; five packages are coming.** `@hfe/core` + four adapters (ag-grid, MUI, plain-table, tanstack), plus `demo/*` and `e2e`. Root says `workspaces: ["hyperformula"]`, the directory name is singular, and `STRUCTURE.md` says the package "keeps the scope it publishes under today" — i.e. `@hfe/*`. Decide now: nested `hyperformula-ui/*` or flat, and where demos and Playwright go. Otherwise the import reshapes the structure this PR freezes.

2. **`catalog:` will hard-fail `npm ci`.** Every dependency in `formula-builder` is `"catalog:"`, a pnpm/bun-only protocol. Import means rewriting all specs, plus deciding Biome-vs-ESLint and a second test runner in CI. Worth costing as its own task in "what the move still owes" — today it reads as a one-line change.

3. **Root-only linting won't survive the second package.** `.eslintrc.js` hardwires `parserOptions.project: './hyperformula/tsconfig.json'` with `createDefaultProgram: true`, and `npm run lint` is `eslint . --ext .js,.ts` — no `.tsx`. UI files either go unlinted or get type-aware linting outside their own TS program. `eslint-plugin-license-header` will demand a GPL header in every UI file, and root TS `^4.0.8` conflicts with the TS 5.6 the UI needs in one hoisted tree. This argues for per-package lint/typecheck with root fan-out — the opposite of the decision taken here.

4. **The rationale for excluding `docs/` contradicts the plan.** "A large old dependency tree must not reach an engine install" — but UI + four adapters pull React, Vue, ag-grid 36, MUI, handsontable 18, sass, Playwright: heavier than VuePress, and it *will* be in the workspace. Either the same rule applies to the UI, or state plainly that the engine install gets fat.

5. **Every root script is pinned to one workspace** (`--workspace=hyperformula`). Once the UI lands, `npm test` at the root still tests only the engine, silently — the same failure class as the swallowed flags this PR caught. `--workspaces --if-present`.

6. **One version, one changelog** is the biggest *product* risk, not a technical one. Core is `0.0.1` today with `peerDependencies: hyperformula ^3.0.0`; after the move its first public version is 3.4.x, every engine patch republishes it, and a UI breaking change forces an engine major or smuggles a break into a minor. Fine as a deliberate choice — but it can't be undone after the first public release.

7. **Licensing needs an explicit decision.** Root and engine are `GPL-3.0-only` in a public repo. The UI lives in a private repo, is flagged Premium, and its `package.json` has no `license` field. A contributor is already asking publicly when the formula builder ships. `check-licenses.mjs` will also start seeing the adapters' prod deps.

8. **Merge logistics.** Draft, `mergeable_state: dirty`, 339 files, against **38 open PRs** that will all take path conflicts. Pick a window, land it in one shot, warn contributors. GitHub blame breaks at the rename commit (`--follow` works locally).

9. **Small fixes for this PR:**
   - `.claude/settings.json` deny rules are leading-slash (`Read(/hyperformula/lib/**)`), which reads as absolute-from-filesystem-root — likely matching nothing. Try `hyperformula/lib/**`.
   - `.eslintignore`: `script` and `doc` are unanchored, matching at any depth — exactly the `languages` trap this PR just uncovered.
   - `docs/package.json` carries `version: 3.4.0`; the release script only moves `hyperformula/package.json` (`PKG_DIR`, `on_npm "hyperformula@$1"`). Under one-version that's a third manifest to bump, not a second.

qunabu (changes requested):
see my comment above

hyperformula/package.json
```diff
@@ -0,0 +1,614 @@
...
+    "env-cmd": "^10.1.0",
+    "eslint": "^8.57.1",
+    "eslint-config-prettier": "^9.1.0",
+    "eslint-plugin-jasmine": "^4.2.2",
+    "eslint-plugin-jest": "^27.9.0",
+    "eslint-plugin-jsdoc": "^50.5.0",
+    "eslint-plugin-license-header": "^0.6.1",
+    "eslint-plugin-prettier": "^5.2.1",
+    "exceljs": "^4.4.0",
+    "jasmine": "^5.4.0",
```
budnix, line 139: This is why every `build` job is red right now — `hyperformula/.config/webpack/languages.js:10` does `require('fs-extra')` and nothing in the repo declares it. On develop it resolved by luck: the VuePress deps sat in the root manifest and npm hoisted `fs-extra@7.0.1` to the top. Now that the portal installs from its own `docs/package.json` there is no top-level copy after `npm ci`, and `bundle:languages` dies with `Cannot find module 'fs-extra'` on every OS and Node version. `@babel/helper-plugin-utils` in `.config/babel/add-import-extension.js` is undeclared the same way — it still resolves today, but worth declaring both.

```suggestion
    "exceljs": "^4.4.0",
    "fs-extra": "^11.4.0",
    "jasmine": "^5.4.0",
```

script/check-licenses.mjs
```diff
@@ -0,0 +1,112 @@
...
+      offenders.push(`${id} — not found in the licence report`)
+      continue
+    }
+
+    const declared = String(entry.licenses ?? '')
+    const parts = declared.replace(/[()]/g, '').split(/\s+OR\s+|\s+AND\s+/).map(s => s.replace(/\*$/, '').trim())
+
+    if (!parts.some(part => ALLOWED.has(part))) {
+      offenders.push(`${id} — ${declared || 'no licence declared'}`)
+    }
```
budnix, line 102: `AND` and `OR` are split the same way here, so `Apache-2.0 AND LGPL-3.0` passes on the `Apache-2.0` half alone — and `AND` means both licences apply, which is the one case this check exists to catch. The other half is that `license-checker` returns `licenses` as an array when a package declares several, and `String(['MIT','ISC'])` becomes `"MIT,ISC"`, which matches nothing and fails a package that is fine. Splitting on `OR` first and requiring every `AND` part to be allowed handles both, though it still will not read a nested expression like `LGPL-3.0 AND (MIT OR Apache-2.0)` correctly.

```suggestion
    const raw = entry.licenses ?? ''
    const declared = Array.isArray(raw) ? raw.join(' OR ') : String(raw)
    const allowed = declared.replace(/[()]/g, '').split(/\s+OR\s+/)
      .some(alt => alt.split(/\s+AND\s+/).every(part => ALLOWED.has(part.replace(/\*$/, '').trim())))

    if (!allowed) {
      offenders.push(`${id} — ${declared || 'no licence declared'}`)
    }
```

.github/workflows/performance.yml
```diff
@@ -34,16 +34,16 @@ jobs:
         with:
           ssh-key: ${{ secrets.DEPLOY_TOKEN }}
           repository: handsontable/hyperformula-tests
-          path: test/hyperformula-tests
+          path: hyperformula/test/hyperformula-tests
 
       - name: Fetch hyperformula-tests and sync branches
-        run: cd test && ./fetch-tests.sh
+        run: cd hyperformula/test && ./fetch-tests.sh
```
budnix, line 40: This is failing on the PR right now with `./fetch-tests.sh: No such file or directory` and exit 127. For a pull request the workflow file comes from the head branch, but the checkout above it uses `base.sha`, and a base commit from before the move has no `hyperformula/test/`. The `--workspace=hyperformula` benchmark call below would break the same way. It heals once develop carries the move, so the question is whether one red window is fine or the base steps need a fallback.

script/check-licenses.mjs
```diff
@@ -0,0 +1,112 @@
...
+  }
+  return found
+}
+
+const production = new Set()
+
+for (const workspace of WORKSPACES) {
+  collect(json('npm', ['ls', `--workspace=${workspace}`, '--omit=dev', '--all', '--json']), production)
+  production.delete(`${workspace}@${json('npm', ['pkg', 'get', 'version', `--workspace=${workspace}`, '--json'])[workspace]}`)
+}
```
budnix, line 71: This delete spawns an extra `npm pkg get version` just to build an id that the loop right below already removes by prefix. Since that loop drops every id starting with `hyperformula@`, this line can never catch anything it would miss. Dropping it saves a subprocess on each run.

```suggestion
for (const workspace of WORKSPACES) {
  collect(json('npm', ['ls', `--workspace=${workspace}`, '--omit=dev', '--all', '--json']), production)
}
```

hyperformula/jest.config.js
```diff
@@ -4,7 +4,7 @@
 module.exports = {
   // An array of glob patterns indicating a set of files for which coverage information should be collected
   collectCoverageFrom: [
-    'src/**',
+    'src/**/*.ts',
```
budnix, line 7: Curious whether this narrowing is because of the new `AGENTS.md` and `CLAUDE.md` files under `src/` — the old `src/**` would hand those to the babel coverage provider. If so, a one-line comment here would help, since the next person who spots a glob change in a move-only PR will want to put it back. It also moves the coverage denominator, so the percentage in the PR description is not directly comparable with develop's.

README.md
```diff
@@ -1,116 +1,41 @@
...
-  <a href="https://github.com/handsontable/hyperformula/actions?query=workflow%3Abuild+branch%3Amaster"><img src="https://img.shields.io/github/actions/workflow/status/handsontable/hyperformula/build.yml?branch=master" alt="GitHub Workflow Status"></a>
-  <a href="https://codecov.io/gh/handsontable/hyperformula"><img src="https://codecov.io/gh/handsontable/hyperformula/branch/master/graph/badge.svg?token=5k9ZQv8azO" alt="codecov"></a>
-</p>
+This repository holds the engine and everything built around it.
 
----
+| Directory | What it is | Published |
+|---|---|---|
+| [`hyperformula/`](hyperformula/) | The calculation engine. **Start here** — its [README](hyperformula/README.md) is the product documentation. | yes |
+| [`hyperformula-ui/`](hyperformula-ui/) | UI components for working with HyperFormula. Not imported yet. | yes |
```
budnix, line 10: Does the Published column mean the package ships from this repo, or that it exists on npm somewhere? Right now `hyperformula-ui/` holds only a README — no `package.json`, not in the `workspaces` array — so a reader following this row will not find anything here. Might be worth a third value, or a note saying where it is published from for now.

dev-docs/BUILD.md
```diff
@@ -0,0 +1,45 @@
...
+How the workspace installs, and the repository-level steps: generating the documentation, deploying the portal, and cutting a release.
+
+The engine's own build — the intermediate `lib/`, the bundles, and packaging — is in [`hyperformula/dev-docs/BUILD.md`](../hyperformula/dev-docs/BUILD.md).
+
+## Install
+
+Node version is pinned in [`.nvmrc`](../.nvmrc) — 22, the same in every package. The repository uses npm workspaces with a committed `package-lock.json`, so install with `npm ci`, not `npm install`.
+
+```bash
+npm ci              # the workspace: hyperformula, hyperformula-ui
```
budnix, line 12: `workspaces` in the root `package.json` lists only `hyperformula`, and `STRUCTURE.md` line 59 says `hyperformula-ui` is deliberately not listed yet. So two dev-docs pages on the same branch say opposite things, which is the failure the single-source rule is meant to stop.

```suggestion
npm ci              # the workspace: hyperformula
```

docs/guide/building.md
```diff
@@ -15,10 +15,22 @@ The build process uses Webpack and Babel, as well as npm tasks
 listed in package.json. During this process, the source located in
 the `src/*` directory is transformed into the output files.
```
budnix, line 16: The source is at `hyperformula/src/` now, and line 19 just below already says `hyperformula/tsconfig.json`, so this reads as if there were still a `src/` at the repository root. Small thing, but this is the public building guide, so it is the first place a new contributor looks.

```suggestion
the `hyperformula/src/*` directory is transformed into the output files.
```

docs/guide/building.md
```diff
@@ -15,10 +15,22 @@ The build process uses Webpack and Babel, as well as npm tasks
...
+```bash
+npm ci               # the engine and its toolchain
+npm run docs:install # the documentation portal, only if you are building the docs
+```
+
+The commands below are the engine's own scripts. Run them from the repository
+root, where each is forwarded to the package, or from `hyperformula/` directly.
+Only `bundle-all` and the `test:*`, `verify:*` and `docs:*` entry points are
+forwarded from the root; for anything else, add `--workspace=hyperformula` or
+work inside `hyperformula/`.
```
budnix, line 33: The root also forwards `build`, `compile`, `clean`, `typedoc:build-api`, `verify-bundles`, `audit`, `check:licenses` and `release`, so this list is shorter than what actually works. A contributor reading it will add `--workspace=hyperformula` to `npm run compile`, which already runs fine from the root, and may think `lint` has to be run per package when running it once from the root is the whole point.

```suggestion
Nearly all of them are forwarded from the root, so `npm run compile` works there
just as well as inside `hyperformula/`. `lint` runs once from the root over the
whole repository. If a script is not forwarded, add `--workspace=hyperformula`
or work inside `hyperformula/`.
```

package.json
```diff
@@ -1,632 +1,66 @@
...
+    "docs:build:cf": "npm run docs:install && npm run docs:generate-function-docs && npm run bundle-all && npm run typedoc:build-api && npm --prefix docs run build:cf --",
+    "docs:dev:cf": "npm --prefix docs run dev:cf --",
+    "docs:deploy:cf": "npm --prefix docs run deploy:cf --",
+    "docs:preview:cf": "npm --prefix docs run preview:cf --",
     "audit": "npm audit --omit=dev",
-    "clean": "rimraf coverage/ commonjs/ dist/ es/ languages/ lib/ typings/ test-jasmine/",
-    "compile": "tsc",
-    "check:licenses": "license-checker --production --excludePackages=\"hyperformula@3.1.1\" --onlyAllow=\"MIT; Apache-2.0; BSD-3-Clause; BSD-2-Clause; ISC; BSD; Unlicense\"",
-    "tsnode": "ts-node --transpile-only -O {\\\"module\\\":\\\"commonjs\\\"}",
+    "check:licenses": "node script/check-licenses.mjs",
```
budnix, line 47: `script/check-licenses.mjs` calls `npx license-checker`, but `license-checker` is a devDependency of the `hyperformula` workspace only. It works today because npm hoists it into the root `node_modules/.bin`. If a version conflict ever pushes it down into `hyperformula/node_modules`, `npx` will not fail — it will quietly fetch an unpinned copy from the registry and run that. Worth declaring it here, next to its only caller.

hyperformula/package.json
```diff
@@ -0,0 +1,614 @@
...
+    "core-js": "^3.39.0",
+    "cross-env": "^7.0.3",
+    "env-cmd": "^10.1.0",
+    "eslint": "^8.57.1",
+    "eslint-config-prettier": "^9.1.0",
+    "eslint-plugin-jasmine": "^4.2.2",
+    "eslint-plugin-jest": "^27.9.0",
+    "eslint-plugin-jsdoc": "^50.5.0",
+    "eslint-plugin-license-header": "^0.6.1",
+    "eslint-plugin-prettier": "^5.2.1",
```
budnix, line 137: The engine's `lint` is now `npm --prefix .. run lint`, so ESLint never runs from this package, and the root already declares all seven of these plus both `@typescript-eslint` packages on lines 123-124. Keeping copies here means two version ranges that can drift apart with nothing to catch it. This PR already removed two dead devDependencies for the same reason, so these look like the same cleanup.

hyperformula/LICENSE.txt
```diff
@@ -0,0 +1,34 @@
+Copyright (c) HANDSONCODE sp. z o. o.
```
budnix, line 1: This is byte-for-byte the same as the root `LICENSE.txt` and nothing keeps the two in step. The webpack banner reads this copy into every bundle, so it is what ships to users, while GitHub shows the root one — edit one and the shipped licence and the displayed one quietly disagree. A symlink probably will not survive `npm pack`, so maybe a check in `verify:publish-package` instead?

script/release/release.sh
```diff
@@ -1145,7 +1152,11 @@ elif $DRY_RUN; then
   echo "    (would ask for confirmation, then run: npm publish --tag $NPM_TAG)"
 else
   confirm_publish "$VERSION" "$NPM_USER"
-  run npm publish --tag "$NPM_TAG"
+  # The changelog lives at the repository root and npm cannot pack outside the
+  # package, so copy it in first. And publish the workspace, not the root, whose
+  # manifest is private.
+  run npm run copy-changelog --workspace=hyperformula
```
budnix, line 1158: `npm publish` runs `prepack`, which is `copy-changelog`, so this line does the same copy a second time. Harmless, but it prints as its own release action, which reads as if the publish would ship without a changelog otherwise. The same doubling is in `verify:publish-package`, where `npm pack` fires `prepack` too.

```suggestion
  # Publish the workspace, not the root, whose manifest is private. The
  # changelog lives at the repository root and npm cannot pack outside the
  # package, so the package's own prepack copies it in.
```

.eslintignore
```diff
@@ -5,24 +5,41 @@ node_modules
...
+hyperformula/typings
+docs/.vuepress/dist
 doc
-es
-languages
-lib
-script
-test-jasmine
-test-jest
 typedoc
```
budnix, line 44: `doc` is already on line 14, so this one is a duplicate. Also `doc`, `typedoc` and `script` are the only entries left unanchored — an unanchored pattern matches a directory of that name at any depth, which is the same trap the comment above describes for `languages`. Worth anchoring them too, or writing down why these three stay loose.

```suggestion
typedoc
```

hyperformula/src/interpreter/functionMetadata/FunctionDescription.ts
```diff
@@ -5,7 +5,7 @@
 
 /**
  * The function categories, in the order the built-in functions guide page presents them. This array is the source of
- * truth: `script/renderBuiltinFunctionsTable.ts` emits one `### <Category>` section per entry that has functions, in
+ * truth: `docs/script/renderBuiltinFunctionsTable.ts` emits one `### <Category>` section per entry that has functions, in
```
budnix, line 8: The edit is right, but the PR description says in bold that nothing under `hyperformula/src/` changed and that every file there is byte-identical. That claim is what lets a reviewer skip 200 files, so it is worth restating as comments only, or naming this file as the exception.

docs/guide/building.md
marcin-kordas-hoc: Adding the other half of why this sentence needs to go: it also over-claims. `verify:*` is *not* forwarded as a family — root `package.json` forwards 3 of the 8 `verify*` scripts `hyperformula/package.json` defines (`verify-bundles`, `verify:publish-package`, `verify:typings`); `verify:umd`, `verify:umd:min`, `verify:umd:full`, `verify:umd:full.min` and `verify:cjs` don't exist at root — `npm run verify:umd` from the repo root fails with `npm error Missing script: "verify:umd"`.

The file's own "Verify the build" section (lines 77-79) already says the individual checks are package scripts, so lines 31-33 contradict lines 77-79 within one file; both landed in the same squash-merge commit (72001ee1). Your rewrite drops the explicit `verify:*` claim, which fixes this too — worth keeping the pointer to `verify-bundles` as the one forwarded entry so the two sections agree.

hyperformula/src/interpreter/functionMetadata/FunctionDescription.ts
marcin-kordas-hoc: To make the exception list complete for the description: `git diff --name-status -M 286a7311..b562c145 -- 'src/**' 'hyperformula/src/**'` gives 200 R100, 1 R097 (this file, 2/2 lines in 15bc20a5) and 14 A — the 14 are `AGENTS.md` plus a `CLAUDE.md` symlink in `src/`, `src/DependencyGraph/`, `src/i18n/`, `src/interpreter/`, `src/interpreter/functionMetadata/`, `src/interpreter/plugin/` and `src/parser/`, with no old-location counterpart.

Suggested wording: "All 201 moved source files are renames — 200 byte-identical, one (`FunctionDescription.ts`) with a 2-line path-comment fix — plus 14 new `AGENTS.md`/`CLAUDE.md` files for agent navigation."

.claude/settings.json
```diff
@@ -0,0 +1,24 @@
+{
+  "enabledPlugins": [
+    "typescript-lsp@claude-plugins-official"
+  ],
+  "permissions": {
+    "deny": [
+      "Read(/hyperformula/lib/**)",
```
marcin-kordas-hoc, line 7: @qunabu re point 9a — the single leading slash here is not the filesystem root: in Claude Code permission paths a single `/` anchors at the settings source and the filesystem-root form is a double slash (`Read(//…)`). Reference: code.claude.com/docs/en/permissions.md, "Read and Edit" table (`/path` → "Path relative to the settings source"; `//path` → "Absolute path from filesystem root"). So `Read(/hyperformula/lib/**)` does match `<project>/hyperformula/lib/**` when Claude Code is launched from the repository root.

The caveat that does apply is closer to point 3: for project settings, `/path` resolves to `<primary working directory>/path` — the launch directory. A session started inside `hyperformula/` (which `building.md` presents as a normal way to work) either doesn't load the root file or resolves these rules to `hyperformula/hyperformula/lib/**`, so they match nothing there. Dropping the slash (`Read(hyperformula/lib/**)`) is cwd-relative and moves the same way. Cheapest fix: one sentence under `## .claude/settings.json` in `dev-docs/AGENT-TOOLING.md` saying the rules assume a root launch — or per-package `.claude/settings.json` once `hyperformula-ui` lands.

marcin-kordas-hoc (commented):
Three numbers in the description don't survive a recount at b562c145 — all in the direction of "counted once, edited later" (item 2, "Stale pull request descriptions", in this branch's own AGENTS.md list):

- Counts table: `wc -l` over the 12 `AGENTS.md` gives **140**, not 147; over the 9 `SKILL.md` gives **377**, not 378 (the 378 was exact at a49b8693, then 380 → 376 → 377 across 2e86b5cc, a000d8f8, dd62fadf). The file/glob counts — 12 / 12 symlinks / 9 / 5 with `paths:` — are exact.
- "Twelve call sites stopped reaching across a package boundary" — the pre-restructure root `package.json` (0c2372b6) has **13** scripts invoking a `script/*.js|ts` file: 5× `if-ne-env.js`, 5× `check-file.js`, `check-publish-package.js`, `generate-builtin-functions-doc.ts`, `prepare-cf-assets.js`. All 13 land in `hyperformula/package.json` or `docs/package.json` — only the word is off.
- "three PNGs were byte-identical duplicates … the other five files were referenced nowhere" — by blob SHA, **4 of the 8** deleted `examples/` files have an identical twin in `docs/.vuepress/public/`: `ast.png`, `ranges.png`, `sample sheet.png` (→ `sample-sheet.png`, different name, same blob) and `topsort.png`. The four CSVs are the unreferenced set.

Nothing functional; one pass over the description before merge. Two additive notes went into Budzio's threads on `docs/guide/building.md:33` and `FunctionDescription.ts:8`, and one on the `.claude/settings.json` deny rules.

Tobiadefami:
The version lookup in [release.sh:649](https://github.com/handsontable/hyperformula/blob/b562c145d5e68e7805c911f62542e0424f2118f7/script/release/release.sh#L649) is unchanged in this PR, but needs updating to match the new monorepo lockfile structure:

```js
require("./package-lock.json").packages["hyperformula"].version
```

The lockfile no longer has a top-level `version`, so `LOCK_VERSION` is always empty. Resuming a code freeze therefore deletes `node_modules` and `package-lock.json` and reinstalls dependencies, even when installation already completed for that version.

Could we update this lookup as part of the migration?

script/release/release.sh
Tobiadefami, line 649: The version lookup in release.sh:649 is unchanged in this PR, but needs updating to match the new monorepo lockfile structure:

`require("./package-lock.json").packages["hyperformula"].version
`

The lockfile no longer has a top-level version, so LOCK_VERSION is always empty. Resuming a code freeze therefore deletes  node_modules and package-lock.json and reinstalls dependencies, even when installation already completed for that version.

Could we update this lookup as part of the migration?

PR #1761, opened 2026-09-10
HF-131 (1/7): every interpreter cell error carries a message

marcin-kordas-hoc:
### What and why

HF-131 gives every cell error a real cause instead of just a type. This is the first of 7 stacked
PRs implementing that: every cell error thrown inside `src/interpreter/` — 78 previously
message-less `new CellError(...)` sites across 15 source files — now carries an `ErrorMessage` constant
describing why it happened.

Nearly all 78 sites collapse into two shared constants (46 `DIV_BY_ZERO` sites, one constant; most
of the 26 `NUM` sites reuse an existing or lightly-new constant); six anomalies needed an
individually-authored message (`Interpreter.ts`, `LookupPlugin.ts`, `FinancialPlugin.ts` ×2,
`InformationPlugin.ts`).

Also adds an ESLint rule (`no-restricted-syntax`, scoped to `src/interpreter/**/*.ts`) so a
message-less `new CellError(...)` construction there is now a lint error — mechanical enforcement
that this guarantee can't silently regress. Widened to all of `src/` in the next PR in this stack.

**The public `CellError` constructor is unchanged** — `message` stays optional, for custom
functions. Only construction inside the engine's own `src/` is now required to supply one.

### Verified

- `npx tsc --noEmit` — clean
- `npx eslint src/` — 0 errors (same pre-existing warning count as `develop`, no new warnings)
- The ESLint rule was proven to actually fire, not just parse: a plugin site was temporarily
  reverted to a bare `new CellError(ErrorType.NUM)`, confirmed to produce the expected lint error,
  then restored (`git diff` clean afterward)
- Full private test suite (`test/hyperformula-tests`, `--cacheDirectory` isolated,
  `--maxWorkers=2`), reproduced against this branch's tip: no regressions beyond the pre-existing
  branch-pinning mismatch unrelated to this change (5 known failures on bare `develop` too)

### Stack

1 of 7 — targets `develop`. Next: `feat/hf-131-error-messages-outside-interpreter`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Behavioral error types are unchanged; only optional detail messages are added plus lint guardrails in the interpreter. Low risk unless consumers assumed empty error messages.
> 
> **Overview**
[...]

PR #1762, opened 2026-09-10
HF-131 (2/7): the origin/propagation mechanism, plus the same guarantee outside the interpreter

marcin-kordas-hoc:
### What and why

Stacked on #1761.

**Revision note:** an earlier version of this description covered only the second half below.
The first half — the origin/propagation mechanism — was always in this PR's actual diff; the
description just didn't say so. #1764's description has been corrected to match (it does not
build the mechanism, only exposes two of its fields publicly).

**1. The origin/propagation mechanism.** `CellError` needs to know, for any error value anywhere
in a sheet, which function or operator originally produced it, whether it's since been merely
propagated by a cell that read it, and (for static errors with no formula root) which address it
originated at. The hard problem isn't computing that — it's stamping it exactly once, at the
moment of truth, regardless of which of the ~200 call sites across the interpreter and dependency
graph produced or merely relayed the error.

Two choke points, first-wins immutability (the same shape `attachRootVertex` already used for
`root`):
- `Cell.ts` gains `withOrigin`/`withArgumentIndex`/`asPropagated`/`withOriginAddress` — each a
  no-op once the relevant field (or `propagated`) is already set, otherwise a fresh immutable copy.
- `Interpreter.ts`'s `evaluateAst` stamps `originFunction` in a single postprocessing tail
  (`stampOriginForAstNode`) that runs after every AST node evaluates — including the NaN-overflow
  guard's freshly-minted error, and including the fix that a *propagated* error no longer gets
  `attachRootVertex`'d to whichever cell happened to read it.
- `DependencyGraph.ts`'s `getCellValue` — the one place every cross-cell read funnels through —
  marks every error read from another cell as `propagated`, and additionally stamps the address
  for a vertex holding a *static* error (no `FormulaVertex` to serve as a root).
- `Exporter.ts` falls back to the stamped `originAddress` when there is no `root`.

**Measured:** `A1='=1+'` reports `Sheet1!A1` where it reported no address; `B1='=A1'` reports
`Sheet1!A1` where it reported its own address; the same holds two hops out, through a range, and
through a named expression; it follows the cell across `addRows`; `=1/0` with a propagating `SUM`
still reports `A1`.

Also consolidates the `toEqualError` matcher: it existed twice, once per test runner, with
hand-copied logic and two strip lists that could drift — and had drifted in a way that made a
change appear inert under Jest. The decision now lives in one shared module both wrappers call
with their own equality function.

**Not yet public:** `originFunction`/`argumentIndex` are carried on `CellError` and stamped here,
but `DetailedCellError` does not expose them yet — that's #1764.

**2. The same guarantee outside the interpreter.** Extends #1761's guarantee to the 17
message-less sites (+2 with an explicit `undefined`) outside `src/interpreter/`: the parser's
`#REF!`/`#NAME?` branches, dependency transformers' 4 reference-removal sites, `DependencyGraph`'s
array-source-removed site, `FormulaVertex`'s outside-array-result site, and `Evaluator`'s two
`#CYCLE!` sites.

Kuba's scope ruling was *all* places throwing a cell error, not only `src/interpreter/` — these 19
sites are outside the spec's original 78-site counter but within that ruling.

Widens the ESLint rule from #1761 to all of `src/`, and documents the one call site it structurally
cannot catch: `CellContentParser.ts`'s `CellContent.Error` constructor keeps `message` optional on
purpose (its one no-message caller — a user typing e.g. `#REF!` literally into a cell — has no
engine-side cause to state), and `new CellError(errorType, message)` is syntactically a two-argument
call that passes the rule regardless of what the caller actually passes.

One new `ErrorMessage` constant, for an error literal written into a formula, is deliberately honest
rather than invented: an error literal typed into a formula or a value typed directly into a cell
round-trips exactly what the user wrote, so the message says that rather than fabricating an
engine-side cause.

### Verified

- `npx tsc --noEmit` — clean
- `npx eslint src/` — 0 errors (same pre-existing warning count)
- The widened rule was proven to reach files outside `src/interpreter/`: a `#CYCLE!` site in
  `src/Evaluator.ts` was temporarily reverted to a bare `new CellError(ErrorType.CYCLE)`, confirmed
  to produce the expected lint error, then restored (`git diff` clean afterward)
  pre-existing branch-pinning mismatch as #1761
- Nested-attribution correctness re-confirmed by re-running the test suite live: `=SUM(SQRT(-1))`
  reports `SQRT` not `SUM`; `=1/0` then `=SUM(A1)` reports `divide`, `propagated: true`; no code
  path exists to overwrite a propagated error's address; zero surviving message-less `CellError`
  construction sites in `src/` (338 call sites individually audited)

### Stack

2 of 7 — stacked on #1761. Next: `feat/hf-131-has-message`.

---

> [!NOTE]
> **Medium Risk**
> Touches core evaluation, dependency reads, and export paths for every cell error; behavior changes for error origin addresses (especially static errors after sheet structure changes) are user-visible.
> 
> **Overview**
> Extends **HF-131** with an **origin/propagation model** for `CellError` and applies the **mandatory cause message** rule outside the interpreter.
> 
> **Origin tracking:** `CellError` gains immutable helpers (`withOrigin`, `asPropagated`, `withOriginAddress`, etc.) so errors record which function/operator produced them, whether they were only read from another cell, and—for static errors without a formula root—a snapshot address plus transformation version. The interpreter stamps `originFunction` in one `evaluateAst` post-pass and no longer attaches the reading cell as `root` for propagated errors. **`DependencyGraph.getCellValue`** marks cross-cell reads as propagated and stamps static-error origins; **`Exporter`** resolves reported addresses via `root` or by replaying structural transforms through **`LazilyTransformingAstService.applyTransformationsToAddress`** (omitting the address after compaction when history is gone).
> 
> **Messages everywhere:** Parser, dependency transformers, evaluator cycle handling, array cleanup, and related paths now pass **`ErrorMessage`** constants instead of bare `CellError` types. New constants cover unresolved/removed references, circular refs, array edge cases, and formula error literals. ESLint **HF-131** applies to all `src/**/*.ts`, with a documented exception for user-typed error literals in `CellContentParser`.
> 
> **Tests:** Jest and Jasmine **`toEqualError`** matchers share one comparison module so structural ignore lists cannot drift between runners.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d8cd6cff59c5aeb8b3ce214703743c7f5162fff3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### Performance

Measured with the repo's own benchmark harness (`npm run benchmark:write-to-file`), `origin/develop`
c920375cb vs this stack's tip, three alternating base/head rounds on a 4-core machine. Total median
**−0.24%** (per-round totals +3.1 / −1.6 / −4.4%, i.e. the sign flips and no direction is
measurable; base-vs-base spread per benchmark 1.7–18.7%). One benchmark is slower in all three
rounds: **"Column ranges - add column", +2.8 / +3.3 / +4.2%, median +3.3% against a 1.9% base
spread.** The candidates are the extra `instanceof CellError` test at the `evaluateAst` tail and the
error-only `getCell` lookup in `DependencyGraph.getCellValue`; the two were not separated.

### Follow-up in this PR: a static error's address is now kept current

`DependencyGraph.getCellValue` records the address on the error value itself when a dependent reads
a cell that merely holds a static error. The reading cell caches that value and is not recomputed by
a row or column change that only shifts it, so the recorded address went stale and the dependent
kept naming the pre-shift cell — which could by then be empty or hold an unrelated value. The record
now carries the `LazilyTransformingAstService` version it was taken at, and `Exporter` replays the
intervening transformations over it, the same way an error produced by a formula already resolves
its address through its root vertex. Once a compaction has discarded those transformations the
address cannot be reconstructed and is omitted rather than guessed. Seven regression tests cover the
dependents, which the previous CRUD test did not: direct, two hops, through a range, on another
sheet, after an insert and after a removal, plus the compaction case.


qunabu:

Tobiadefami (commented):
Looks good. Verified that propagated errors retain their original source and that static-error addresses follow cell moves and row changes.

PR #1763, opened 2026-09-10
HF-131 (3/7): a consumer can tell 'no message' from 'empty message'

marcin-kordas-hoc:
### What and why

Stacked on #1762. Adds `DetailedCellError.hasMessage: boolean`.

`DetailedCellError.message` collapses two different situations into the same `''`: an error that
never carried a message at all (e.g. a custom function that omitted one — the public `CellError`
constructor keeps `message` optional), and an error with a deliberately empty message. A consumer
that needs to tell them apart previously couldn't. `hasMessage` does: `true` means a message was
set (even an empty string), `false` means none was.

Purely additive — no existing consumer of `.type`/`.message`/`.value`/`.address` needs a code
change. `toString()`/`valueOf()` are unaffected (both still return `.value` only).

### Verified

- `npx tsc --noEmit` — clean
- `npx eslint src/` — 0 errors
- Full private test suite reproduced against this branch's tip — no regressions beyond the same
  pre-existing branch-pinning mismatch as #1761/#1762

### Stack

3 of 7 — stacked on #1762. Next: `feat/hf-131-origin-function`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Additive public field on DetailedCellError with no changes to existing error string behavior or required consumer updates.
> 
> **Overview**
> Adds **`DetailedCellError.hasMessage`** so callers can distinguish an error that never had a `message` from one with a deliberately empty string — both still expose `message` as `''`.
> 
> The flag is set in the `DetailedCellError` constructor from whether the underlying `CellError` had `message` defined (`hasMessage: error.message !== undefined`). Existing `.type`, `.message`, `.value`, and `.address` behavior is unchanged; `toString()` / `valueOf()` still return only `.value`.
> 
> **CHANGELOG** documents the addition; **custom-functions** docs explain optional custom `CellError` messages vs engine errors. Test **`toEqualError`** structural comparison ignores `hasMessage` like other metadata fields.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 915ca1746e48272b958b181d4c35c1b9cf620317. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1764, opened 2026-09-10
HF-131 (4/7): expose which function produced an error, and which argument

marcin-kordas-hoc:
### What and why

Stacked on #1763. Exposes on the public `DetailedCellError` type the two fields the
origin/propagation mechanism (introduced in #1762) already computes internally:
`originFunction?: string` and `argumentIndex?: number` — which function or operator produced a
cell error, and, when it failed coercing one of its own arguments, which argument that was.

**Revision note:** an earlier version of this description said this PR built the mechanism
itself (the two choke points, `Cell.ts`'s builders, etc.). That was wrong — the mechanism landed
in #1762 (see its revised description). This PR only does the public-facing exposure step the
mechanism commit's own message called out as a separate, later concern:

> "originFunction and argumentIndex are carried on CellError and stamped in evaluateAst, but
> DetailedCellError does not expose them yet... That split is the natural PR boundary here."

This PR is that boundary: expose the two fields, and wire the one direct caller outside the two
choke points.

**What actually changes here (4 files, +32/-3):**
- `src/CellValue.ts` — `DetailedCellError` gains `originFunction`/`argumentIndex`, copied from the
  underlying `CellError` in the constructor. Additive only: enumerable property count on
  `DetailedCellError` grows from 5 (`type`/`message`/`value`/`address`/`hasMessage` from #1763) to 7.
- `src/interpreter/plugin/FunctionPlugin.ts` — `coerceArgumentsToRequiredTypes`'s two error-return
  paths (wrong-type, non-scalar) now call `.withArgumentIndex(i)` — the only direct callers of
  that builder outside the two choke points #1762 introduced.
- `src/Cell.ts` — `withArgumentIndex`'s no-op gate also checks `originFunction !== undefined`, not
  just `argumentIndex`: needed so a nested call's own error (`SQRT(-1)` inside
  `=DATE(1,1,SQRT(-1))`) doesn't get `DATE`'s argument index attached after already claiming its
  own origin.
- `CHANGELOG.md` — documents the two new public fields.

**Verified against nested/propagated cases directly:** `=SUM(SQRT(-1))` reports `SQRT`, not `SUM`;
`=1/0` then `=SUM(A1)` reports `divide`, `propagated: true`, not re-claimed by `SUM`;
`=DATE(1,1,SQRT(-1))` reports `{originFunction: 'SQRT', argumentIndex: undefined}`, not a leaked
index on `DATE`'s argument slot.

**Known, deliberate limit:** a binary operator's error names the operator (`addWithEpsilon`,
`divide`, etc.), not which operand failed — `argumentIndex` is only set for a *function's* own
coercion loop; the two operands of a binary op have no equivalent loop. Not solved here.

### Verified

- `npx tsc --noEmit` — clean
- `npx eslint src/` — 0 errors
  pre-existing branch-pinning mismatch as the rest of this stack
- Zero surviving message-less `CellError` construction sites in `src/` (carried over from #1762;
  re-confirmed at this branch's tip)

### Stack

4 of 7 — stacked on #1763. Next: `fix/hf-131-rootless-origin`.

---

> [!NOTE]
> **Medium Risk**
> Changes public error metadata and attribution rules during argument coercion; formula results should be unchanged but consumers and diagnostics depend on the new fields.
> 
> **Overview**
> Exposes **`originFunction`** and **`argumentIndex`** on public **`DetailedCellError`**, copying them from the internal **`CellError`** so callers can see which function/operator produced an error and which formula argument failed coercion.
> 
> **`FunctionPlugin`** now tracks the **user-written argument index** through range expansion (`listOfScalarValues` / `evaluateArguments` add a 4th tuple field) and passes it into **`coerceArgumentsToRequiredTypes`**, which calls **`.withArgumentIndex(...)`** on wrong-type and non-scalar coercion failures so indices match the formula (e.g. a range counts as one argument, not N flattened cells).
> 
> **`Cell.withArgumentIndex`** is tightened so it is a no-op once **`originFunction`** is already set, preventing outer functions from attaching their argument slot to errors that nested calls already claimed (e.g. `SQRT(-1)` inside `DATE(...)`).
> 
> **CHANGELOG** documents the new fields and related error-metadata work in the HF-131 stack.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit efefce383c775f7ffd84f53d1a1a8bea9a4f981d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### Performance

The hot-path change lives in #1762 and is measured there: total median −0.24% over three alternating
base/head rounds against `origin/develop`, within the run-to-run spread, with one CRUD benchmark
("Column ranges - add column") consistently about 3% slower. This PR adds fields to an object that
is only built when an error is exported, so it carries no measurable cost of its own.


qunabu:

Tobiadefami (changes requested):
Could we preserve the original argument index through range expansion? With `A1:A3` containing `1`, `=AND(A1:A3,"x")` reports `argumentIndex: 3`, but `"x"` is the second argument, so the index should be `1`. The `.withArgumentIndex(i)` calls in `FunctionPlugin.ts:488–492` use the flattened position.

Carrying the original index alongside each expanded value fixed this for `AND`, `OR`, and `XOR` in a temporary patch. All 164 tests in the selected suites passed, along with TypeScript and lint.

marcin-kordas-hoc (commented):
Fixed in `223711dd` — thanks, the diagnosis was exact.

`sourceArgumentIndexes` now travels with the expanded values, and `coerceArgumentsToRequiredTypes` reports `writtenArgumentIndex(i)` instead of the loop position. On the current head `=AND(A1:A3,"x")` with `A1:A3` filled reports `argumentIndex: 1`; same for `OR` and `XOR`. Pinned in the tests repo by "points at the argument the user wrote, not at its position after range expansion".

Could you re-review when you get a chance? The CHANGES_REQUESTED is the only thing holding this one.

PR #1765, opened 2026-09-10
HF-131 (5/7): no error reports a cell that did not produce it

marcin-kordas-hoc:
### What and why

Stacked on #1764. Closes the remaining cases where an error could report a cell that did not
produce it — the general case (propagation through a function/operator) is already covered by
the mechanism in #1762; this handles the rest:

- A parsing error now reports `originFunction: 'parser'` (`ParsingErrorVertex.getCellValue`).
- A value typed directly into a cell (e.g. literal `#REF!`, no leading `=`) reports
  `originFunction: 'user input'`.
- An error literal written into a formula (e.g. `=#REF!`) reports `originFunction: 'literal'`.
- `#SPILL!` errors now report the address of the cell that failed to spill, instead of no address
  at all — `setNoSpace()` now attaches the vertex as root, the same way every other formula result
  does.

**Known, documented limit:** the six reference-checking branches in the formula parser that build
a `#REF!`/`#NAME?` error for a malformed reference (e.g. a range spanning more than one sheet)
still report no `originFunction` — left unattributed rather than mislabeled as a function that only
read the value. Noted in `docs/guide/known-limitations.md`.

### Verified

- `npx tsc --noEmit` — clean
- `npx eslint src/` — 0 errors
  pre-existing branch-pinning mismatch as the rest of this stack
- Probed directly against the live engine (not just the test suite's own assertions): a formula
  reading another cell's error reports that cell's address, not its own reading cell's, through
  chains up to 3 hops and across sheets; a parsing error's address is preserved through a reader;
  a `#SPILL!` fixture now reports the failing cell's address where it previously reported none

### Stack

5 of 7 — stacked on #1764. Next: `fix/hf-131-issue-444-addresses`.

---

> [!NOTE]
> **Medium Risk**
> Touches core evaluation, parsing, and dependency-graph error paths; behavior of `DetailedCellError` metadata changes but formula results should stay the same.
> 
> **Overview**
> **HF-131 (5/7)** tightens **error origin and address attribution** so consumers of `DetailedCellError` see who produced an error and which cell it belongs to.
> 
> **Origins at the source:** Parsing failures use `parser`; literal `#REF!` typed in a cell uses `user input`; `=#REF!` in a formula uses `literal`. Broken or missing references (parser, interpreter, dependency graph) use `reference`; references invalidated by row/column deletes use `removed reference`; `OFFSET` out-of-bounds uses `OFFSET`. That stops enclosing functions from claiming errors they only read (e.g. `SUM` on an oversized range).
> 
> **`#SPILL!`:** `setNoSpace()` now attaches the formula vertex as root so the spill error reports the cell that could not spill, not a missing address.
> 
> **Arrays:** After evaluation, errors inside ad-hoc array results get `stampOriginForArrayElements`, so broadcast gaps and per-element failures credit the operator that built the array (e.g. `+` in `INDEX(A1:A2+B1:B3, …)`) instead of the function that indexes the array.
> 
> CHANGELOG and **known-limitations** document the origin vocabulary and that `#SPILL!` / `#CYCLE!` have no `originFunction`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e2ced02177de579cbd9a4154a508b6a8f80c3dea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### Follow-up in this PR: a function no longer claims an error it only read

An error describing a broken reference exists before any function sees it, so the enclosing call is
only a reader. `originNameForAstNode` returns `undefined` for reference and error nodes, so such an
error reached the enclosing `FUNCTION_CALL` unclaimed and non-propagated and the postprocessing tail
handed it over: `=SUM(A2:A99999999999)` reported `SUM`, `=ABS(A2:A99999999999)` reported `ABS` with
an argument index that belonged to nothing, and `=SUM(#REF!)` after removing rows reported `SUM`.

The identity is now attached where the error is built, so the existing first-wins rule blocks the
reader by itself. A reference that cannot be resolved reports `reference` — whether the parser
rejected it, the sheet is missing, the range spans several sheets, or the name is unknown — and one
destroyed by removing rows or columns reports `removed reference`. The out-of-sheet error raised
inside the `OFFSET` heuristic reports `OFFSET`, because `OFFSET` really did produce it.

**Open question for review:** `removed reference` is a new public vocabulary word, and it is the
naming question raised earlier on this PR. It follows the existing lowercase-prose identities
(`parser`, `user input`, `literal`).

Errors that arise inside an array result are unchanged and still take the reading function's name —
`=INDEX(A1:A2+B1:B3, 3, 1)` reports `INDEX` for an `#N/A` the `+` operator produced. Covering them
needs either a hook over every element of every ad-hoc array, which changes values in unrelated
specs, or the function name threaded through three methods of the plugin base class. Left out
deliberately and stated in `known-limitations.md`.


qunabu:

marcin-kordas-hoc:
Ran a closed-loop comparison of this stack's origin/address behavior against real Excel via
MS Graph (13 scenarios, full write-up + raw evidence in a private handoff repo — happy to share the
methodology if useful). Everything on the CRUD-address side of this PR checked out — a REF error
from a deleted column reports the same type as Excel and the address stays consistent before/after
the delete.

One thing surfaced that isn't a bug, but is worth a product call: the CRUD-produced `#REF!` from
`Transformer.ts`'s column/row-removal branches never calls `.withOrigin(...)`, so `originFunction`
comes back `undefined` for it — confirmed live, the field is genuinely absent, not empty. That's
consistent with how the mechanism works (an `AstNodeType.ERROR` node has no entry in
`originNameForAstNode`, so the stamp is a no-op), but neither this PR's own "deliberate limits"
nor `known-limitations.md`'s existing bullet (which only covers the 6 parser reference-checking
branches) actually says whether this specific case is intentional.

Should a CRUD-removed reference get an `originFunction` too (e.g. something like `'row/column
removed'`), or is leaving it unattributed here the right call, same as the parser branches? Either
answer is fine by me — just flagging that it isn't currently written down either way.

Tobiadefami (commented):
The implementation looks good.

Issue #1545, opened 2025-09-25
Feature: Iterative Calculation


sequba:
From [Excel docs](https://support.microsoft.com/en-gb/office/remove-or-allow-a-circular-reference-in-excel-8540bd0f-6e97-4483-bcf7-1b49cd50d123#bkmk_learn_about_iterative):

<img width="819" height="908" alt="Image" src="https://github.com/user-attachments/assets/cde66394-dbfb-4c56-98b1-545e720f3119" />

Google Sheets also has this feature: https://workspaceupdates.googleblog.com/2016/12/new-iterative-calculation-settings-and.html

Discussion about this feature in our GH: https://github.com/handsontable/hyperformula/discussions/1110

PR #1770, opened 2026-09-11
Docs: document the circular-reference default-behavior difference from Excel

marcin-kordas-hoc:
### What and why

Documentation only. Adds a row to `docs/guide/list-of-differences.md` for a genuine circular
reference under default settings (`A1:=A1+1`), distinct from the existing "Dependency collection"
row, which is scoped specifically to *false* cycles caught by static dependency analysis
(`IF(FALSE(), A1, 0)`) — a different mechanism, a different result shape.

- HyperFormula always returns `CYCLE` unless iterative calculation is explicitly enabled
  (iterative calculation isn't yet part of a released version — [#1545](https://github.com/handsontable/hyperformula/issues/1545)).
- Excel's default (iterative calculation off) silently settles the cell to a computed value
  (`0` for a simple self-reference), with no cell-level error at all and no way to detect the
  cycle from the result alone.

Measured live against Excel Online via MS Graph while closed-loop testing a different, unrelated
stack (message/origin/address on cell errors) — this difference predates that work and isn't
something it changes. Google Sheets' behavior for this specific case wasn't verified this session
and is marked "Not verified" rather than assumed from the adjacent row.

### Verified

- `npm run docs:build` — clean; the new row renders correctly and the `#1545` link resolves

---

> [!NOTE]
> **Low Risk**
> Documentation-only change to the compatibility guide; no runtime or API impact.
> 
> **Overview**
> Adds a **Circular references** row to the compatibility table in `list-of-differences.md`, using `A1:=A1+1` to document default behavior separately from the existing **Dependency collection** row (static/false cycles).
> 
> The new row states that HyperFormula always surfaces a `CYCLE` error (with a link to [#1545](https://github.com/handsontable/hyperformula/issues/1545) for unreleased iterative calculation), marks Google Sheets as not verified for this case, and notes that Excel with iterative calculation off shows no cell error and settles to a value (e.g. `0`) with no cycle visible from the result alone.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 838df28bbd09da4ea8fce652e17c511ded4c0a44. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Tobiadefami (commented):
Looks good from the checks performed.  

Verified that =A1+1 in A1 displays 0 in Excel Online.

PR #1771, opened 2026-09-12
Fix: parser rejects formulas with Excel's internal prefixes (#1655)

### LINKED ISSUE #1655: [Bug]: Parser fails on formulas with _xlfn.xxx prefix added by Excel for backwards compatibility
### Description

HyperFormula fails to parse formulas containing Excel's internal function prefixes. These prefixes are present in the raw XML of every .xlsx file for any function added after the original OOXML spec (~Excel 2007) and affect 180+ functions including commonly used ones like IFS, SWITCH, XLOOKUP, CONCAT, TEXTJOIN, MAXIFS, and FILTER.

Users who parse .xlsx files with libraries like SheetJS or ExcelJS get the raw prefixed formula strings (e.g., =_xlfn.IFS(C5<C7,"Pass",C5>=C7,"Fail")). Passing these to HyperFormula produces:

Parsing error. Redundant input, expecting EOF but found: (

Every major spreadsheet library (SheetJS, ExcelJS, openpyxl, Apache POI, LibreOffice) strips these prefixes on import. HyperFormula currently does not.

**Root cause**
The `ProcedureName` token in `src/parser/LexerConfig.ts` requires the first character to be a letter:

```
pattern: new RegExp(`([${UNICODE_LETTER_PATTERN}][${NON_RESERVED_CHARACTER_PATTERN}]*)\\(`)
// UNICODE_LETTER_PATTERN = 'A-Za-z\u00C0-\u02AF' — no underscore
```

Since every Excel internal prefix starts with `_`, none of them match as a function name. The lexer tokenizes the prefix as something else, then reports "expecting EOF but found (".

Reproduction
```
const hf = HyperFormula.buildFromSheets({
  Sheet1: { cells: [[{ cellValue: 10 }, {}, { cellValue: 5 }]] } // A1=10, C1=5
});

// Works
hf.setCellContents({ sheet: 0, row: 0, col: 3 }, [['=IFS(A1>C1,"Pass",A1<=C1,"Fail")']]);
hf.getCellValue({ sheet: 0, row: 0, col: 3 }); // "Pass"

// Fails — same formula with Excel's internal prefix
hf.setCellContents({ sheet: 0, row: 0, col: 3 }, [['=_xlfn.IFS(A1>C1,"Pass",A1<=C1,"Fail")']]);
hf.getCellValue({ sheet: 0, row: 0, col: 3 }); // CellError: Parsing error
```

Verified against HyperFormula 3.2.0.

Prefixes to strip
All of these produce the same parse error and should be stripped before lexing:

Prefix	Introduced	Meaning	Example
_xlfn.	Excel 2010	Future functions	_xlfn.IFS(...), _xlfn.XLOOKUP(...)
_xlfn._xlws.	Excel 365	Future + worksheet-scoped	_xlfn._xlws.FILTER(...)
_xlws.	Excel 365	Worksheet-scoped	_xlws.SORT(...)
_xlpm.	Excel 365	LAMBDA/LET parameter names	_xlpm.x inside a LAMBDA body
_xludf.	Excel 365	User-defined (LAMBDA in Name Manager)	_xludf.MyCustomFn(...)
All verified to fail against HyperFormula 3.2.0 with the same "Parsing error. Redundant input, expecting EOF but found: (" error.

Suggested fix
A formula preprocessing step that strips the prefixes before they reach the lexer:

`formula.replace(/_xlfn\._xlws\.|_xlfn\.|_xlws\.|_xlpm\.|_xludf\./g, '')
`
(Longer prefixes ordered first so _xlfn._xlws. matches fully rather than being partially consumed by _xlfn..)

This is low risk — these prefixes are serialization artifacts, not real function names, and stripping them is the standard approach used by every other spreadsheet engine.

Raised in passing in #1392 but never addressed. Also reported on the Handsontable [forum](https://forum.handsontable.com/t/error-message-when-using-ifs-function/9057) by users importing .xlsx files.

### Video or screenshots

_No response_

### Demo

https://forum.handsontable.com/t/error-message-when-using-ifs-function/9057

### HyperFormula version

3.2

### Your framework

_No response_

### Your environment

Chrome on MacOS

marcin-kordas-hoc:
## Problem

The parser requires a function name to start with a letter, so it rejects formulas
that carry the internal prefixes Excel writes into `.xlsx` for every function added
after ~2007 (`_xlfn.`, `_xlws.`, and combinations) — e.g. `_xlfn.IFS(...)`. Whether a
user hits this depends on their import library: ExcelJS (the library our own
[file-import guide](docs/guide/file-import.md) uses in its example) passes the
prefixes straight through, and SheetJS strips `_xlfn.` but not `_xlws.`, so `FILTER`
and `SORT` still fail. Registering the prefixed name as a custom function is not a
workaround — the lexer rejects the leading underscore before name resolution runs.

Fixes #1655

## Fix

- The five known prefixes (`_xlfn.`, `_xlfn._xlws.`, `_xlws.`, `_xlpm.`, `_xludf.`)
  are now an optional part of the `ProcedureName` and `OffsetProcedureName` lexer
  tokens, so `_xlfn.IFS(...)` tokenizes the same as `IFS(...)`.
- The prefix is stripped in the **lexer**, not by rewriting the raw formula string.
  A string-level `replace` (the fix sketched in the issue) also rewrites the prefix
  inside string literals — `="_xlfn.IFS is a literal"` would silently become
  `="IFS is a literal"`. There's a regression test for this.
- `FormulaParser` and `ParserWithCaching.computeHashFromTokens` both derive the
  canonical function name through shared helpers (`canonicalProcedureNameFromToken`,
  `canonicalOffsetProcedureNameFromToken`), so a prefixed formula and its unprefixed
  equivalent — OFFSET included — produce the same cache hash.
- `getCellFormula()` / `getCellSerialized()` return the formula without the prefix.

**Scope notes:**
- A bare `_xlpm.x` identifier (a `LAMBDA`/`LET` parameter, not a call) is not
  stripped — HyperFormula has no `LAMBDA`/`LET`, so such a formula is `#NAME?`
  either way.
- Prefixes are matched in lower case only for `ProcedureName` (every ordinary
  function), which is all Excel ever writes. `OffsetProcedureName` is the one
  exception — see "Post-review changes" below.
- A prefix doesn't add a function: `_xlfn.CONCAT(...)` is `#NAME?` (CONCAT isn't
  implemented), not a parsing error — strictly better than before, not silently
  swallowed.

## Changed files

| File | Change |
|------|--------|
| `src/parser/parser-consts.ts` | new `EXCEL_INTERNAL_FUNCTION_PREFIX_PATTERN` |
| `src/parser/LexerConfig.ts` | `ProcedureName`/`OffsetProcedureName` accept an optional prefix; new `canonicalProcedureNameFromToken()` / `canonicalOffsetProcedureNameFromToken()` helpers |
| `src/parser/FormulaParser.ts` | uses the shared helper |
| `src/parser/ParserWithCaching.ts` | uses the shared helpers for the cache hash, incl. OFFSET |
| `docs/guide/file-import.md` | documents the prefixes and which import libraries leave them in |
| `CHANGELOG.md` | Unreleased entry |

## Tests

Regression tests in `handsontable/hyperformula-tests` (branch
`fix/hf-217-excel-internal-prefixes`, 28 tests, all pass):

| Group | Tests | Coverage |
|-------|-------|----------|
| Prefix ignored on a call | 5 | each of the 5 prefixes evaluates like the unprefixed call |
| Works for every kind of function | 5 | OFFSET (own grammar rule), SUM (predates prefixes), XLOOKUP (array fn), a custom function, a localized name |
| Nested / non-leading calls | 3 | prefix mid-expression, nested argument, every function in the formula prefixed |
| Stripped from the stored formula | 5 | `getCellFormula`, `getCellSerialized`, shared cache hash (incl. OFFSET, two prefixes) |
| Left alone / precedence unchanged | 4 | string literal, string argument, named expression, a named expression shadowed by a call same as without a prefix |
| Does not over-match | 5 | unimplemented function is `#NAME?` not a parse error, upper-cased prefix rejected (except OFFSET, pinned separately), unrelated underscore name, prefix alone |
| AST shape | 2 | function-call node type, canonical name |

- Verified against unpatched `src/` at each stage: the relevant new tests go red,
  confirming they detect the bug rather than passing regardless.
- Full private unit suite: 502 suites / 6257 tests, no regressions.
- `npm run test:performance`: exit 0. Targeted lexer microbenchmark (parser is on
  the `buildFromSheets` hot path) shows no measurable regression across every
  round of changes below.

## Post-review changes

An exhaustive multi-agent review (5 independent dimensions, each finding
adversarially verified by 3 more agents before being trusted) ran against this PR
after the first green CI, on top of my own manual review. It surfaced 17 raw
findings; 6 survived verification, 3 of which converged independently on the same
root cause. What changed as a result:

1. **CHANGELOG entry linked the issue instead of the PR** (Bugbot). Fixed — now
   links #1771, matching the majority of neighboring Unreleased entries.
2. **`_xlfn.OFFSET(...)` and `OFFSET(...)` did not share a cache hash**, unlike
   every other function — `computeHashFromTokens` only canonicalized
   `ProcedureName` tokens; `OffsetProcedureName` is a separate, uncategorized token
   type (OFFSET has its own grammar rule) and fell through to the generic
   raw-image branch. Verified empirically before/after with the parser's own
   `tokenizeFormula()`. Fixed with a dedicated branch + `canonicalOffsetProcedureNameFromToken()`.
3. **`canonicalProcedureNameFromToken` was measurably slower per call than the
   inline code it replaced** for the common (no-prefix) case — real but bounded
   (the full-pipeline lexer benchmark never showed it). Added a cheap
   leading-underscore guard before invoking the prefix-stripping regex at all,
   since every prefix starts with `_`. A reorder alternative suggested by the
   review was measured and found to *not* help (and to slightly hurt the prefixed
   case), so it was not adopted — verifying a benchmark's own claim mattered here.
4. **`OffsetProcedureName`'s prefix matching is case-insensitive**, unlike
   `ProcedureName` — `_XLFN.OFFSET(...)` is silently accepted where `_XLFN.SUM(...)`
   is a parse error. Real and now documented and pinned by a test, not fixed in
   code: OFFSET predates the OOXML cutoff this whole prefix scheme exists for, so
   Excel can never actually emit a prefixed OFFSET call in any case — a dedicated
   case-sensitive-prefix matcher would add real lexer complexity for an input
   nothing can produce.
5. **A named expression named e.g. `_xlfn.Sum` is shadowed by a call to `SUM`**
   the moment it's followed by `(`. Verified this is pre-existing, general engine
   behavior (an unprefixed named expression named `Sum` is shadowed identically,
   unrelated to this PR) — this PR only changes what used to be a parse error into
   the same precedence rule every other name already follows. Pinned with a test
   rather than changed.
6. **`docs/guide/file-import.md` overclaimed `_xlpm.` handling** — it said
   HyperFormula "ignores" the prefix, but the code only strips a prefix in front of
   a function call, and `_xlpm.` never appears there in a real `.xlsx` (it always
   prefixes a bare `LAMBDA`/`LET` parameter name, which this fix deliberately does
   not touch). Corrected.

Also found and fixed in the same pass: the existing "shares one cache entry" test
tokenized through a separately-built lexer instead of the parser's own
`tokenizeFormula()` — harmless for `ProcedureName` (a module-level singleton token
type) but would have silently never caught a bug in `OffsetProcedureName` (rebuilt
fresh per `buildLexerConfig()` call, so a second lexer's token fails the identity
check `tokenMatcher` relies on). Both the existing test and the new ones now
tokenize via the parser's own method.

---

> [!NOTE]
> **Medium Risk**
> Changes core formula lexing/parsing on a hot path for every function call, but behavior is additive (ignore known prefixes) with shared canonicalization for caching; main risk is incorrect prefix matching edge cases.
> 
> **Overview**
> Fixes formula parsing when workbooks are imported with Excel’s internal function-name prefixes (`_xlfn.`, `_xlfn._xlws.`, `_xlws.`, `_xlpm.`, `_xludf.`), which commonly appear in `.xlsx` strings from parsers such as ExcelJS. Formulas like `=_xlfn.IFS(...)` previously failed at lex time; they now tokenize and evaluate like `=IFS(...)`.
> 
> The lexer treats those prefixes as optional on **`ProcedureName`** and **`OffsetProcedureName`**, strips them via shared **`canonicalProcedureNameFromToken`** / **`canonicalOffsetProcedureNameFromToken`** helpers (fast path when the name does not start with `_`), and uses the same canonicalization in **`ParserWithCaching`** so prefixed and unprefixed formulas share a cache hash. Prefixes are not rewritten in the raw formula string, so text inside string literals stays unchanged. **`getCellFormula()`** is documented to return the unprefixed form; unsupported functions still yield **`#NAME?`**, not a parse error.
> 
> Docs in **`file-import.md`** and **`CHANGELOG.md`** describe the prefixes and import-library behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fbf730682c73edbefd934d229c69b21dfcd7334c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Fixed
 
+- Fixed the parser rejecting formulas that carry the function-name prefixes Excel writes into `.xlsx` files (`_xlfn.`, `_xlfn._xlws.`, `_xlws.`, `_xlpm.`, `_xludf.`), which made every function added after Excel 2007 (e.g. `IFS`, `XLOOKUP`, `TEXTJOIN`, `FILTER`, `SORT`) fail to parse when a workbook was imported with a library that passes the prefixes through. The prefixes are now ignored, and `getCellFormula()` returns the formula without them. [#1655](https://github.com/handsontable/hyperformula/issues/1655)
```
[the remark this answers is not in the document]
marcin-kordas-hoc: Fixed in d7f8e4a65 — the entry now links the PR (#1771) instead of the issue (#1655), matching the majority convention in this Unreleased section.

GreenFlux (approved):
Looks good, thanks for the quick response!

src/parser/LexerConfig.ts
```diff
@@ -93,7 +145,15 @@ export const buildLexerConfig = (config: ParserConfig): LexerConfig => {
...
-  const OffsetProcedureName = createToken({ name: 'OffsetProcedureName', pattern: new RegExp(offsetProcedureNameLiteral, 'i') })
+  // OFFSET has its own token because it has its own grammar rule, so it needs the prefix handling of
+  // ProcedureName repeated here. The 'i' flag exists for the translated OFFSET name (pre-dates this
+  // prefix support) and applies to the whole pattern, so unlike ProcedureName it also makes the prefix
+  // case-insensitive: `_XLFN.OFFSET(...)` is accepted here where `_XLFN.SUM(...)` is a parsing error.
+  // Left as-is rather than given a dedicated case-sensitive-prefix matcher: OFFSET predates the OOXML
+  // cutoff this whole prefix scheme exists for, so Excel can never actually emit a prefixed OFFSET call
+  // in any case — the asymmetry has no reachable real input, only hand-typed formulas (see the
+  // "an upper-cased prefix is accepted on OFFSET, unlike everywhere else" test).
+  const OffsetProcedureName = createToken({ name: 'OffsetProcedureName', pattern: new RegExp(`(?:${EXCEL_INTERNAL_FUNCTION_PREFIX_PATTERN})?${offsetProcedureNameLiteral}`, 'i') })
```
Tobiadefami, line 156: This currently breaks an existing named-expression reference:

```ts
const engine = HyperFormula.buildFromArray([[null]], {
  licenseKey: 'gpl-v3',
})
const address = {sheet: 0, row: 0, col: 0}

engine.addNamedExpression('_xlfn.OFFSET_RATE', 42)
engine.setCellContents(address, [['=_xlfn.OFFSET_RATE']])

engine.getCellValue(address) // Expected: 42
```

The regression test passes on base `c920375c`, but fails on head `a2185233` with:

```text
Expected: 42
Received: #ERROR!
Parsing error: expected LParen, found '_RATE'
```

The new pattern recognizes `_xlfn.OFFSET` without requiring `(`, splitting the complete name into an `OFFSET` token and `_RATE`. Could we preserve the named-expression reference when there is no function call?

Tobiadefami (changes requested):
Could we restrict the prefixed `OFFSET` token to function calls?

docs/guide/file-import.md
```diff
@@ -23,6 +23,27 @@ To import CSV files, use a third-party [CSV parser](https://www.npmjs.com/search
...
+
+Excel stores some function names with an internal prefix. It marks every function added after Excel 2007 this way, so a file can contain `_xlfn.IFS(...)` where the user typed `IFS(...)`. The prefixes are an artifact of how Excel saves the file, not part of the function name.
+
+Which prefixes reach your code depends on the parser you use. ExcelJS passes them through. SheetJS removes `_xlfn.` but keeps `_xlws.`.
+
+HyperFormula ignores these prefixes, so you can pass the formula straight to the engine:
+
+| Prefix | Example |
+| --- | --- |
+| `_xlfn.` | `=_xlfn.IFS(A1>B1,"Pass","Fail")` |
```
Tobiadefami, line 36: Could we correct the `IFS` example here and in the `getCellFormula()` explanation on line 43?

The current formula has three arguments:

```text
=_xlfn.IFS(A1>B1,"Pass","Fail")
```

`IFS` expects condition/result pairs, so this returns `#N/A` with “Wrong number of arguments.” Could we use:

```text
=_xlfn.IFS(A1>B1,"Pass",A1<=B1,"Fail")
```

The corresponding `getCellFormula()` output would be:

```text
=IFS(A1>B1,"Pass",A1<=B1,"Fail")
```

